PR #51, opened 2015-01-30
check for $accession_source to adapt for eg! use

ckongEbi:


magaliruffier:
that looks fine, thanks for the change

PR #57, opened 2015-02-13
Remove some old modules

willmclaren:
TranscriptSNPs and TranscriptAlleles were long ago superceded by modules in the ensembl-variation repo. The only other reference I can find was a use statement in the webcode which then makes no further reference to the method pulled in (I've put in a pull request to remove this too).


avullo:
Thank you, much appreciated.

PR #58, opened 2015-03-18
Not sure why a regexp is in use here, but this is the maximum quantifier

dbolser-ebi:


magaliruffier:
The change looks fine, but I'm not sure I understand the commit message coming with it

nerdstrike:
Are you actually requesting FASTA files with 30000 characters per line?

The Regex is used because it is performant when reformatting lines to a specified width.  It used to be done by other means but was prohibitively slow. You could probably achieve the same with a bunch of substr() action but the code would be less pretty and not necessarily faster.

andrewyatz:
@dbolser-ebi is referring to the regex we use later on `$seq =~ s/(.{1,$width})/$1\n/g;`. Seems that there's an inbuilt maximum limit to what `$width` can be (32766) without Perl's regex system having a paddy. Perldoc from http://perldoc.perl.org/perlre.html#Regular-Expressions says:

> The "*" quantifier is equivalent to {0,} , the "+" quantifier to {1,} , and the "?" quantifier to {0,1} . n and m are limited to non-negative integral values less than a preset limit defined when perl is built. This is usually 32766 on the most common platforms. The actual limit can be seen in the error message generated by code such as this:

Not sure someone would want to create FASTA files of such length but hey this is kinda of an important thing to mention. Also it's not something that's in tablets of stone but something you can find out from running

```perl
$_ **= $_ , / {$_} / for 2 .. 42;
```

+1 for the change but would like some more information in the documentation about why the limit exists. Just to avoid this kind of thing again

dbolser-ebi:
Actuality, we want 'un wrapped fasta, with all sequence on one line.

BioPerl uses unpack, but not sure about performance.

nerdstrike:
I've benched it for jollies. unpack(); join() is about half again as fast as regex for this purpose. The code is dramatically less legible, and will use a more memory, but it's probably worth dropping in as it lifts the width limit.

andrewyatz:
@dbolser-ebi wow. ok. any particular reason why? just really interested :)

dbolser-ebi:
@andrewyatz https://sourceforge.net/p/kmer/bugs/28/#9ea8

muffato:
Fasta files with the sequences on 1 line are easy to create with awk / SQL.
From the Compara database, it is very easy to create a Fasta file with all the genes in a tree / homologues to a query gene / etc with a SQL query. The maximum length of a transcript is 108kb

andrewyatz:
That's fair enough but the code we're dealing with here is one for formatting into a smaller fixed width. It doesn't impact the parsing of large single line sequences

Matthieu Muffato wrote:

nerdstrike:
Right, I can get unpack to do the job of the regex and pass the basic unit testing, but I can't quite figure out how to handle crazy things like +inf, sorry @dbolser-ebi . unpack is quite dangerous for what it can do if you start putting silly things in. I've seen some bonkers outputs if I give it big strings with peculiar line lengths.

andrewyatz:
How big are we talking? Could we put an upperlimit on it? Say 2^32?

Kieron Taylor wrote:

nerdstrike:
I think we're in trouble if we have 4 gigabases in a single FASTA record. 2^32 will do.

dbolser-ebi:
​http://en.wikipedia.org/wiki/Paris_japonica ? ... (Don't feed the troll).

Good news that unpack is faster. I'm sure any problems of legibility can be
overcome with clear comments.

An alternative fix would be to allow any width between 1 and 120 (the
recommended maximum width), and have a separate option for 'one line per
sequence' i.e. by setting width = 0. As suggested, lines of width > ~ 1k
are just silly anyway.


Cheers,
Dan.

nerdstrike:
I've cooked up an alternative on experimental/fasterfasta

andrewyatz:
Looks pretty much as Dan described. Will be interesting to see how well it works. What's the plan?

dbolser-ebi:
Definitely 'feels' faster. Fails over when width is set to 'unlimited' (2**32).

nerdstrike:
There seems to be a 32-bit limit on the Perl you're using. It probably has a sign bit in there so it doesn't fit. I'll finish this up at the next appropriate juncture.

PR #59, opened 2015-03-23
Updated column definition of genome_statistics.value from INT to BIGINT

arnaudxk:
So we can cope with the size of some plants genomes, beyond max int value (ie 4294967295).


arnaudxk:
Just added a patch file that comes along with the schema change

avullo:
Thanks Arnaud, with the patch the proper test managed to pass successfully. Let's wait the end of the whole travis build.

PR #61, opened 2015-04-02
More flexible attribute output for GFF dumping

james-monkeyshines:
Allow 'special' attributes (e.g. Dbxref, Ontology_term) to be specified as a list and written as a single comma-separated value. (This was already possible for generic attributes with uncapitalized first letters.)


magaliruffier:
Looks good, although I am wondering whether it's possible to extend the existing test case to check for those as well?

james-monkeyshines:
Sure, I can add a test case.

james-monkeyshines:
Hmm, that got a bit ugly, even though I thought I'd rebased correctly. I'll submit a new clean pull request.

PR #62, opened 2015-04-09
When an executable can't be found in the default location, the script at...

thomasmaurel:
...tempt to find it using the UNIX locate command which return a wrong path location. I've updated the locate command to the UNIX find command. This now return the expected executable location (if specified in the PATH variable)

### DIFF
--- a/misc-scripts/CopyDBoverServer.pl
+++ b/misc-scripts/CopyDBoverServer.pl
@@ -311,7 +311,7 @@ foreach my $key ( keys(%executables) ) {
   my $rc     = $? >> 8;
 
   if ( $rc != 0 ) {
-    my $possible_location = `locate -l 1 $key 2>&1`;  # Ugly but simple.
+    my $possible_location = `find -name $key 2>&1`; 
     my $loc_rc            = $? >> 8;
 
     if ( $loc_rc == 0 ) {


nerdstrike:
Agreed, locate requires an updatedb to remain current, find will be lower maintenance.

andrewyatz:
I know I put this code in place. Mostly because of non-standard locations of binaries and the alike. I'm worried a mis-placed `find` will cause a huge lookup of disk.

Quick suggestion. How about dumbing down the program? Still check if we can execute a binary with the expected name but leave it at that. It's up to the executing ENV to get their path correct

danstaines:
> Quick suggestion. How about dumbing down the program? Still check if we
> can execute a binary with the expected name but leave it at that. It's
> up to the executing ENV to get their path correct

+1 to that

mcast:
```
find -name $key
```
This is a bit scary if
* `$PWD eq '/'`(which is a reasonable choice in some cases)
* and you have access to any network filesystem; or worse a few hundred such farm nodes.

thomasmaurel:
Dear all,

Thank you for your comments, I will close the pull request, dumb down the script as suggested by Andy and re-create a pull request.

Cheers,
Thomas
> On 9 Apr 2015, at 13:38, Coveralls <notifications@github.com> wrote:
> 
>  <https://coveralls.io/builds/2299521>
> Coverage increased (+0.28%) to 74.69% when pulling 9b9d0cb <https://github.com/Ensembl/ensembl/commit/9b9d0cbf8e2b8b55b5c87ba3ba8a17872bcdfb1a> on thomasmaurel:master into 349e371 <https://github.com/Ensembl/ensembl/commit/349e371370404e8fb7dff59f0251e89a6a2707e7> on Ensembl:master.
> 
> —
> Reply to this email directly or view it on GitHub <https://github.com/Ensembl/ensembl/pull/62#issuecomment-91217917>.
> 

--
Thomas Maurel
Bioinformatician - Ensembl Production Team
European Bioinformatics Institute (EMBL-EBI)
European Molecular Biology Laboratory
Wellcome Trust Genome Campus
Hinxton
Cambridge CB10 1SD
United Kingdom

PR #66, opened 2015-05-06
Run the rsync command with chmod parameter set to 755 (rwxr-xr-x) to mak...

thomasmaurel:
...e sure we consistently copy our databases with the same permission


magaliruffier:
Thanks for the change, looks fine to me

PR #68, opened 2015-05-10
Can be removed since we are using the new Blast interface

muffato:
Especially, it relies on Bio::Search::HSP::EnsemblHSP which used to be in the web-code but has been removed in e77


nerdstrike:
Laudable, but there are other references in the Core code that would need cleaning up too.

modules/t/canonicalDBAdaptor.t
Bio/EnsEMBL/Utils/ConfigRegistry.pm
Bio/EnsEMBL/Registry.pm
Bio/EnsEMBL/DBSQL/DBAdaptor.pm

Do you want to hunt them down or hand it off to us for tidying?

muffato:
I'll hand it off to you guys :) I only spotted that error because I was checking my setup (trying to compile all the .pm files).

nerdstrike:
Dealt with by us. Thanks for bringing it to our attention.

PR #69, opened 2015-05-18
Retired the UDR parameter since it's twice as slower as plain rsync w…

thomasmaurel:
…hen copying databases between servers with different hardware, you also need to recompile UDR for every new server. I replaced UDR with TCP arcfour (default parameter) which works in all the situations, worst case scenario it will be as fast as plain rsync, in all the other scenarios it will be twice as fast as plain rsync, copy can go up to 110 MB/s or faster


nerdstrike:
Fair enough.

magaliruffier:
fine with me, thanks for the change

PR #70, opened 2015-05-19
Update individual to sample to match the new variation schema

at7:
Replace individual by sample. For the next release variation stores all genotype data on sample level.

### DIFF
--- a/modules/Bio/EnsEMBL/DBSQL/StrainSliceAdaptor.pm
+++ b/modules/Bio/EnsEMBL/DBSQL/StrainSliceAdaptor.pm
@@ -158,20 +158,20 @@ sub fetch_by_name {
     return '';
   }
   
-  # now get an individual adaptor
-  my $ind_adaptor = $variation_db->get_IndividualAdaptor;
+  # now get a sample adaptor
+  my $sample_adaptor = $variation_db->get_SampleAdaptor;
   
   # check we got it
-  unless(defined $ind_adaptor) {
-    warning("Not possible to retrieve IndividualAdaptor from variation database");
+  unless(defined $sample_adaptor) {
+    warning("Not possible to retrieve SampleAdaptor from variation database");
     return '';
   }
   
-  # fetch individual object for this strain name
-  my $ind = shift @{$ind_adaptor->fetch_all_by_name($name)};
+  # fetch sample object for this strain name
+  my $sample = shift @{$sample_adaptor->fetch_all_by_name($name)};
   
   # check we got a result
-  unless(defined $ind) {
+  unless(defined $sample {
     warn("Strain ".$name." not found in the database");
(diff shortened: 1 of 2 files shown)


nerdstrike:
Not your fault, but I'm slightly alarmed by all the "return '';" lines all over the place. I don't know why it's useful to return an empty string unless the calling code has some exotic responses.

Is there any chance of a simple unit test for the changes? It would set my heart at ease to know these instantiate properly.

at7:
- I have a feature branch (variation API) with all the changes needed to work with the new schema. Running unit tests with the changes in StrainSlice and StrainSliceAdaptor pass without errors. 
- Can you run the core tests with the changes I suggest to make sure it doesn't effect the rest of the core API?
- I can replace return ''; by return undef;.

nerdstrike:
return undef has rare but unfortunate consequences. Better to return; than anything else, or just leave it as working with return '';

Travis is happy with the changes to core, I just don't think these modules get any coverage. StrainSlice compiles and that's it. No code is run.

mjg17:
The issue with “return undef” is explained clearly here:
https://metacpan.org/pod/Perl::Critic::Policy::Subroutines::ProhibitExplicitReturnUndef

perlcritic ++

Michael


On 19 May 2015, at 17:05, Kieron Taylor <notifications@github.com> wrote:

> return undef has rare but unfortunate consequences. Better to return; than anything else, or just leave it as working with return '';
> 
> Travis is happy with the changes to core, I just don't think these modules get any coverage. StrainSlice compiles and that's it. No code is run.
> 
> —
> Reply to this email directly or view it on GitHub.
> 




-- 
 The Wellcome Trust Sanger Institute is operated by Genome Research 
 Limited, a charity registered in England with number 1021457 and a 
 company registered in England with number 2742969, whose registered 
 office is 215 Euston Road, London, NW1 2BE.

magaliruffier:
this module should really move out of core and into the variation api
I'll close this pull request until we can safely remove these

at7:
What about AlignStrainSlice.pm? Should the module move into the variation API too? I would prefer to leave StrainSlice and StrainSliceAdaptor with all the updates to match the sample schema in the core API for 81 and then after updating the variation API for 81 investigate which parts of the core API can be moved to the variation API for 82. At the moment it would be too much to do both for 81.

magaliruffier:
That's fair. There have never been any tests for any of these modules in the core API, so we just have to hope the tests in the variation repo cover what needs to.
If you need the changes in for 81, we can pull them in

at7:
Yes please merge the updates into the core master branch. I already updated the variation API with the new schema and only tests that use StrainSlice etc are failing at the moment. So they should pass after the merge. We will look into moving variation related modules from core into variation for 82.

PR #71, opened 2015-06-18
Switchable adaptor support

andrewyatz:
This branch brings in changes to the Registry to support a very basic implementation of dependency injection/inversion of control container. You can switch an adaptor for an alternative implementation using the Registry directly where you are responsible for switching the implementation back after finishing with the code.

Alternatively you can use a method on the DBAdaptor instance which takes in a callback subroutine and will deal with the nitty gritty of switching implementations back and forth.

Main use case is to switch where DNA sequence comes from.

### DIFF
--- a/modules/Bio/EnsEMBL/DBSQL/DBAdaptor.pm
+++ b/modules/Bio/EnsEMBL/DBSQL/DBAdaptor.pm
@@ -70,10 +70,12 @@ use Bio::EnsEMBL::DBSQL::BaseFeatureAdaptor;
 use Bio::EnsEMBL::Utils::SeqRegionCache;
 use Bio::EnsEMBL::Utils::Exception qw(throw warning deprecate);
 use Bio::EnsEMBL::Utils::Argument qw(rearrange);
-use Bio::EnsEMBL::Registry;
+use Bio::EnsEMBL::Utils::Scalar qw(check_ref scope_guard);
 use Bio::EnsEMBL::Utils::ConfigRegistry;
 
-my $registry = "Bio::EnsEMBL::Registry";
+
+my $REGISTRY = "Bio::EnsEMBL::Registry";
+require Bio::EnsEMBL::Registry;
 
 =head2 new
 
@@ -162,7 +164,6 @@ sub new {
   if ( defined($species) ) { $self->species($species); } 
   if ( defined($group) )   { $self->group($group) }
 
- 
   $self = Bio::EnsEMBL::Utils::ConfigRegistry::gen_load($self);
 
   if (defined $species_id) {
@@ -197,7 +198,7 @@ sub new {
 
 sub clear_caches {
   my ($self) = @_;
-  my $adaptors = Bio::EnsEMBL::Registry->get_all_adaptors(
+  my $adaptors = $REGISTRY->get_all_adaptors(
     $self->species(), $self->group());
   foreach my $adaptor (@{$adaptors}) {
     if($adaptor->can('clear_cache')) {
@@ -287,8 +288,6 @@ sub dbc{
   return $self->{_dbc};
 }
 
-
-
 =head2 add_db_adaptor
 
   Arg [1]    : string $name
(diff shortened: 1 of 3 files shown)


modules/Bio/EnsEMBL/DBSQL/DBAdaptor.pm
```diff
@@ -1051,6 +1024,88 @@ sub to_hash {
...
+=head2 switch_adaptor
+
+  Arg [1]     : String name of the adaptor type to switch out
+  Arg [2]     : Reference The switchable adaptor implementation
+  Arg [3]     : (optional) CodeRef Callback which provides automatic switchable adaptor cleanup 
+                once the method has finished
+  Arg [4]     : (optional) Boolean override any existing switchable adaptor
+  Example     : $dba->switch_adaptor("seqeunce", $my_replacement_sequence_adaptor);
+                $dba->switch_adaptor("seqeunce", $my_other_replacement_sequence_adaptor, 1);
+                $dba->switch_adaptor("seqeunce", $my_replacement_sequence_adaptor, sub {
```
magaliruffier: shouldn't that be 'sequence'?

modules/Bio/EnsEMBL/DBSQL/DBAdaptor.pm
```diff
@@ -1051,6 +1024,88 @@ sub to_hash {
...
+    my $guard = scope_guard(sub { $REGISTRY->remove_switchable_adaptor($species, $group, $adaptor_name); } );
+    $callback->();
+  }
+  return;
+}
+
+=head2 has_switched_adaptor
+
+  Arg [1]     : String name of the adaptor type to switch back in
+  Example     : $dba->has_switchable_adaptor("seqeunce"); #explicit switching back
```
magaliruffier: 'sequence' again

modules/Bio/EnsEMBL/DBSQL/DBAdaptor.pm
```diff
@@ -1051,6 +1024,88 @@ sub to_hash {
...
+sub has_switched_adaptor {
+  my ($self, $adaptor_name) = @_;
+  return $REGISTRY->has_switchable_adaptor($self->species, $self->group, $adaptor_name);
+}
+
+=head2 revert_adaptor
+
+  Arg [1]     : (optional) String name of the adaptor type to switch back in
+  Example     : $dba->revert_adaptor(); #implicit switching back whatever was the last switch_adaptor() call
+                $dba->revert_adaptor("seqeunce"); #explicit switching back
```
magaliruffier: 'sequence'

andrewyatz:
It's almost as if I was using autocomplete ;)

magaliruffier:
the changes here look relatively harmless on their own
do we have an example on how the VEP REST code would look like with this?
https://github.com/Ensembl/ensembl-rest/blob/release/80/lib/EnsEMBL/REST/Controller/VEP.pm

andrewyatz:
Think it all changes on line 155 in REST VEP.pm. So we'd need to change the controller to something like

```perl
my $fasta_index = $self->get_fasta_index();
$dba->switch_adaptor('sequence', $fasta_index, sub {
  $consequences = $self->get_consequences($c, $config, \@vfs);
});
```

Of course how you get the fasta index, do it for multiple species and Rishi's code still needs to be in place.

Magali Ruffier wrote:

modules/Bio/EnsEMBL/DBSQL/DBAdaptor.pm
```diff
@@ -1051,6 +1024,88 @@ sub to_hash {
...
 
+#########################
+# Switchable adaptor methods
+#########################
+
+=head2 switch_adaptor
+
+  Arg [1]     : String name of the adaptor type to switch out
+  Arg [2]     : Reference The switchable adaptor implementation
+  Arg [3]     : (optional) CodeRef Callback which provides automatic switchable adaptor cleanup
```
nerdstrike: Does this mean the callback has to do cleanup, or that cleanup is done after the callback is run?

andrewyatz: Cleanup is done after the callback has executed.

Kieron Taylor wrote:

PR #72, opened 2015-06-18
Experimental/alternative sequence stores

andrewyatz:
This code brings with it support for switchable adaptors. Might cause a merge conflict. Please merge #71 first.

The code has divided SequenceAdaptor into two objects.

- SequenceAdaptor: communicates with the database to retrieve sequence
- BaseSequenceAdaptor: has the ability to cache sequence in-between calls making the operation potentially more efficient

There are also some additional objects called FastaSequenceAdaptor and Faidx support. These are two example alternative stores. Faidx implements the indexing scheme from samtools/htslib to provide fast seekable access into FASTA files. The FastaSequenceAdaptor expects an instance of this object and delegates calls for sequence onto it.


nerdstrike:
The only thing I'm not quite clear on here is: Do seq-region-edits get applied to the stuff in the FASTA files or only from the database? I would imagine they shouldn't affect static dumps, but I can't tell if they'll trigger or not.

andrewyatz:
Sequence edits should already be in the fasta files. I think any other way could result in more confusion.

Kieron Taylor wrote:

nerdstrike:
Given that everything works as intended, I don't see any gaping holes in the logic. I think this is sound.

andrewyatz:
Woohoooo. Sound logic!!!!

Kieron Taylor wrote:

magaliruffier:
the remove method has not been conserved
I am not convinced it was used before, but was it done on purpose?

andrewyatz:
Remove is back in. Think it was a mistake

magaliruffier:
this has now been cross-checked by the entirety of the team, pulling the changes in

magaliruffier:
found some possible memory leak issues when dumping fasta sequences
reverting changes until I can find out what is happening

PR #73, opened 2015-07-10
changes made to handle multiple provider name

ckongEbi:


nerdstrike:
Poor literary style using many ands, but I guess it works for GFF.
There is a danger of it choking if it doesn't get any providers back.

danstaines:
not keen on using "and" as a separator - ; or , would be better I feel. Also, the |'' bit should probably be |[] though I much prefer an explicit if(defined && scalar>0) rather than a binary statement, but that might be too verbose for some.

james-monkeyshines:
The problem with ';' and ',' as separators is that they have special meaning in GFF; so they would have to be escaped... It's pretty uncommon to have more than one provider (I think?) and if there are multiple providers, then there will probably be two of them. So " and " isn't a good separator in general, but here I think it works.

danstaines:
where does this go in GFF anyway? Would seem overkill to be on each feature!

magaliruffier:
it is just part of the meta data in the header

danstaines:
So, do ; and ' have special meaning in the header?

james-monkeyshines:
No, metadata is just text; I was thinking of attributes in my earlier comment, so that's not pertinent.

magaliruffier:
I'm happy to take the changes in, as it does not change anything for Ensembl species where we don't have any providers
would be nice if you could come to a consensus as to what you'd like to have for eg though

danstaines:
My inclination is ';' as a separator as this is in the header, but thats 
just my view, not an authorative statement.

danstaines:
@ckongEbi can you make the appropriate changes in your repo for the pull request?

ckongEbi:
Sure, i can do that.

PR #74, opened 2015-07-17
Feature/genepredbed

andrewyatz:
This was some code to support the UCSC format extension to BED to support Gene predictions. Details of the format is here:

http://genome.ucsc.edu/goldenPath/help/bigGenePred.html

Some example data from here:

http://genome.ucsc.edu/goldenPath/help/examples/bigGenePred.txt

This data was GRCh37 data so you have to go back to release 76 or the 37 archive server to get the right data out. As far as I can tell we are doing the right thing. It also fixes a number of issues with incorrect phase representation. It seems the convention is to map them as -1 if they are non-coding and 0 if they are a starting coding exon. We seem to store these as -1 so that wasn't quite in sync with the way UCSC work.

### DIFF
--- a/modules/Bio/EnsEMBL/Utils/IO/BEDSerializer.pm
+++ b/modules/Bio/EnsEMBL/Utils/IO/BEDSerializer.pm
@@ -47,6 +47,7 @@ use base qw(Bio::EnsEMBL::Utils::IO::FeatureSerializer);
     Arg [1]    : Ontology Adaptor
     Arg [2]    : Optional File handle
     Arg [3]    : Default source of the features. Defaults to .
+    Arg [4]    : RBG colour to emit. Defaults to black (0,0,0)
 
     Returntype : Bio::EnsEMBL::Utils::IO::GFFSerializer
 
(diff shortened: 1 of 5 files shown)


modules/Bio/EnsEMBL/Utils/IO/BEDSerializer.pm
```diff
@@ -118,11 +140,29 @@ sub _write_Transcript {
...
+      $cds_end_status = 'incmpl' if $three_prime_nc;
+    }
   }
   else {
     # apparently looking at UCSC's own BED output formats we do not need to bother
     # coverting $coding_start into 0 based coords for this one ... odd
-    $coding_start = $new_transcript->seq_region_end();
+    $coding_start = $new_transcript->seq_region_start();
+    $coding_start--;
     $coding_end = $coding_start;
```
magaliruffier: does that mean we do have to convert to 0-based coordinates?

magaliruffier:
does it need to be merged on release branch or should that be master instead?

modules/Bio/EnsEMBL/Utils/IO/BEDSerializer.pm
andrewyatz: That comment might be misleading now looking at some other outputs ucsc has. The format is ambiguous as to what a column should contain beyond it being the data point in question.

andrewyatz:
Should be master :)

andrewyatz:
Closing down this pull request to reopen against master

PR #75, opened 2015-07-20
increasing width of xref dbprimary_acc and display_label

danstaines:
As discussed with Mag et al. recently, to ensure schema compatibility with EG


magaliruffier:
do you need this for release 81?
otherwise the patch file would be patch_81_82_b.sql

danstaines:
Whoops, didn't notice you'd not created a patch file for 81_82 yet. Probably cleanest to wait until you've created patch_81_82_a.sql and then resubmit.

PR #77, opened 2015-07-20
xref field width patch

danstaines:
increased xref dbprimary_acc/display_label field width to support very long identifier strings


magaliruffier:
You need to patch the test databases, otherwise the test suite won't pass
We have some guidelines here: http://www.ebi.ac.uk/seqdb/confluence/display/ENSCORE/Patching+test+databases
and I will update the contribution guide for those cases
You also need to update the meta table in the table.sql file so it is in sync with the patch files
http://www.ebi.ac.uk/seqdb/confluence/display/ENSCORE/Core+release+coordination#Corereleasecoordination-Schemachanges

danstaines:
One of these days I'll remember the meta table... Thanks for the tip about patching the test dbs, now done as well.

modules/t/test-genome-DBs/circ/core/xref.txt
```diff
@@ -128,13 +128,13 @@
 88581	50728	CP003765.1:CDS:complement(234076..235323)	CP003765.1:CDS:complement(234076..235323)	0	\N	NONE	
 88582	50727	CP003765.1:CDS:complement(234076..235323)	CP003765.1:CDS:complement(234076..235323)	0	\N	NONE	
 88583	50719	BTF1_31777 1-1	BTF1_31777-1	0	\N	NONE	
-88584	50726	CP003765.1:CDS:complement(join(235337..235425,1..685))	CP003765.1:CDS:complement(join(235337..235425,1..685))	0	\N	NONE	
+88584	50726	CP003765.1:CDS:complement(join(235337..235425,1..6	CP003765.1:CDS:complement(join(235337..235425,1..685))	0	\N	NONE
```
magaliruffier: funny, looks like our test data had some accessions which were already way too long for the table definition

magaliruffier:
much better now, thanks a lot

PR #78, opened 2015-08-07
Fixed fetch_all_by_Slice_and_external_dbname_link

simonbrent:
$self->get_DBEntryAdaptor() doesn't exist

### DIFF
--- a/modules/Bio/EnsEMBL/DBSQL/GeneAdaptor.pm
+++ b/modules/Bio/EnsEMBL/DBSQL/GeneAdaptor.pm
@@ -549,7 +549,7 @@ sub fetch_all_by_Slice_and_external_dbname_link {
   my ($self, $slice, $logic_name, $load_transcripts, $db_name) = @_;
 
   # Get the external_db_id(s) from the name.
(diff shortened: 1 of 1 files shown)


magaliruffier:
thanks for the fix, goes to show how much we've been using that method...

simonbrent:
Yeah, well, I tried to use it because I wanted all genes with HGNC names, but even after fixing it looked like it would be MUCH slower than just fetching all genes and then grepping for $_->display_xref->dbname eq 'HGNC'

magaliruffier:
do you need it slice by slice? or could you just use the 'list_gene_ids_by_external_db_id' directly?
either way, we'll look into the performance issues as well

simonbrent:
list_gene_ids_by_external_db_id give back just a list of gene dbIDs, so I'd still have to get all the genes after that and check that their dbIDs are in that list, so the speed is about the same either way.

Also, I only want genes on 1-22, X, Y, MT, so I have to check that and throw the rest away.

nerdstrike:
Seems sane. Thanks Simon

PR #79, opened 2015-08-10
fastmap expects seq region name not id

nicklangridge:
This is broken in earlier versions also (I checked 80 and 81) - is it possible to have the fix applied to least to 81 as well as master?


nerdstrike:
That's a sane request. No objections here.

magaliruffier:
change has been added to release/81 as well
I suspect it has been wrong since it was introduced (https://github.com/Ensembl/ensembl/blame/fc6ced683c9b544127e6f50711ba833b430baf88/modules/Bio/EnsEMBL/DBSQL/BaseFeatureAdaptor.pm#L656) because there is no test case for it
Do you need it on other branches?

nicklangridge:
Thanks.
If it is not too much trouble it would be useful to have it on release/80, but it's not critical.

magaliruffier:
added onto release/80 as well

nicklangridge:
Thanks

PR #81, opened 2015-08-11
Need to clean-up the additional lookups too

muffato:
Without this line, the additional lookups keep links to the previous object

### DIFF
--- a/modules/Bio/EnsEMBL/DBSQL/Support/FullIdCache.pm
+++ b/modules/Bio/EnsEMBL/DBSQL/Support/FullIdCache.pm
@@ -85,6 +85,7 @@ sub put {
   my ($self, $key, $object) = @_;
   my $old = $self->SUPER::put($key, $object);
   #Add to additional lookup
+  $self->remove_from_additional_lookup($key, $old) if $old;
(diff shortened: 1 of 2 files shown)


magaliruffier:
I can't see any major issue with it, I am just unsure where this code is actually used
Do you have any examples?

muffato:
A few Compara adaptors are using the FullIdCache. I don't understand why but the Compara was inserting adding an object twice to the cache, and the error was that get_by_additional_lookup() would fail since the additional lookups contain two references to the same object.

muffato:
I think I can add a test to fullIdCaching.t to show the problem

muffato:
I have added a test to fullIdCaching.t
If you run the amended test on the original version of FullIdCache.pm, it will fail

magaliruffier:
that would be much appreciated

muffato:
Last time I checked, Compara was indeed the sole user of the ID-caching mechanism

nerdstrike:
I think the changes are ok, but the generally legibility of the FullIdCache is quite poor. This type of functionality is tough to debug at the best of times, and it's just not self-explanatory.

PR #82, opened 2015-09-01
Update GTFSerializer.pm

amonida:
Skipping trans-splicing transcripts and the gene if it's a single transcript gene.


avullo:
Different calls to get_all_Transcripts can be optimised into a single one.

PR #83, opened 2015-09-01
Update GTFSerializer.pm

amonida:
Skipping trans-splicing transcripts and the gene if it's a single transcript gene with optimisation call.


avullo:
This is now good, thanks!

PR #84, opened 2015-09-02
Die if there is no available config file in Registry::load_all()

muffato:
Hi Core,

Registry::load_all by default does not report any error if the config file is missing / if the file does not exist. It will only print a message if $verbose is switched on.
We have scripts and Runnables that load the Registry and they basically have to duplicate the tests "if (defined ...)" and "if (-e ...)" to be able to output meaningful messages.

I propose that load_all() actually die if $config_file is not given / points to an unexisting file.

It passes registry.t,  I hope it passes the other tests too !

### DIFF
--- a/modules/Bio/EnsEMBL/Registry.pm
+++ b/modules/Bio/EnsEMBL/Registry.pm
@@ -202,23 +202,39 @@ my %group2adaptor = (
                the database if not used properly.  If in doubt, do
                not use it or ask in the developer mailing list.
 
+  Arg [5]:     (optional) boolean
+               This option will make load_all() throw if the configuration file
+               is missing and cannot be guessed from the environment
+
   Example    : Bio::EnsEMBL::Registry->load_all();
   Returntype : Int count of the DBAdaptor instances which can be found in the 
                registry due to this method being called. Will never be negative
-  Exceptions : none
+  Exceptions : Throws if $throw_if_missing is set and ($config_file is missing
+               and cannot be guessed from the environment
   Status     : Stable
 
 =cut
 
 sub load_all {
-    my ($class, $config_file, $verbose, $no_clear, $no_cache ) = @_;
+    my ($class, $config_file, $verbose, $no_clear, $no_cache, $throw_if_missing ) = @_;
 
     if ( !defined($config_file) ) {
       if ( defined( $ENV{ENSEMBL_REGISTRY} ) ) {
-        $config_file = $ENV{ENSEMBL_REGISTRY};
+        if (-e $ENV{ENSEMBL_REGISTRY}) {
+          $config_file = $ENV{ENSEMBL_REGISTRY};
+        } else {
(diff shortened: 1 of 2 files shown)


andrewyatz:
I'm a little against this in its current form. A number of pipelines I know I saw before relied on being able to pass in a configuration file that may not exist and assume the code will silently ignore the error. If we are going to start getting the code throwing an exception by default could we still enable the old interface by passing a `no_throw` flag through?

muffato:
I can see it would help those pipelines / scripts that rely on the current behaviour, but can some of them be rewritten to not call load_all() if the file is not defined ? (just asking)

andrewyatz:
And all of those external pipelines/scripts we don't know about? Considering the main use case of giving a registry file ignores that the method offers functionality to load from two alternative sources. A user shouldn't have to check 3 locations, two of which require prior knowledge, to know if this method is going to fail.

muffato:
I agree a script shouldn't test the three locations, and I wasn't suggesting that. I was thinking of a script that wants to load a registry file if given by the user. It could do:
```perl
$reg->load_all($config_file) if defined $config_file;
```
or
```perl
$reg->load_all($config_file, -NO_THROW=>1);
```
It is just a matter of taste, but I find the first version clearer.


But yes, there is still the question of backwards compatibility that has to be addressed. The change I am proposing can potentially break many scripts / pipelines and they can't all be changed. This is a decision for Core@

magaliruffier:
the Registry is too crucial to the rest of the API, we cannot introduce such a breaking change
the safest option would be to add a throw_if_missing flag
that way, your pipelines can die if the file is missing, but existing pipelines will keep on running as before

muffato:
All right, what about this ?

magaliruffier:
if that provides you with the functionality you need, the change is good for me

muffato:
I've found in Compara 5 Runnables and 42 scripts that call load_all() with a file and most of them require the file to be defined and existing, so I would definitely be happy to have a proper exception if the file is not right. I will add the `throw_if_missing` option to those calls

modules/Bio/EnsEMBL/Registry.pm
```diff
@@ -202,23 +202,39 @@ my %group2adaptor = (
...
-        $config_file = $ENV{HOME} . "/.ensembl_init";
+        if (-e ($ENV{HOME} . "/.ensembl_init")) {
+          $config_file = $ENV{HOME} . "/.ensembl_init";
+        }
+      }
+      if ($throw_if_missing and !defined($config_file) ) {
+        throw("No registry configuration to load, and no default could be guessed.\n");
       }
+    } elsif ($throw_if_missing and !(-e $config_file)) {
+      throw(printf("Configuration file '%s' does not exist. Registry configuration not loaded.\n", $config_file ));
```
avullo: To be 100% sure the following operations succeed, I would also check the file is readable (-r) wherever the file is checked for existence.

PR #86, opened 2015-09-04
Experimental/merged alternative sequence store

magaliruffier:
the alternative_sequence_store branch has been successfully tested with the fasta dumping pipeline and the xref pipeline, the updated clear_cache method seems to behave
I now need a sanity check that reverting the revert commit does the right thing and we have not lost anything

### DIFF
--- a/modules/Bio/EnsEMBL/DBSQL/FastaSequenceAdaptor.pm
+++ b/modules/Bio/EnsEMBL/DBSQL/FastaSequenceAdaptor.pm
@@ -1,0 +1,155 @@
+=head1 LICENSE
+
+Copyright [1999-2014] Wellcome Trust Sanger Institute and the EMBL-European Bioinformatics Institute
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+=head1 CONTACT
+
+  Please email comments or questions to the public Ensembl
+  developers list at <dev@ensembl.org>.
+
+  Questions may also be sent to the Ensembl help desk at
+  <helpdesk@ensembl.org>.
+
+=cut
+
+=head1 NAME
+
+Bio::EnsEMBL::DBSQL::FastaSequenceAdaptor
+
+=head1 DESCRIPTION
+
+This sequence adaptor extends the BaseSequenceAdaptor code to provide an 
+implementation of the .fai index lookup as defined by samtools. The code uses
+this indexing system to access portions of sequence and translates Slice requests
+into sensible locations for our FASTA query layer.
+
+The adaptor must be initalised with access to a Faidx compatible object and the FASTA
+file backing must use the same seq_region_name as the querying slices otherwise
+we cannot return the required data.
+
+=cut
+
+package Bio::EnsEMBL::DBSQL::FastaSequenceAdaptor;
+
+use strict;
+use warnings;
+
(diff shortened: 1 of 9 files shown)


modules/Bio/EnsEMBL/Utils/IO/FileFaidx.pm
```diff
@@ -0,0 +1,469 @@
...
+sub file {
+  my ($self, $file) = @_;
+  if(defined $file) {
+    throw "No file found at '${file}'" unless -f $file;
+    $self->{'file'} = $file;
+  }
+  return $self->{'file'};
+}
+
+=head2 store
```
avullo: should be 'write_index_to_disk

modules/Bio/EnsEMBL/DBSQL/FastaSequenceAdaptor.pm
```diff
@@ -0,0 +1,155 @@
...
+=head2 can_access_Slice
+
+  Description : Checks the lookup to see if we have access to the Slice given (using 
+                seq region name as the ID). We reject any Circular Slice
+
+=cut
+
+sub can_access_Slice {
+  my ($self, $slice) = @_;
+  return 0 if $slice->is_circular();
```
avullo: I would rather throw in case of circular slice.

modules/Bio/EnsEMBL/DBSQL/SequenceAdaptor.pm
```diff
@@ -80,64 +69,14 @@ our $SEQ_CACHE_MAX   = (2 ** $SEQ_CHUNK_PWR) * $SEQ_CACHE_SZ;
 =cut
 
 sub new {
-  my $caller = shift;
-
+  my ($caller, $db, $chunk_power, $cache_size) = shift;
```
avullo: Yes, you initialise $caller and declare the rest in one shot, but I don't see where the rest of the array variables are initialised.

avullo:
Ok, it makes sense to me, added a few comments but nothing dramatically serious.
Had a look at DBAdaptor and Registry to figure out the logic behind switchable adaptors.

modules/Bio/EnsEMBL/DBSQL/FastaSequenceAdaptor.pm
andrewyatz: The idea is for the sequence adaptors to say can they handle the given slice. I viewed the adaptors acting like Decorators and introducing another object which could use this method to figure out which adaptor can provide the required sequence. In that case giving it a CircularSlice is not an erroneous situation and should be left to the calling class to decide if it is a bad thing it cannot access the Slice

modules/Bio/EnsEMBL/DBSQL/SequenceAdaptor.pm
andrewyatz: Oh. That's a bug surely. Should be assigned to the stack so we can set all of these values. Can we change it?

modules/Bio/EnsEMBL/Utils/IO/FileFaidx.pm
andrewyatz: Ahh yeah!

andrewyatz:
Okay I've done those changes as suggested apart from throwing an exception for CircularSlices as it isn't my intended behaviour.

modules/Bio/EnsEMBL/DBSQL/FastaSequenceAdaptor.pm
avullo: Ok, I see your point. But the caller in this case doesn't know the reasons why the slice cannot be accessed; an exception would force it to become aware. Anyway, just a thought ;)

andrewyatz: I think my intention was so the caller didn't have to know why it was rejected :)

avullo: All right, so if that was the intention I have nothing to say ;)

magaliruffier:
all changes added, thanks Andy

PR #87, opened 2015-09-04
Specify 5' or 3' UTR in the GTF file

james-monkeyshines:
Specify 5' or 3' UTR in the GTF file


magaliruffier:
I was looking into adding this myself, so am in theory happy with the change
can you just update the test data so it matches the new format?

magaliruffier:
thanks for the test patches, looks good now

PR #88, opened 2015-09-09
Master

james-monkeyshines:
I hope somebody somewhere appreciates all this work to cater for trans-spliced transcripts... This change is required because CDS features within a trans-spliced transcript can have different strands (or different seq region names, although that's not the case with fruit fly). The CDS ID is ordinarily the translation stable ID, which implies that all the CDS rows are a single feature; which cannot therefore have multiple strands...


magaliruffier:
what happens to the value field if it is not a trans_spliced transcript?
if I read this correctly, CDS objects will never reach the last else, and value will not get the so_term concatenation

james-monkeyshines:
Ah yes, you're quite right. I'll fix that...

magaliruffier:
looks fine now

PR #89, opened 2015-10-13
Master

khowe:


magaliruffier:
thanks for the fix on the array express parser
the rest is wormbase specific, so you've probably tested it better than us ;-)
changes can go in as far as I'm concerned

avullo:
I've seen you've added the method gene_description_sources but can't find it called in your commits.

magaliruffier:
the gene_description_sources method is called in the DisplayXrefs module, but it can be overwritten by species-specific mappers

avullo:
I see, thanks.

PR #90, opened 2015-10-20
Exclude ncbi_taxonomy from db adaptor list

nicklangridge:


nicklangridge:
If this could also go into the 82 branch that would be great, thanks.

PR #91, opened 2015-10-20
vport and cport options are now used in the code

ens-carlos:


nerdstrike:
That seems wholly sensible.

PR #92, opened 2015-10-20
Add new DBAdaptor

at7:
Add new DBAdaptor used by the ensembl-gene2phenotype API


nerdstrike:
Do you want the gen2phen adaptor to be available via load_registry_from_db? If so, I suspect an appropriate code block would need to be added.

at7:
using load_registry_from_db is not planned at the moment.

nerdstrike:
Then I see no problem.

PR #97, opened 2015-11-11
added taxonomy adaptor support

danstaines:
Added support for including ncbi_taxonomy databases using the new ensembl-taxonomy API

### DIFF
--- a/modules/Bio/EnsEMBL/Registry.pm
+++ b/modules/Bio/EnsEMBL/Registry.pm
@@ -162,6 +162,7 @@ my %group2adaptor = (
       'pipeline'      => 'Bio::EnsEMBL::Pipeline::DBSQL::DBAdaptor',
       'production' => 'Bio::EnsEMBL::Production::DBSQL::DBAdaptor',
       'stable_ids' => 'Bio::EnsEMBL::DBSQL::DBAdaptor',
+      'taxonomy'  => 'Bio::EnsEMBL::DBSQL::TaxonomyDBAdaptor',
       'variation' => 'Bio::EnsEMBL::Variation::DBSQL::DBAdaptor',
       'vega'      => 'Bio::EnsEMBL::DBSQL::DBAdaptor',
       'vega_update' => 'Bio::EnsEMBL::DBSQL::DBAdaptor',
@@ -374,7 +375,6 @@ sub load_all {
                     $adaptor_args{'-no_cache'} = 1;
                   }
                 }
-
                 if ($verbose) {
                     printf( "Configuring adaptor '%s' "
                               . "for configuration section '%s'...\n",
@@ -1734,6 +1734,8 @@ sub load_registry_from_db {
   my $ontology_db;
   my $ontology_version;
 
+  my $taxonomy_db;
+
   my $production_dba_ok = 
     eval { require Bio::EnsEMBL::Production::DBSQL::DBAdaptor; 1 };
   my $production_db;
@@ -1804,6 +1806,8 @@ sub load_registry_from_db {
         $ontology_db      = $db;
         $ontology_version = $1;
       }
+    } elsif ( $db =~ /^ncbi_taxonomy$/ ) {
+        $taxonomy_db      = $db;
     } elsif ( $production_dba_ok and $db =~ /^ensembl(?:genomes)?_production(_\d+)?/x ) {
       # production db can come with no version (i.e. that on ens-staging1),
       # but it's backed up with a release number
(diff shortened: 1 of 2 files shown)


magaliruffier:
where are the test cases? ;-)

muffato:
And would we still able to create a Compara adaptor on that database ?
Our production registry file loads everything from staging1 and stagin2, and then create a few DBAdaptors on manually listed databases. We use that during the release and one of them is the ncbi_taxonomy database
https://github.com/Ensembl/ensembl-compara/blob/release/82/scripts/pipeline/production_reg_conf.pl#L147

Matthieu

danstaines:
thats actually a good point. There are tons of test cases for this API 
itself in ensembl-taxonomy - but having a test for taxonomy in core 
would mean that core would need to check out taxonomy as well and have a 
test taxonomy db. Is this a dependency you'd want, or are you happy with 
this being tested in the ensembl-taxonomy repo?

muffato:
Can MultiTestDB create an empty database ?
Perhaps the test should not import the repo, and test that the API dies with the message "Bio::EnsEMBL::DBSQL::TaxonomyDBAdaptor not found in @INC) :D

danstaines:
Good question, I'll look into this. If you're manually adding the DB to 
the registry, the species and group are different, I don't see why not.

danstaines:
Your suggestion of an fail is not a bad one I guess - I'll work on the 
registry tests in ensembl-taxonomy in any case and then come back to this.

danstaines:
Actually, testing the registry seems to be a big problem, where you need 
to test whether a particular db is picked up and added by one of the 
load methods, as the name is crucial. In my case, my little test NCBI 
database gets called dstaines_test_db_multi_taxonomy_20151112_111753 
which won't get matched with by the registry. I think the only way to 
test the registry is to load our test server using predefined names and 
make sure the registry finds the database in question. Is this something 
that MultiTestDB can do easily?

magaliruffier:
should the taxonomy db not be added to the list of known databases anyway?
I am thinking about this section:
https://github.com/Ensembl/ensembl/blob/release/82/modules/Bio/EnsEMBL/Registry.pm#L1780
where all 'special' databases are meant to be picked up
We are currently not dealing with the taxonomy database there, which has caused problems in the past as it gets included along with the core-like databases
This means we add to explicitly exclude it
https://github.com/Ensembl/ensembl/blob/master/modules/Bio/EnsEMBL/Registry.pm#L3012
If we pick it up looking for 'taxonomy', would that also work in your use case?

danstaines:
Its in that list already, unless I'm missing something:
https://github.com/danstaines/ensembl/blob/master/modules/Bio/EnsEMBL/Registry.pm#L1809

I've followed a similar approach to ontology, but without the 
versioning. AFAIK, we don't need it in get_species_and_object_type.

Am happy to make any changes you suggest though. The main problem is 
still testing, but I think thats something generally applicable to much 
of the registry code as far as I can tell.

magaliruffier:
Oops, yes, my bad.
What happens if you allow a prefix for the db name?

danstaines:
Thats not supported with my code change, mainly since production and 
ontology don't support it as far as I can tell (these were my models for 
taxonomy). I can add it if we really think we need it, but I can't think 
of a scenario where we would.

andrewyatz:
The registry is fairly untestable at the moment and because of our unique test db naming scheme I can't see an easy way to do this unless we fancy injecting a fake db handle and responding with a result list of dbs we expect it to handle/process

danstaines:
Given what Andy said, about the best I can do is to point at the public 
EG server and make sure I get a tax adaptor back from el Reg. I've added 
this to ensembl-taxonomy repo (which currently has one other test using 
the full size public ncbi_taxonomy db), so would you be happy for the 
pull req go through? I've tested the scenarios that Matthieu was 
concerned about.

lairdm:
Continuing my theme of making mock testing servers and coming in with
minimal context for the issue at hand, I just thought I'd leave this
sitting here and wander away...

https://metacpan.org/pod/DBD::Mock

Create a mock driver, program it with the expected results for queries,
profit.

danstaines:
On 12/11/15 13:16, Matthew Laird wrote:
> Continuing my theme of making mock testing servers and coming in with
> minimal context for the issue at hand, I just thought I'd leave this
> sitting here and wander away...
>
> https://metacpan.org/pod/DBD::Mock
>
> Create a mock driver, program it with the expected results for queries,
> profit.

This kind of approach is definitely something to look at, I agree, 
paritcularly where you have complicated logic like this. As always the 
devil is in the details - for instance, I'm not sure how Mock deals with 
queries on multiple schemas (presumably with distinct bits of SQL that 
you hard code). Also the registry has the mysql hardcoded in it (less of 
a big deal, but still).

Something I feel I need to leave with my core colleagues for further 
consideration, and also wander away.

-- 
Dan Staines, PhD
Genomics Technology Infrastructure Coordinator
EMBL-EBI, Wellcome Trust Genome Campus
Cambridge CB10 1SD, UK
Tel: +44-(0)1223-492507

PR #100, opened 2015-11-20
Fix for sub_Slice when slice->start is <1

willmclaren:
If a slice has start<1, then it is valid to request an end that is
&gt; slice->end as long as it is <=slice->length, since sub_Slice args
are relative to the original slice.

Please do correct me if I'm wrong here! I am thoroughly confused by the concept of slices with start<1 to begin with :-)

### DIFF
--- a/modules/Bio/EnsEMBL/Slice.pm
+++ b/modules/Bio/EnsEMBL/Slice.pm
@@ -1270,12 +1270,14 @@ sub constrain_to_seq_region {
 sub sub_Slice {
   my ( $self, $start, $end, $strand ) = @_;
 
-  if( $start < 1 || $start > $self->{'end'} ) {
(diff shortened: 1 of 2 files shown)


magaliruffier:
the strainSlice tests are failing on this change, might need to check what is happening there

magaliruffier:
is there a plan to fix this PR or should I just close it until Anja's changes come in?

willmclaren:
Will take a look. Might as well fix it here, at least then we will know where it's going to break when Anja moves the tests :-)

willmclaren:
All fixed now

magaliruffier:
Travis is happy now, thanks

nerdstrike:
We should really get rid of all the return undefs. One day they'll bite someone.

PR #101, opened 2015-11-23
fix for VariationFeatures using a fake adaptor on circular slices

willmclaren:
Should work to fix @nicklangridge 's problems

### DIFF
--- a/modules/Bio/EnsEMBL/Feature.pm
+++ b/modules/Bio/EnsEMBL/Feature.pm
@@ -1100,7 +1100,7 @@ sub seq_region_start {
   if ( defined($slice) ) {
 
     return $self->_seq_region_boundary_from_db('start')
-      if $slice->is_circular() and $self->adaptor() && $self->adaptor->dbc;
+      if $slice->is_circular() and $self->adaptor->dbc;
 
     my $start;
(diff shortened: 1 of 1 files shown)


avullo:
I don't think you need to call $self->adaptor twice, suffice the call to the dbc submethod.

nicklangridge:
Thanks Will.
Looks like the same check should be applied at this line https://github.com/Ensembl/ensembl/blob/release/82/modules/Bio/EnsEMBL/Feature.pm#L1145

willmclaren:
I've update this PR with the suggested changes.

magaliruffier:
thanks for the updates, looks fine now

nicklangridge:
> I don't think you need to call $self->adaptor twice, suffice the call to the dbc submethod.

It turns out we do need the call to $self->adaptor

nicklangridge:
Should I submit another PR?

magaliruffier:
yes please
will you need this on release/83 branch as well?

nicklangridge:
yes we will - do you want a PR on each branch?

magaliruffier:
one will be fine, I'll just rebase the change

PR #104, opened 2015-11-30
To use dbprimary_acc instead of display_id;Fix GO source duplication

ckongEbi:


magaliruffier:
IIRC, this is not needed for this release, is that right?
changes look fine, but I'd rather wait a few days given we've just branched

ckongEbi:
That's fine, this is not required for this release.

ckongEbi:
Added changes to cater for eg! specific annotation source in the comments section of the EMBL/GenBank dumps

PR #105, opened 2015-12-01
Enscoresw 1381

lairdm:
- Added support in the Registry for version ids (\.\d+) support in stable_id when determining idenfitier type
- Added support to the ArchiveStableIdAdaptor for versions in stable_ids on object creation
- Test cases for identifiers with and without versions when doing lookups
- handlers in Exon, Gene, Operon, OperonTranscript and Transcript Adaptors for a versions fetch_by_stable_id call and new handler for fetch_by_stable_id_version
- Translation is handled implicitly through calls to Transcript
- Test cases for the new handlers and functionality

### DIFF
--- a/modules/Bio/EnsEMBL/DBSQL/ArchiveStableIdAdaptor.pm
+++ b/modules/Bio/EnsEMBL/DBSQL/ArchiveStableIdAdaptor.pm
@@ -131,7 +131,9 @@ use constant NUM_HIGH_SCORERS => 20;
   Arg [2]     : (optional) string $type
   Example     : none
   Description : Retrives an ArchiveStableId that is the latest incarnation of
-                given stable_id.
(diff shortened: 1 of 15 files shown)


magaliruffier:
I understand that we can get the versioned translation with the normal translation_adaptor->fetch_by_stable_id call, as the modifications to the TranscriptAdaptor take care of the possible version
However, for consistency with the other adaptors, do we want to implement a fetch_by_stable_id_version method in the TranslationAdaptor as well?
Or, because the fetch_by_stable_id method now takes care of versions for all features, should the fetch_by_stable_id_version method be private instead?

magaliruffier:
And a whole 0.1% increase in test coverage, yeah!

lairdm:
Feedback accepted and pushed.

PR #106, opened 2015-12-01
Bugfix/externaldbname

andrewyatz:
Method did not support more advanced filtering as expected from the args. This is now fixed via method delegation


magaliruffier:
yup, makes sense

avullo:
I agree, thanks!

PR #107, opened 2015-12-04
Check that adaptor exists before trying to use dbc

nicklangridge:


nicklangridge:
Please could this go onto release/83 branch also.

magaliruffier:
yes, we'll add that in, thanks

nicklangridge:
Thanks!

magaliruffier:
on branch as well now

PR #108, opened 2015-12-04
Check that adaptor exists before trying to use dbc

nicklangridge:


nicklangridge:
Sorry, forgot there were two places this is needed

nicklangridge:
Could this go onto release/83 also? Thanks

magaliruffier:
it's there now

nicklangridge:
Thanks!

PR #109, opened 2015-12-18
Jira ENSCORESW-1381

lairdm:
stable_id_version setter/getter for core object types, sets stable_id with a . delimited version or fetches stable_id with .version concatonated (if version is set)


magaliruffier:
looking good, thank you

nerdstrike:
Code is fine. In my view the usage comments should be in the description where they will be seen, rather than code comments.

PR #110, opened 2016-01-11
Allow a seq_region_name of 0 to be returned from Slice, and do a litt…

lairdm:
…le test to ensure we can create a Slice with name 0.


nerdstrike:
Good stuff. It's lazy checking, and should probably be a little more rigorous.

PR #112, opened 2016-01-18
Update insert sql for protein_feature to avoid duplicate unique keys

ckongEbi:


nerdstrike:
Hi ChuangKee, we need to be careful of insert/ignore SQL, as sqlite has iffy support for it. In BaseAdaptor we have a method insert_ignore_clause() which can be used to generate the correct ignore for the driver. See https://github.com/Ensembl/ensembl/blob/release/83/modules/Bio/EnsEMBL/DBSQL/DBEntryAdaptor.pm#L791

ckongEbi:
Thanks Kieron, sounds sensible, let me try the method in the BaseAdaptor.

Cheers,

PR #113, opened 2016-01-21
Make stable_id version nullable

danstaines:
Make version nullable for gene, transcript, translation, exon, operon… and operon_transcript

### DIFF
--- a/sql/patch_83_84_e.sql
+++ b/sql/patch_83_84_e.sql
@@ -1,0 +1,31 @@
+-- Copyright [1999-2016] Wellcome Trust Sanger Institute and the EMBL-European Bioinformatics Institute
+-- 
+-- Licensed under the Apache License, Version 2.0 (the "License");
+-- you may not use this file except in compliance with the License.
+-- You may obtain a copy of the License at
+-- 
+--      http://www.apache.org/licenses/LICENSE-2.0
+-- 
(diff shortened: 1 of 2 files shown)


sql/patch_83_84_e.sql
```diff
@@ -0,0 +1,31 @@
...
+
+ALTER TABLE gene MODIFY COLUMN version SMALLINT UNSIGNED DEFAULT NULL;
+ALTER TABLE transcript MODIFY COLUMN version SMALLINT UNSIGNED DEFAULT NULL;
+ALTER TABLE translation MODIFY COLUMN version SMALLINT UNSIGNED DEFAULT NULL;
+ALTER TABLE exon MODIFY COLUMN version SMALLINT UNSIGNED DEFAULT NULL;
+ALTER TABLE operon MODIFY COLUMN version SMALLINT UNSIGNED DEFAULT NULL;
+ALTER TABLE operon_transcript MODIFY COLUMN version SMALLINT UNSIGNED DEFAULT NULL;
+
+# Patch identifier
+INSERT INTO meta (species_id, meta_key, meta_value)
```
nerdstrike: We've already released 83, and we don't have an 84_a patch yet. We should probably keep this out of the previous release in any event.

nerdstrike: Sorry, forget I spoke. I'm wrong.

PR #114, opened 2016-01-22
Update DBEntryAdaptor.pm

markmcdowall:
Added an extra check onto the condition type in the API as there are some strings that have the text value '0'. These were then getting returned to the user via the API as ''.

### DIFF
--- a/modules/Bio/EnsEMBL/DBSQL/DBEntryAdaptor.pm
+++ b/modules/Bio/EnsEMBL/DBSQL/DBEntryAdaptor.pm
@@ -907,9 +907,14 @@ sub _store_object_xref_mapping {
             my $associatedXid = undef;
             my $sourceXid = undef;
             
-            $ax[0]->is_stored( $self->dbc ) || $self->store($ax[0]);
+            if (!$ax[0]->dbID) {
+              $self->store($ax[0]);
+            }
             $associatedXid = $ax[0]->dbID;
-            $ax[1]->is_stored( $self->dbc ) || $self->store($ax[1]);
+
+            if (!$ax[1]->dbID) {
+              $self->store($ax[1]);
+            }
             $sourceXid = $ax[1]->dbID;
             
             if (!defined $associatedXid || !defined $sourceXid) {
--- a/modules/Bio/EnsEMBL/DBSQL/DBEntryAdaptor.pm
+++ b/modules/Bio/EnsEMBL/DBSQL/DBEntryAdaptor.pm
@@ -896,8 +896,9 @@ sub _store_object_xref_mapping {
           my $group = $annotext->{$ax_group};
           my $gsth = $self->prepare( " 
                   INSERT INTO associated_group 
-                    SET description = ?;" );
-          $sth->bind_param( 1, $ax_group,     SQL_INTEGER );
+                    ( description )
+                  VALUES ( ? )" );
+          $gsth->bind_param( 1, $ax_group,     SQL_VARCHAR );
           $gsth->execute();
           my $associatedGid = $self->last_insert_id();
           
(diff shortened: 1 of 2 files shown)


nerdstrike:
I suppose it was being too clever. I think your change is fine.

magaliruffier:
could do with a test case ;-)

magaliruffier:
@markmcdowall do you have any test data you could add for this?
we don't know much about associated xrefs so we can't guarantee any changes won't affect the current code

magaliruffier:
@markmcdowall can we have some test data for this PR?
we currently have no way of checking for regression on associated xrefs, so will have to close the PR before code branching if nothing happens

markmcdowall:
Hi @magaliruffier,

Sorry I do not have any code at the moment for testing this.

andrewyatz:
Have you got any timelines when you could do? Also is this going to become a blocker for pombase?

markmcdowall:
If associated_xrefs were removed it would be a blocker for PomBase. Looking at the dropping of internal services I'll see if I can get some HCs done today.

magaliruffier:
we will not remove code that is already in place and if this update is required for release, we can pull it in even without tests
however, we will need tests at some point and I am unwilling to take in more changes after this without a minimum of regression testing
we have some documentation on Confluence on how to create test databases and write minimal tests, although obviously that is currently unavailable

markmcdowall:
I have the majority of a test class written for loading and then retrieving the annotations. Will test the test after lunch. It's only a simple can it load, can it get style of test. I'll add it to this patch when it is ready and working.

magaliruffier:
Thank you, that would be great
We just want to make sure we have a way of ensuring basic functionality is maintained when updating other areas of the code

markmcdowall:
I have added a new test to this patch for the associated_xrefs.

Seems to be generating a warning on line 915 of the DBEntryAdaptor.pm module where a DBEntry has a dbID, but not an adaptor.

What would be the fix for this?

markmcdowall:
Would modifying the DBEntryAdaptor to use exists to test for if the xref is present in the xref table be better than using the is_stored function as all the function requires is the dbID?

markmcdowall:
Modified the DBEntryAdaptor to check for the existence of the dbID in the DBEntry when loading associated xrefs rather than using is_stored.

This is now passing the tests when run locally and no warning is raised.

magaliruffier:
If I understand the code correctly, you are already storing one of the associated xrefs when storing the SourceXref for the OntologyXref (somewhere around here: https://github.com/markmcdowall/ensembl/blob/patch-1/modules/Bio/EnsEMBL/DBSQL/DBEntryAdaptor.pm#L874)
this means that it is not needed to store/check is_stored for the second associated xref (https://github.com/markmcdowall/ensembl/blob/patch-1/modules/Bio/EnsEMBL/DBSQL/DBEntryAdaptor.pm#L915)

markmcdowall:
Correct

magaliruffier:
I think the tests are failing now because, by inserting new data, subsequent tests run on data that's different from their expected results
you can circumvent that by using the hide and restore methods
You can see an example here:
https://github.com/Ensembl/ensembl/blob/release/83/modules/t/gene.t#L299

markmcdowall:
It is happy on MySQL, much the code is not so happy when it comes to the SQLite install.

magaliruffier:
it is missing a _tables method in the DBEntryAdaptor
something like https://github.com/Ensembl/ensembl/blob/release/83/modules/Bio/EnsEMBL/DBSQL/GeneAdaptor.pm#L84

markmcdowall:
Looks like it is now passing the SQLite tests. I defined the table in the function rather than create an _table method for the whole module. This looks to be how iut has been done in the rest of the module for other calls to last_insert_id.

magaliruffier:
seems reasonable
Thanks a lot for the tests, I will sleep better tonight knowing we will know if we break it :-)

PR #115, opened 2016-01-26
Support g. in a location string and some comments

andrewyatz:
Extending support so locations like X:g.1-10 are parsed as a genomic
coordinate. Also added some comments as the regular expressions are
slightly cryptic.


magaliruffier:
straightforward enough, thanks

nerdstrike:
I've always thought we needed more ascii in our number lines.

PR #116, opened 2016-02-04
Use Feature::species() as the only source for species name

muffato:
This would ease making Compara's ConstrainedElement a sub-class of Feature. ConstrainedElement's adaptor does not derive from BaseFeatureAdaptor. It is Compara's ConstrainedElementAdaptor which "species" is "Multi".
With the proposed change in, we only have to override ConstrainedElement::species()


magaliruffier:
that looks reasonable to me

nerdstrike:
Agreed, it probably should have been using that method all along.

PR #117, opened 2016-02-04
nasty way to make versions unique

dbolser-ebi:
could be done better I'm sure

### DIFF
--- a/misc-scripts/assembly/ensembl_assembly_to_chain.pl
+++ b/misc-scripts/assembly/ensembl_assembly_to_chain.pl
@@ -161,17 +161,14 @@ sub get_liftover_mappings {
   my ($core_dba) = @_;
   my $mappings = $core_dba->get_MetaContainer()->list_value_by_key('liftover.mapping');
   my %unique_mappings;
-  for (@$mappings){
-      die "Can't parse mapping string for coord system version '$_'!\n"
-          unless /.+:(.+)#.+:(.+)/;
+  foreach my $mapping (@{$mappings}) {
+      die "Can't parse mapping string for coord system version '${mapping}'!\n"
+          unless $mapping =~ /.+:(.+)#.+:(.+)/;
       my $version_1 = $1;
       my $version_2 = $2;
-      $unique_mappings{$version_1. ":". $version_2}++;
+      $unique_mappings{$version_1. ":". $version_2} = [$version_1, $version_2];
   }
-  my @unique_mappings;
-  push @unique_mappings, [split(/:/, $_)]
-      for keys %unique_mappings;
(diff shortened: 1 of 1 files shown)


dbolser-ebi:
One one database tested, the output file was strictly identical, but returned 4 times faster.

misc-scripts/assembly/ensembl_assembly_to_chain.pl
```diff
@@ -160,7 +160,18 @@ sub run_on_dba {
...
+  my %unique_mappings;
+  for (@$mappings){
+      die "Can't parse mapping string for coord system version '$_'!\n"
+          unless /.+:(.+)#.+:(.+)/;
+      my $version_1 = $1;
+      my $version_2 = $2;
+      $unique_mappings{$version_1. ":". $version_2}++;
+  }
+  my @unique_mappings;
+  push @unique_mappings, [split(/:/, $_)]
```
nerdstrike: Seems a little odd to join the versions together and then split them up again.

dbolser-ebi: Quickest way I could think of to make a unique list of pairs... hacky I
know.

andrewyatz: What no schwarzian transform? Boo :)

dbolser-ebi:
my @unique_mappings = List::UtilsBy::uniq_by { join ":", @$_ } @mappings; #?

andrewyatz:
Sorry @dbolser-ebi that this went dead. Using an external module is probably overkill. How about this?

```perl
my %unique_mappings;
foreach my $mapping (@{$mappings}) {
    die "Can't parse mapping string for coord system version '${mapping}'!\n"
       unless /.+:(.+)#.+:(.+)/;
    my $version_1 = $1;
    my $version_2 = $2;
    $unique_mappings{$version_1. ":". $version_2} = [$version_1, $version_2];
}
return [values %unique_mappings];
```

Avoids the need to join and split, avoids use of `$_` and reduces the name clobbering between hash and arrays.

dbolser-ebi:
Looks nice.

magaliruffier:
Can you please incorporate Andy's suggestion and test it still works for you?

magaliruffier:
@dbolser-ebi we'd like to incorporate the change before code branching, which is on Thursday
can you please test Andy's suggestion and update the PR accordingly if that works for you?

dbolser-ebi:
Tested fine, i.e. the MD5SUM of the output files was identical before and after. Didn't observe the pointless looping.

magaliruffier:
Great, thank you

PR #118, opened 2016-02-05
Experimental/travis reorg

lairdm:
Setting a good example of organizing our repositories properly in the travis build


andrewyatz:
Just a quick comment. There's the `$ENSDIR` ENV. Couldn't that be set to `$PWD/..` to stop you from having to decorate everything with `$ENSDIR/..`?

lairdm:
I suppose I could remove those now. I originally made the $ENSDIR change so that the harness.sh could handle both the current Travis setup (sub-repositories in the current repository) and the more "standard" configuration, so I could use harness.sh in the VM built too. If you set the ENSDIR env before invocation, it'll look for the dependent repositories where you tell it to, in the case of the VM a:
ENSDIR=/home/ensembl/ensembl-api-folder travisci/harness.sh
(pre ../ change obviously)

It's potentially useful so you can put those dependencies in either config, or elsewhere, and harness.sh will still work, but I'm not married to it, if you think it's potentially confusing they can go.

PR #119, opened 2016-02-12
Feature 84/test var methods

at7:
- deprecate all the variation related methods in the core API
- provide a variation test db for testing the relevant methods in the core API


magaliruffier:
Yes to deprecation, and thanks for the tests!
Travis is complaining about the key length, this is usually due to the size (in bytes rather than characters) of the keys on the tables.
This can be solved by forcing the charset to be latin, rather than the default UTF8 that travis is using.
You can see how we have done it for the core schema here: https://github.com/Ensembl/ensembl/blob/release/83/modules/t/test-genome-DBs/homo_sapiens/empty/table.sql or copy the schema table from the rest tests directly: https://github.com/Ensembl/ensembl-rest/blob/release/83/t/test-genome-DBs/homo_sapiens/variation/table.sql

at7:
Thank you. I had that error before but forgot about it.

magaliruffier:
we might need to comment out the Test::Warnings in the dependencies.t test until the deprecated methods have been removed, as this causes the test to fail

at7:
Would there be a way of allowing deprecation warnings?

magaliruffier:
there is the warns_like method, which accepts warnings matching a certain pattern
That might work
You can find some examples in the core tests:
https://github.com/Ensembl/ensembl/blob/release/83/modules/t/registry.t

at7:
The regex in line 28 https://github.com/Ensembl/ensembl/blob/master/modules/t/dependencies.t matches the deprecation message. e.g. _Please use **Bio::EnsEMBL::Variation**::DBSQL::VariationFeatureAdaptor::fetch_all_genotyped_by_Slice instead._

Can I delete the key from the grep result hash if it contains 'deprecate'? Like:
foreach my $key (keys %result) {
  if ( $key =~ /($exceptions)/ ) {
     delete($result{$key});
  }
  if ( key =~ /deprecate/ ) {
    delete($result{$key});
  }
}

magaliruffier:
well spotted
would it not be possible to change the regex to match lines starting with use?
we can even allow for spaces before, but that should ignore all the lines starting with Please (hence the use of being polite!)

at7:
Yes, I like the starting with use. Shall I include this in my pull request?

magaliruffier:
Yes, please, then we will hopefully remember why we did it ;-)
and if that fixes travis, we should be able to take the PR in as is

at7:
I have seen the sqlite error before. Can I create a separate file (sliceVarMethods.t) and exclude this from the sqlite tests? I've done this for strainSlice.t (https://github.com/Ensembl/ensembl/blob/release/83/travisci/harness.sh).

nerdstrike:
I'm not sure variation data makes any sense in volumes suitable for SQLite, so perhaps the testing can be foregone.

magaliruffier:
I have a vague suspicion this might be due to the size of those auto-increments, as SQLlite might have smaller indexes, especially in a sparse test environment like travis
it should be fine to ignore this test for sqllite in the harness

at7:
Excluding the slice.t. Or extracting variation related tests into a different file and excluding that one? I had the same problem with strainSlice.t. I think the problem was because of some sql we use in the variation API which is not supported by sqlite.

magaliruffier:
I had forgotten about the SQLlite specific schema file, sorry
If you run the patch_test_databases.pl script, it will generate one automatically

magaliruffier:
Victory!
All good now, looking forward to release 88 and the big cleanup ;-)

PR #120, opened 2016-02-19
 added MKG in location validation

ens-ma7:
Hi Mag
this is the extra validation for the location, can you please let me know if there is anything not right.

thanks
Ridwan


magaliruffier:
Looks straightforward enough ;-)
If that works for your example cases, happy to take the change in

ens-ma7:
Hi Mag
It did work on my sandbox.
Thanks
Ridwan

PR #121, opened 2016-02-23
Import deprecate method

at7:


avullo:
Should be ok, thanks.

at7:
return Variation::StrainSlice. We probably want to get rid of this as well in the future? For now it avoids the StrainSlice module in the core API and uses the new one in the variation API.

at7:
I try to move this into the variation API now and remove the dependency.

at7:
OK. I moved get_by_strain to the variation API.

magaliruffier:
thank you

PR #122, opened 2016-02-23
At7 84/minor updates deprecation

at7:


avullo:
Thanks, that's good for me.

PR #123, opened 2016-03-02
Configuration to allow alignments to be exported to GFF

james-monkeyshines:
In order for GFFSerializer to export alignments, I found I needed two small changes: one to define the relevant SO terms; the other to get alignments on MT chromosomes to produce seq_region coords.


nerdstrike:
All updates to the SO mappings are welcome.

magaliruffier:
I don't think we've ever tried exporting alignments, the changes look sensible

PR #124, opened 2016-03-10
Correct serializer behaviour, if an attribute for column 9 of the GFF…

lairdm:
… serializer (the misc, everything column) contains an empty array in the feature structure, skip printing the attribute rather than attribute=ARRAY(0xABCD)


nerdstrike:
How does splitting the condition make the outcome different?

lairdm:
In the old method of a one line condition it fell through to the else if the array was empty, which simply tries to print it out. Now it will get trapped in the ref == ARRAY half, and if it's empty just skip and not write the row out.

nerdstrike:
I see, good!

PR #126, opened 2016-03-18
Update schema_patcher.pl

ilavidas:
Funcgen has renamed its sql file from 'efg.sql' to 'table.sql', therefore schema_patcher.pl has to be modified to reflect this change.


nerdstrike:
Tidy, thank you.

lairdm:
Seems simple enough

PR #127, opened 2016-03-31
store ontology term synonym xref separately from name

sarahhunt:
Allows searching by synonym via the API.

### DIFF
--- a/misc-scripts/ontology/sql/patch_84_85_c.sql
+++ b/misc-scripts/ontology/sql/patch_84_85_c.sql
@@ -1,0 +1,28 @@
+-- Copyright [1999-2016] Wellcome Trust Sanger Institute and the EMBL-European Bioinformatics Institute
+-- 
+-- Licensed under the Apache License, Version 2.0 (the "License");
+-- you may not use this file except in compliance with the License.
+-- You may obtain a copy of the License at
+-- 
+--      http://www.apache.org/licenses/LICENSE-2.0
+-- 
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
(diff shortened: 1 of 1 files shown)


sarahhunt:
Required for https://github.com/Ensembl/ensembl-production/pull/40

This leaves no way to access dbxrefs via the API - I can add one if it would be useful?

magaliruffier:
I have not found any data where that field was actually populated, so you have any examples?
Given no one has needed it so far, I can survive without the API support for the time being
Otherwise, change looks fine

sarahhunt:
GO terms often have an xref:

mysql> select * from synonym where term_id in (12,14,24);
+------------+---------+------------------------------------------------------------------+---------+
| synonym_id | term_id | name                                                             | type    |
+------------+---------+------------------------------------------------------------------+---------+
|          7 |      12 | "single-stranded DNA specific endodeoxyribonuclease activity" [] | RELATED |
|          8 |      12 | "ssDNA-specific endodeoxyribonuclease activity" [GOC:mah]        | RELATED |
|         10 |      14 | "lactase-phlorizin hydrolase activity" [EC:3.2.1.108]            | BROAD   |
|         11 |      14 | "lactose galactohydrolase activity" [EC:3.2.1.108]               | EXACT   |
|         29 |      24 | "50S ribosomal subunit assembly" [GOC:mah]                       | NARROW  |
|         30 |      24 | "60S ribosomal subunit assembly" [GOC:mah]                       | NARROW  |
+------------+---------+------------------------------------------------------------------+---------+

lairdm:
Potentially silly question, are there downstream implications of this, how that new column will be loaded and accessed?

sarahhunt:
@lairdm - there is a PR to update the loader : Ensembl/ensembl-production#40 and as discussed, there is now no way to extract dbxrefs, where present, via the API - happy to add one.

magaliruffier:
bioperl is currently down, so all test suites fail
We're going to wait until Monday to see if it is resolved before taking new changes in

lairdm:
Bioperl fixed, merging.

PR #128, opened 2016-04-05
added support for ensembl-metadata

danstaines:
Bio::EnsEMBL::Registry now has support for the new ensembl-metadata API, handled in the same way as ensembl-taxonomy.


magaliruffier:
I guess it's going to be hard to add a test for this in the core API
as long as it's tested in the ensembl-metadata API, change looks fine to me

andrewyatz:
If that gets integrated I have an idea for a speed increase for the grep @danstaines modified in get_species_and_object_type. After Will's optimisation talk I don't think multiple `$dba->dbc->dbname` calls is going to help speed.

danstaines:
is this ok to be pulled then? Integration testing in Registry is a tricky problem - and one I happily leave to others... 
@andrewyatz  - I'll leave you to make optimisation changes as you see fit.

lairdm:
Looks sane :)

PR #129, opened 2016-05-03
Retain the backslash on escaped single quotes when generating a hash from a flattened string.

james-monkeyshines:
Data::Dumper is used to flatten hashes into strings, for insertion into the 'external_data' columns of some tables. By default, it uses single quotes, and automatically does any necessary escaping. The code to subsequently extract the values from the 'external_data' column removed all backslashes, resulting in an invalid hash; e.g. `'Desc' => '3\'UTR',` becomes `'Desc' => '3'UTR',`.
Because Data::Dumper escapes backslashes, we can remove all superfluous backslashes by removing all double-backslash instances, which leaves the single-backslash-escaped quotes intact.

### DIFF
--- a/modules/Bio/EnsEMBL/DBSQL/BaseAdaptor.pm
+++ b/modules/Bio/EnsEMBL/DBSQL/BaseAdaptor.pm
@@ -1109,7 +1109,7 @@ sub get_dumped_data {
     my $self = shift;
     my $data = shift;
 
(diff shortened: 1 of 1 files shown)


magaliruffier:
As far as I know, we don't really use the external_data field, so the change should be fine with us
The test suite fails because of changes to the variation schema (!), not something we need to worry about here

PR #130, opened 2016-05-06
Test/transcript mapper

lairdm:
- Mag fixed the error in the mapper
- Associated tests which this fix broke have been fixed (transcript.t and exon.t)


nerdstrike:
This pull deals with two different issues. Merge message should mention both of them.

lairdm:
Separated

PR #131, opened 2016-05-10
Remove variation_feature_bak

at7:


nerdstrike:
Yeah, that doesn't belong there

PR #132, opened 2016-05-17
store version for ontology

sarahhunt:
We need to be able to track the versions of the ontologies used in each release. This updates the schema, API and test files.


nerdstrike:
I suppose this should have been in there from the start. Code looks fine, and I see you've modified load_OBO_file.pl too!

nerdstrike:
Bad luck, you missed the sqlite table defs.

sarahhunt:
Yes - I'm told up to 10 terms can be deprecated a month from one of the ontologies we use, so it's essential to know which we are using.

magaliruffier:
neat and tidy, thanks

PR #133, opened 2016-05-31
Updated copyright script for transition to EBI-only copyright

dzerbino:


magaliruffier:
👍

nerdstrike:
Wouldn't it be nice to not have to revisit this after the copyright handover?

dzerbino:
Indeed, but coming up with a clever regex that detects the presence/absence of the old one-line copyright did not seem worth the trouble. Next year we can update the script to its definitive form.

PR #134, opened 2016-06-02
Corrected copyright updater to deal with multi format licenses

dzerbino:
D'oh! I hadn't realised that the different copyrights have different line starts. Had to update the regex.

Sorry.


magaliruffier:
I guess you also meant to lose the Copyright text?

dzerbino:
Copyright (and whatever is before it on the line) is now represented as \1 in the replacement string.

lairdm:
Regex fun.

PR #135, opened 2016-06-02
Update TranscriptSelector.pm

amonida:
Added polymorphic_pseudogene to the biotype priority list at the same ranking as NMD and NSD.


lairdm:
Seems straight forward enough.

PR #140, opened 2016-06-24
New attrib table for dna_align_features

james-monkeyshines:
The new dna_align_feature_attrib table is to replace functionality that would be lost when the 'external_data' column is removed from the dna_align_feature table.

I've added tests for the new functionality, and they run fine when I use a preloaded test db with the schema modification. I didn't feel confident in updating the test-genome-DBs directory, however, so the Travis tests will fail until the table is added into the table defs there (the tests do not assume that there will be any rows in the dna_align_feature_attrib table).

### DIFF
--- a/modules/Bio/EnsEMBL/DBSQL/AttributeAdaptor.pm
+++ b/modules/Bio/EnsEMBL/DBSQL/AttributeAdaptor.pm
@@ -179,6 +179,14 @@ sub store_batch_on_Translation {
   return;
 }
 
+sub store_batch_on_DnaDnaAlignFeature {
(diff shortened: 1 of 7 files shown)


avullo:
Your changes won't be incorporated until 85 is out.
In the meantime, have a look at:
https://www.ebi.ac.uk/seqdb/confluence/display/ENSCORE/Creating+a+test+database
https://www.ebi.ac.uk/seqdb/confluence/display/ENSCORE/Patching+test+databases
for anything related to updating the test databases.
Might suggest we close this PR and reconsider it after Jul 12.

james-monkeyshines:
OK. I'll try that patching script and see how I get on...

PR #141, opened 2016-06-24
bugfix: get_all_DBAdaptors() should return an empty list if the species name (alias) does not exist

muffato:
Otherwise it returns the list of all DBAdaptors ... (when -GROUP is not set)

### DIFF
--- a/modules/Bio/EnsEMBL/Registry.pm
+++ b/modules/Bio/EnsEMBL/Registry.pm
@@ -669,7 +669,10 @@ sub get_all_DBAdaptors {
 
   my ( $species, $group ) = rearrange( [qw(SPECIES GROUP)], @args );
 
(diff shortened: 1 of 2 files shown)


nerdstrike:
I fully expect there to be consequences for this change, but the change is valid.

muffato:
I did run registy.t but only focused on the two tests and missed the error:
```
not ok 21 - no (unexpected) warnings (via done_testing)
#   Failed test 'no (unexpected) warnings (via done_testing)'
#   at modules/t/registry.t line 136.
1..21
```
I'll try to fix that

muffato:
It looks like I need to test all the warning messages too. It doesn't hurt :)

magaliruffier:
sorry, I am not sure I fully understand.
Are you planning to add some changes or is this good to go?

muffato:
It's good to go

PR #143, opened 2016-07-08
Renamed a couple of keys for improved consistency

dbolser-ebi:
What's with the key on line 1153?


magaliruffier:
to modify the schema, you will need to provide a patch file as well as update the test databases
Some instructions on how to do it can be found here:
https://github.com/Ensembl/ensembl/blob/release/84/CONTRIBUTING.md

dbolser-ebi:
Thanks for the tip.​

PR #144, opened 2016-07-12
New attrib table for dna_align_features

james-monkeyshines:
The new dna_align_feature_attrib table is to replace functionality that would be lost when the 'external_data' column is removed from the dna_align_feature table.

Tests and test databases updated accordingly.

### DIFF
--- a/modules/Bio/EnsEMBL/DBSQL/AttributeAdaptor.pm
+++ b/modules/Bio/EnsEMBL/DBSQL/AttributeAdaptor.pm
@@ -179,6 +179,14 @@ sub store_batch_on_Translation {
   return;
 }
 
+sub store_batch_on_DnaDnaAlignFeature {
+  my ($self, $attributes, $batch_size) = @_;
+
+  $self->store_batch_on_Object('dna_align_feature', $attributes, $batch_size);
+
+  return;
+}
+
 
 ## Single store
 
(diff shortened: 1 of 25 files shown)


nerdstrike:
Thanks James. I think you're being a bit optimistic with no validation on arguments to some of your methods, but the general impression is promising. We should be able to get this into release 86.

magaliruffier:
Two tests failures:
- schemaPatches fails
it is probably because the table.sql file does not contain the patch meta key
It would require something like: 
INSERT INTO meta (species_id, meta_key, meta_value)
 +  VALUES (NULL, 'patch', 'patch_85_86_b.sql|add dna_align_feature_attrib table');

- SQLite fails
with the following error message
DBD::SQLite::db prepare failed: no such table: dna_align_feature_attrib at /home/travis/build/Ensembl/ensembl/modules/Bio/EnsEMBL/DBSQL/DBConnection.pm line 707.
I think that's because the attributeAdaptor.t tests uses the files from the patch test database (https://github.com/Ensembl/ensembl/tree/release/85/modules/t/test-genome-DBs/homo_sapiens/patch) and the SQLite file for that one has not been updated

nerdstrike:
Generally speaking we shouldn't be seeing our own commits in your pull request. Something may have gotten mixed up.

magaliruffier:
looking good now, thanks for the addition

nerdstrike:
I think this is sound

PR #145, opened 2016-07-12
Sequence ontology mappings for RNA pseudogenes

james-monkeyshines:
The pseudogenic_rRNA and pseudogenic_tRNA terms are descendants of transcript, not gene; there are no similar RNA-specific terms underneath pseudogene.


nerdstrike:
Thanks James. Sensible changes

magaliruffier:
that's part of the changes we have added to the consensus SO mapper, to be added once we have an agreement with Gencode
https://github.com/Ensembl/ensembl/blob/feature/so_mapper/modules/Bio/EnsEMBL/Utils/SequenceOntologyMapper.pm
do you need those changes now or can we wait until we've merged everything in?

james-monkeyshines:
I can wait.
I noticed yesterday that transcripts with 'pseudogene' biotype get 'pseudogene' SO terms (which is also true in the new file you linked to); but I think these should be 'pseudogenic_transcript' instead?

magaliruffier:
makes sense, I've updated the branch
I'll close this PR now, but do prod again if we don't get this mapper sorted soon!

PR #146, opened 2016-07-21
Compara test db

muffato:
I have added a test database and tests for these 3 methods:
 -  Bio::EnsEMBL::Slice::get_all_compara_Syntenies()
 -  Bio::EnsEMBL::Slice::get_all_compara_DnaAlignFeatures()
 -  Bio::EnsEMBL::Gene::get_all_homologous_Genes()

If you find any other method that connects to the Compara API, let me know


muffato:
The first build is failing because I forgot to add ensembl-compara to the dependencies. Should work now

muffato:
Now failing because tables.sql is not compatible with sqlite. Will fix that now

muffato:
There was some problem in the data. Fixed now. The _999th_ Travis build has just started !

magaliruffier:
That's great work, thank you :-)

nerdstrike:
Fair enough. I'm not thrilled at the core API testing getting circularly dependent on related APIs, but I suppose that ship has already sailed.

muffato:
@nerdstrike We were asked that (some time ago, admittedly) ! In fact, Compara's Travis build runs the few .t tests from ensembl-rest that use Compara API, and I will soon make it run this new .t from ensembl.
The idea was that we should not break your API, and fix it if we do

PR #148, opened 2016-08-16
Check that translations exist before comparing the sequences

thibauthourlier:
If a translation object does not exist, we do not try to access its sequence


nerdstrike:
I suppose that's sensible. Avoids a "Can't call seq on uninitialised value"?

thibauthourlier:
Yes that's what it's for. I forgot to mention that it's related to ENSCORESW-1896

lairdm:
Seems like reasonable sanity checking.

PR #149, opened 2016-08-19
Master

ens-carlos:
No lazy loading if $load_exons is true.
If $load_exons is true, do not do lazy loading in any case (including when the number of transcripts is 1).

### DIFF
--- a/modules/Bio/EnsEMBL/AltAlleleGroup.pm
+++ b/modules/Bio/EnsEMBL/AltAlleleGroup.pm
@@ -53,14 +53,14 @@ limitations under the License.
   }
   
   # Creating and editing an AltAlleleGroup
-  
-  my $type_flags = [qw(IS_MOST_COMMON_ALLELE AUTOMATICALLY_ASSIGNED)];
+
+  my %type_flags = ('IS_MOST_COMMON_ALLELE' => '1','AUTOMATICALLY_ASSIGNED' => '1');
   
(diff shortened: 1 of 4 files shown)


nerdstrike:
Well, I found the original reason or lack thereof of for lazy loading of 0/1 exons back in 2004... it seems to be just because pre-fetching isn't usually worth the cost unless there are multiple things to fetch.
Can you briefly outline the circumstances where this wasn't working?

ens-carlos:
I'd like to remove a dna align feature as transcript supporting feature from a database so I (tried to) fully load the transcript and then remove it from the database and store the modified transcript object but I couldn't get all the exons because of the lazy loading by default for slices containing 1 transcript.
I think that the parameter description should be fixed to match what it does although I'd prefer having the option to fully load the exons.

nerdstrike:
Your modifications are sensible, but I'm wary of unintended consequences. We could roll with it and see what happens.

magaliruffier:
do we still need the check on the length of transcripts in that case?
I can imagine the confusion in a few years' time as to why we still need lazy-loading when there is nothing

nerdstrike:
I think this is fine. Any fallout from this change will serve to expose false assumptions in downstream processes.

ens-carlos:
Just to confirm that I ran some tests and the bit about the check of preloaded exons should stay so all should be fine as I hadn't change that.

PR #150, opened 2016-09-06
Typo in Identity Xref function, variable should be $xref_end, not $xref_start

james-monkeyshines:


nerdstrike:
Hopefully this didn't inconvenience you. Looks like the fix is correct

magaliruffier:
that's slightly embarrassing, thanks for the fix

PR #157, opened 2016-09-26
Updated the new call to create the ProteinFeature object in ProteinFe…

premanand17:
…atureAdaptor to match the attributes expected from ProteinFeature - ENSCORESW-1912


modules/t/proteinFeatureAdaptor.t
```diff
@@ -21,6 +21,7 @@ use Test::More;
 use Test::Warnings;
 
 use Bio::EnsEMBL::Test::MultiTestDB;
+use Test::Differences;
```
nerdstrike: Looks like this isn't being used anywhere.

nerdstrike:
I don't see any reason to decline this change.

PR #159, opened 2016-10-07
Fixed bug in set_gene_descriptions where filtered descriptions were b…

khowe:
…eing ignored in favour of the original description


magaliruffier:
I'm going to test this on human but otherwise it looks good to me

magaliruffier:
no changes spotted on human, I think this is good to go in

PR #161, opened 2016-10-20
[ENSCORESW-1997]. Enable local load data feature at the specific driv…

avullo:
…er level.

Unfortunately, creating an appropriate test case turns out to be extremely inconvenient
as one must make sure the underlying DBD::mysql has been compiled against a C library
where this feature is turned off.

This approach has been tested at the base class level and it seems to be working, so this
is more for:
- the team to quickly have a look and make appropriate suggestions;
- @andrewyatz to set up a proper test environment, run a test and confirm, as we've discussed.

### DIFF
--- a/modules/Bio/EnsEMBL/DBSQL/Driver/mysql.pm
+++ b/modules/Bio/EnsEMBL/DBSQL/Driver/mysql.pm
@@ -34,6 +34,20 @@ use strict;
 
 use base 'Bio::EnsEMBL::DBSQL::Driver';
 
+#
+# override parent's method to enable MySQL local load data in case DBD::mysql 
+# has been compiled against a C client library which has been built with
+# no support for this feature
+#
+sub connect_params {
+  my ($self, $conn) = @_;
+
+  my $params = $self->SUPER::connect_params($conn);
+  $params->{attributes}{mysql_local_infile} = 1;
+
+  return $params;
+}
+
 sub from_date_to_seconds {
     my ($self, $column) = @_;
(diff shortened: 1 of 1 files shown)


andrewyatz:
Confirmed the fix removes the problem. Code output from my test script using `release/86`.

```perl
DBD::mysql::db do failed: The used command is not allowed with this MySQL version at /homes/ayates/ensembl/modules//Bio/EnsEMBL/DBSQL/DBConnection.pm line 766.

-------------------- EXCEPTION --------------------
MSG: Detected an error whilst executing statement 'load data local infile 'load.txt' into table mytab': DBD::mysql::db do failed: The used command is not allowed with this MySQL version at /homes/ayates/ensembl/modules//Bio/EnsEMBL/DBSQL/DBConnection.pm line 766.

STACK Bio::EnsEMBL::DBSQL::DBConnection::do /homes/ayates/ensembl/modules//Bio/EnsEMBL/DBSQL/DBConnection.pm:771
STACK toplevel ./testload.pl:24
Date (localtime)    = Thu Oct 20 21:49:26 2016
Ensembl API version = 86
---------------------------------------------------
```
I do not get the above error and the script correctly executes and loads data into the table.

andrewyatz:
👍  this commit does look good

avullo:
Glad to hear that! Anyone in core wanting to merge so we can flag this to compara for attention?

lairdm:
I'm sure there are good security reasons, but I've always hated that local infile is usually disabled by default.

PR #163, opened 2016-11-07
Enscoresw 1844

premanand17:


nerdstrike:
This might need some more documentation in POD for the method, to explain the scope of the arguments.

magaliruffier:
I would recommend contacting web first so they can check this fits their requirements before merging into master

premanand17:
Discussed with Ridwan about the fix and he has given a go ahead.

PR #165, opened 2016-11-24
Updates to include projection parent id in GTF dumps - attribute code…

premanand17:
…s proj_parent_g, proj_parent_t added to the dumps if available - ENSCORESW-2003


nerdstrike:
Looks ok to me

PR #166, opened 2016-11-24
Added abstract method _tables implementation to TranslationAdaptor as…

premanand17:
… needed by generic_count in BaseAdaptor - ENSCORESW-1995


modules/Bio/EnsEMBL/DBSQL/TranslationAdaptor.pm
```diff
@@ -965,4 +965,18 @@ sub fetch_all {
   return \@translations;
 }
 
+
+# _tables
```
nerdstrike: Strictly speaking this should be a =head2 block of pod in order for it render in the API reference. It is however an internal method so it can pass.

PR #167, opened 2016-11-30
Added an option in BaseAdaptor to define the column of the dbID

muffato:
It defaults to ${table_name}_id but allows individual adaptors to rename it. This feature would be useful for the Compara API.

I haven't changed DnaAlignFeatureAdaptor and ProteinAlignFeatureAdaptor because they're not base classes. Similarly there are still other places (especially direct SQL queries) where ${table}_id is used (ex: https://github.com/Ensembl/ensembl/blob/release/86/modules/Bio/EnsEMBL/DBSQL/TranscriptAdaptor.pm#L2130 )

### DIFF
--- a/modules/Bio/EnsEMBL/DBSQL/BaseAdaptor.pm
+++ b/modules/Bio/EnsEMBL/DBSQL/BaseAdaptor.pm
@@ -666,8 +666,9 @@ sub _uncached_fetch_by_dbID{
   #construct a constraint like 't1.table1_id = 123'
   my @tabs = $self->_tables;
   my ($name, $syn) = @{$tabs[0]};
+  my $col = $self->_dbID_column;
   $self->bind_param_generic_fetch($id,SQL_INTEGER);
-  my $constraint = "${syn}.${name}_id = ?";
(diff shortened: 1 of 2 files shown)


nerdstrike:
This makes me nervous... I doubt this is the only place in the API that assumes table_id column names.

muffato:
Oh, there are a lot of places :/ Just do `rgrep -w _id`
In many cases, this is done on a table without the adaptor of the latter, so my commit doesn't really help

nerdstrike:
@muffato So why does the Compara API need non-standard primary key names? method_link_species_set_db_id?

muffato:
It mostly affects all our tables for trees (incl. the ncbi taxonomy tables). For us it allows to have a base adaptor with shared methods and other things like that. 
I'm working on new methods for our BaseAdaptor: `generic_update()` and `generic_insert()` , and they'd work better with a reliable method of getting the dbID column. If it doesn't go to the core API, I'll likely still add something similar to the Compara API

nerdstrike:
What about patching your table names? That would localise the scope of the change to the unusual tables, rather than altering the execution path of all database calls ever.

PR #168, opened 2016-12-09
Enscoresw 1971

premanand17:
Added transl_table qualifier to genbank and embl flat file dumps (if codon_table attribute is available in seq_region_attribute table for that region)


magaliruffier:
Looks fine

PR #169, opened 2016-12-13
Updates to add projection parent as attributes in gtf and gff dumps -…

premanand17:
Updates to add projection parent as attributes in gtf and gff dumps


nerdstrike:
Looks alright to me

PR #170, opened 2016-12-17
87 -> 88

dbolser-ebi:
Is it really this simple? I'm doing this to run the peptide compara pipeline, which uses the ensembl release version deeply.


nerdstrike:
It's nearly this simple, but as part of our SOP we also patch our database meta keys. It will be done shortly, but your patch will serve you until then.

dbolser-ebi:
ta​

PR #171, opened 2017-01-04
Updated documentation for map_coordinates - ENSCORESW-2089

premanand17:
Updated documentation for map_coordinates - Returntype


nerdstrike:
I guess that's the best we can do under the circumstances.

lairdm:
Yup, best of an unfortunate situation.

PR #172, opened 2017-01-04
Updating script to be able to update the EBI Copyright every year. No…

thomasmaurel:
…w excluding the .git directory from the file search as this was corrupting the git repos

@dzerbino and @magaliruffier updated the script last year so that we now display a copyright like the following in our scripts:

> # Copyright [1999-2015] Wellcome Trust Sanger Institute and the EMBL-European Bioinformatics Institute
> # Copyright [2016] EMBL-European Bioinformatics Institute

I've updated the script so that we can update the EMBL-EBI part every year (similar to what the script was doing in the past). For this year, I believe we would want something like:

> # Copyright [2016-2017] EMBL-European Bioinformatics Institute

I've also excluded the ".git" directory from the grep as this was corrupting the Git repos :-/

### DIFF
--- a/misc-scripts/annual_copyright_updater.sh
+++ b/misc-scripts/annual_copyright_updater.sh
@@ -35,7 +35,7 @@ for var in $dirs; do
   last_year=$(($year - 1))
 
   search="^\(.*\)\\[2016\(-*[0-9]*\)\\] EMBL-European Bioinformatics Institute"
-  replacement="Copyright [2016-$year] EMBL-European Bioinformatics Institute"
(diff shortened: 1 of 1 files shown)


nerdstrike:
I guess that's ok. Does this script deal with alternate expressions, like license in a comment instead of a POD block?

thomasmaurel:
Well spotted @nerdstrike, there was a bug for anything different than a POD block.

nerdstrike:
Excellent, we don't need this coming back to haunt us again.

PR #176, opened 2017-01-17
Made the assumption that all the EBI copyright would start by 2016 bu…

thomasmaurel:
…t it's not the case for the Ensembl genomes repos. Now updated regex to be more generic


nerdstrike:
Fair enough, hope the regexes aren't getting too hairy.

magaliruffier:
definitely useful for all those projects that are not necessarily under the Ensembl umbrella

PR #177, opened 2017-01-18
Use proper arrayref address

willmclaren:
(@array_of_arrayrefs)[0][0] breaks in perl 5.8


avullo:
Thanks Will, didn't know that.

PR #179, opened 2017-01-26
added a statement to query the family table for rest family endpoint,…

waakanni:
… also fixed a bug


nerdstrike:
What do you have against our empty lines?

magaliruffier:
What Kieron meant is, it is generally recommended not to mix stylistic changes with code changes in the same commit.
And it is also better not to change an existing style, unless there is a strong argument for it.
Can you please submit a new PR with only the code changes? I think there are only two lines actually modified.

waakanni:
On what files?

PR #181, opened 2017-01-27
Protein Feature patch

epaule:
this time on the master


epaule:
there is a typo in the apache license header of patch_87_88_c.sql

nerdstrike:
Broadly ok, but are you/should you try to absorb all possible SQL errors e.g. server gone away. I presume the state of RaiseError is true at this point, so you might have to do: local $dbh->{RaiseError} for your print statements to fire. What states have you tried?

magaliruffier:
you will need to patch the test databases to include the change, documentation on this can be found here: https://www.ebi.ac.uk/seqdb/confluence/display/ENSCORE/Patching+test+databases

epaule:
you think that is the reason for:
#   Failed test 'patch_87_88_c.sql|protein_featue_uniqueness in original database'
#   at modules/t/schemaPatches.t line 176.

epaule:
still bars at comparing the meta tables ... is there any other obscure file that needs to be changed and isn't covered by the patch_test_databases ?

magaliruffier:
the table.sql file does not contain the update to the meta table, so creating a database using table.sql will not yield the same result as running the patch file on an existing database

epaule:
no table.sql in modules/t has any content for the meta table. Isn't that loaded from meta.txt ?

magaliruffier:
the original table.sql in the sql directory
if you look at previous patches, we add a patch file that modified meta and the actual patch, and we update table.sql to include the same changes

magaliruffier:
It's looking fine for merge now

PR #182, opened 2017-02-01
Fix in mapper for insertion SeqEdits

willmclaren:


magaliruffier:
We need to ensure the GTF dump still works for eg! before we can pull the change in, see https://www.ebi.ac.uk/panda/jira/browse/ENSCORESW-1318

willmclaren:
OK. Probably a test to cover their use case would be useful too.

magaliruffier:
the strange EG species still dump out ok with this change, maybe this was due to another bug that's been fixed and all is well
I am happy for the change to go in given it fixes your problem, and if we uncover something else that breaks, we'll deal with it then

PR #183, opened 2017-02-14
Make SeqRegionSynonymAdaptor compatible with BaseAdaptor

willmclaren:
I wanted to use fetch_all() on SeqRegionSynonymAdaptor but couldn't because it doesn't implement _tables(), _columns() and _objs_from_sth().


nerdstrike:
It's a fair criticism, and a reasonable fix

PR #184, opened 2017-02-15
Do not skip first mapping when making chain file

james-monkeyshines:
I think this is a bug, but the logic is a bit complicated, so I'm not entirely sure. As I see it, the mappings are pulled into an array, then iterated over with an index, $i. The first mapping (ie $i == 0) is skipped, which seems like an error; the first mapping should be just as valid as the rest (?)


magaliruffier:
that does look reasonable
this script was pulled out of the depths of Andy's brain a long time ago, so hard to know what the logic was behind skipping the first one

andrewyatz:
I would assume I was worried that logic later on would have put in the first mapping as a single event on its own or something similar. Anyway if it looks fine then great.

The logic of producing these is horrible. It all boils down to keep on reading the mappings until there's an event that forces you to close off a block and start a new one. The challenge is that you don't know the next event until you observe it and then decide if it requires you to abort the current block or extend it 🤕

PR #185, opened 2017-03-02
Escape seq_region_name before querying the synonyms

thibauthourlier:
If seq_region_name is not escaped, the API might return the wrong slice if two synonyms are the same except for the '_' character: Contig10_1/Contig1041


magaliruffier:
looks like that syntax is not compatible with SQLlite, that needs fixing

nerdstrike:
While we test on sqlite, we are in principle SQL agnostic. Would you mind giving this a go?

`$esc = $dbh->get_info( 14 );  # SQL_SEARCH_PATTERN_ESCAPE
  $search_pattern =~ s/([_%])/$esc$1/g;`

That way DBI tells you what the escape char is for the driver.

thibauthourlier:
Looks like it worked!! Thanks Kieron

PR #186, opened 2017-03-17
fix: use write credentials for creating indices

Zhicheng-Liu:
Before this change, the script was using read credentials which will fail.

@premanand17 , could you please check and merge if there is no problem?


lairdm:
Looks sane.

PR #188, opened 2017-03-23
Undefine seq when changing start or end unless an adaptor is present

thibauthourlier:
This could be useful when creating a Translation but a change to the start or end is need for a stop codon for example and $obj->seq has already been called.

### DIFF
--- a/modules/Bio/EnsEMBL/Translation.pm
+++ b/modules/Bio/EnsEMBL/Translation.pm
@@ -206,6 +206,9 @@ sub start{
       my $value = shift;
       
       $obj->{'start'} = $value;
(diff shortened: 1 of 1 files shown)


magaliruffier:
it would be good if you could give us a bit more context on why that change is needed
and test cases ;-)

magaliruffier:
can we please get more information on this PR, as well as test cases?

thibauthourlier:
It looks like the information needed will have to be long so it will wait until I have time to write it

magaliruffier:
Ok, I will close this for the moment, feel free to submit a new PR when you have test cases

PR #191, opened 2017-03-31
Ensvar 562

at7:
I deleted deprecated methods from the Slice module and the respective test cases. The plan was to also delete StrainSlice and StrainSlice adaptor. There are however too many dependencies in the core API e.g. grep -r StrainSlice ensembl/modules/* which I don't understand.


at7:
I noticed I need to fix some variation API code first. I'll send another PR later.

at7:
was a quick fix. sorry

magaliruffier:
If you don't know where StrainSlice is used, I am not sure who would ;-)
As far as I can see, we could retire simultaneously AlignStrainSlice, StrainSliceAdaptor and StrainSlice
all other mentions are just adding special treatment or are just not implemented

at7:
I will move the remaining strainSlice related code to variation and deprecate more code in the core api in another PR. This one already deleted a lot of variation related code.

at7:
Not save to remove the methods. I remembered an email from Dan Sheppard, where he pointed out that he had to revert to using the method from the core instead variation API. I cannot see why the variation method is not working, but will need my sandbox first to look into this. I included you in the email, maybe you spot something.

PR #192, opened 2017-04-03
Update Feature.pm

ens-carlos:
Check if "slice" exists before calling "slice->is_circular()".


nerdstrike:
How does a feature have coordinates if it lacks a slice?

ens-carlos:
I had to add that check for the PDB protein features.

nerdstrike:
So your features have a position but no frame of reference? Are they coordinates in protein residues?

ens-carlos:
Yes, that's right. They are coordinates in protein residues.

nerdstrike:
Resolved in person

PR #193, opened 2017-05-03
Set an object's default version to 1 when it is created,…

james-monkeyshines:
…rather than when it is stored, to allow undef versions (which can then be stored as NULLs in the database).


magaliruffier:
that looks fine to me, thanks for the change

PR #194, opened 2017-05-10
Return all adaptors in the registry in get_all_DBAdaptors if no args …

Zhicheng-Liu:
…supplied

Previously, when no args (either species or group) supplied, the implementation
at get_all_DBAdaptors requires looping through '_DBA' array and returns an array
of each DBA's reference. It takes long time if there are large number of DBAs
(e.g. >0.5s for ~88k DBAs).

This change returns the reference to the '_DBA' array itself directly in such cases.


Zhicheng-Liu:
I was meant to compare the two references are pointing to the same object. Thought 'is' is enough and tested locally.
I will look into this later.

Zhicheng-Liu:
Looks like Travis is happy with new test (cmp_deeply). Many thanks for the hint, @avullo

avullo:
Glad to hear everything's fine now, let's merge then.

Thanks

Alessandro

PR #195, opened 2017-05-11
Return all adaptors in the registry in get_all_DBAdaptors if no args supplied

Zhicheng-Liu:
The changes included in this PR are the same as those in the PR #194 . I cherry picked all the commits from there.

I thought it would be good for these changes to go into release/88 too.


magaliruffier:
are you planning to push those changes to the live webservers?
if so, I can't see any objection

Zhicheng-Liu:
Yes, that was the aim.

avullo:
And so be it, thanks.

Zhicheng-Liu:
Thank you all!

PR #196, opened 2017-05-17
Remove $registry_register{_TYPE}

Zhicheng-Liu:
as it does not seem to be used anywhere. Furthermore, the same information
can be derived from $registry_register{_SPECIES}.

This piece of duplicated information take up a lot of spaces for a registry
with large number of species (e.g. ~600MB for Ensembl Bacteria at ~44k species).


nerdstrike:
Astonishing! It certainly doesn't seem to be invoked anywhere

lairdm:
Seems safe... here we go...

PR #197, opened 2017-05-17
Remove $registry_register{_TYPE}

Zhicheng-Liu:
as it does not seem to be used anywhere. Furthermore, the same information
can be derived from $registry_register{_SPECIES}.

This piece of duplicated information take up a lot of spaces for a registry
with large number of species (e.g. ~600MB for Ensembl Bacteria at ~44k species).


Zhicheng-Liu:
The changes included in this PR are the same as those in the PR #196 . I thought it would be good for these changes to go into release/89 too.

avullo:
And so it goes ...

PR #206, opened 2017-07-10
bugfix: get_all_DBAdaptors() should always be returning an array-ref

muffato:


nerdstrike:
While your contribution is bang on, I wouldn't be at all surprised if this had consequences.

muffato:
I think it may actually be fine because @Zhicheng-Liu added those lines quite recently ( https://github.com/muffato/ensembl/commit/522bb6bb8f6a570da202d9cc341e590c25281647 ) so hopefully the number of people who rely on the function returning undef is low

nerdstrike:
I hope so too.

PR #207, opened 2017-07-12
$species must be defined in Registry::add_DBAdaptor

muffato:
TBH, most of the Registry methods don't properly check all their arguments, so this PR is not making the module much safer. I'm adding this test because the lack of it bit me and since the same safety check is in `get_DBAdaptor` I thought it could rightly be in `add_DBAdaptor` too


nerdstrike:
Agreed.

PR #208, opened 2017-07-20
Allow translation start and stop to be defined within a transcript-level seq-edit

james-monkeyshines:
The way that translation start and end are defined in the core db schema make it impossible to insert sequence at the start or end that contains both UTR and CDS; in other words, translation cannot start or stop within a seq-edit.

Where you have a fragmented assembly, and therefore need to add seq_edits, it is common to have this situation, since genes will tend to be truncated, at one or both ends, rather than having missing coding sequence in the middle. The best workaround with the current code is to remove the UTRs, which is obviously not ideal.

There are probably much cleverer, and correspondingly more complicated, ways to address this. The simple/stupid solution I propose here is for a translation_attrib which overrides whatever is derived from the translation table, for start and/or stop. This attrib is only applied in the presence of seq-edits, so shouldn't slow down the module by looking for an attrib that won't exist for most species. I used an underscore prefix for the attrib codes, since that seemed to be the convention for seq-edit things. If this PR is accepted, I'll add the attribs to the production db.

I'd like this to be available for VectorBase, which is currently running on release 88 code. So if possible (and assuming you are happy to accept this change, of course) please could this be cherry-picked onto release/88 and release/89.

### DIFF
--- a/modules/Bio/EnsEMBL/Transcript.pm
+++ b/modules/Bio/EnsEMBL/Transcript.pm
@@ -974,13 +974,20 @@ sub cdna_coding_start {
     # adjust cdna coords if sequence edits are enabled
     if($self->edits_enabled()) {
       my @seqeds = @{$self->get_all_SeqEdits()};
-      # sort in reverse order to avoid adjustment of downstream edits
-      @seqeds = sort {$b->start() <=> $a->start()} @seqeds;
+      if (scalar @seqeds) {
+        my $transl_start = $self->get_all_Attributes('_transl_start');
+        if (@{$transl_start}) {
+          $start = $transl_start->[0]->value;
+        } else {
+          # sort in reverse order to avoid adjustment of downstream edits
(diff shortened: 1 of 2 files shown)


magaliruffier:
It looks like a workable fix. Be aware that we are planning to review the whole seq-edit madness later this year, so we might change the way it is behaving, but with the tests, we should be able to ensure we maintain the expect behaviour for your use case.

nerdstrike:
Code looks good to me.

james-monkeyshines:
Thanks for merging; if it's not too much bother, please could I have this backported to release/88 and release/89 branches.

lairdm:
We don't generally like to do that, but since you asked nicely, done.

james-monkeyshines:
On Tue, 08 Aug 2017 12:56:19 +0000 (UTC)
Matthew Laird <notifications@github.com> wrote:

> We don't generally like to do that, but since you asked nicely, done.
> 

Thanks!

PR #210, opened 2017-08-09
ENSCORESW-1910

jimmyebi:


magaliruffier:
will be replaced with PR with better commit history

PR #211, opened 2017-08-14
ENSCORESW-1910

jimmyebi:


avullo:
It's stil registry_public_core, but we know the problem is with the release version you're trying to simulate. If we wait Thu the problem should vanish so bear with me. Meanwhile, you could update the failing test so that the message is not misleading anymore, i.e. split into two separate tests.

avullo:
Why has this been closed?

magaliruffier:
it needs cleaning up and will be submitted as a new PR

PR #212, opened 2017-08-15
ENSCORESW-1910

jimmyebi:


magaliruffier:
changes are now doing what is expected, waiting until after release to merge

sql/patch_90_91_b.sql
```diff
@@ -0,0 +1,27 @@
...
+-- limitations under the License.
+
+# patch_90_91_b.sql
+#
+# Title: Add species_id field
+#
+# Description:
+# Add species_id column in the mapping_session table is a possible solution to help identifying the strain with the multiSpecies database mapping sessions.
+
+ALTER TABLE mapping_session ADD COLUMN species_id INT(10) UNSIGNED NOT NULL DEFAULT 1;
```
lairdm: This alter statement should be explicit:
ADD COLUMN species_id INT(10) UNSIGNED AFTER ....
since our data loading from flat files, column order matters. Ensure it's always in the correct order.

lairdm:
Two picky but not show stopper comments and one more strong change request.

Can I be demanding and ask the a and b patches be separate commits as previously discussed?
Also the new column in the mapping_session table should probably be added after mapping_session_id making it the second column. There's a performance reason for this, it doesn't have to scan the entire row through all the varchars to find this loan int that could become more important in future multi-species databases. Now essential, but a nice to have if we're making the change.

But the forceful request, please explicitly add an AFTER clause to the ALTER statement so the column is guaranteed to be in the correct place. This is essential for our loading from flat files where column order matters.

PR #215, opened 2017-09-26
Enabled perlcritic on misc-scripts

muffato:
I don't know if misc-scripts was intentionally skipped, but it contains 76 non-compliant perl files


magaliruffier:
misc-scripts is full of legacy scripts which are not much used/maintained any more, so it is not worth going back through them to make them perlcritic compliant
I think we will leave the test out for the moment, thanks for the suggestion though

PR #216, opened 2017-10-04
Removed special treatment for funcgen adaptors, coord systems have be…

mn1:
…en removed from funcgen databases.


magaliruffier:
Hooray to the death of the funcgen coord_system ;-)
Looks fine to me

PR #218, opened 2017-10-24
Extra parameters for the REST endpoint 'overlap/translation'. Will be…

ens-lgil:
… needed for a mapping to the PDB coordinates in a tool/feature using it

### DIFF
--- a/modules/Bio/EnsEMBL/ProteinFeature.pm
+++ b/modules/Bio/EnsEMBL/ProteinFeature.pm
@@ -207,10 +207,13 @@ sub summary_as_hash {
   my %summary;
   $summary{'type'} = $self->analysis->db;
   $summary{'id'} = $self->display_id;
-  $summary{'start'} = $self->start,
-  $summary{'end'} = $self->end,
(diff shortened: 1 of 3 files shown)


ens-lgil:
I believe I will have to update the test on the ensembl-rest repo

avullo:
Hi Laurent,
yes, but can you also add a relevant test to proteinFeature.t and/or proteinFeatureAdaptor.t if that's not too painful?

thanks

ens-lgil:
@avullo, you don't have tests about the method 'summary_as_hash'.
Do you mean I need to create tests for this method ?

avullo:
That would be ideal, I saw this an an opportunity to start covering this uncovered part of the API, especially considering coverage has, even marginally decreased. Feel free to do or not do that. It could be just testing the newly added fields.

ens-lgil:
Ok, I just added few tests on "summary_as_hash"

avullo:
Brilliant, well done!
Let's wait the tests then it goes in.

thanks a mil

PR #220, opened 2017-11-03
bugfix: not all users require a password

muffato:
## Requirements

## Use case

I am not able to run `db-space.pl` with the ensro user, and I'd prefer not using ensadmin unless really needed

## Description

The script doesn't die any more if the password is not provided. It then only adds the `-p` option to the mysql command line if there is actually a password (`-p` alone makes mysql prompt for a password)

## Benefits

I can call  `db-space.pl` with the `ensro` to generate the disk usage reports

## Possible Drawbacks

The error message for someone running the script with `ensadmin` and no password will be different (it will come from . I don't expect anyone to be relying on that

## Testing

No, there were no tests on that script

N/A

No.


avullo:
Excellent, thanks for adhering to the template structure :+1: 

P.S. The "Requirements" is just communication to the puller, it can be discarded.

muffato:
I have added another commit to make the script executable

lairdm:
Approved.

muffato:
Super, thank you !

PR #223, opened 2017-11-30
Throw if features array is empty and moved checks to the top.

ens-carlos:


modules/Bio/EnsEMBL/BaseAlignFeature.pm
```diff
@@ -676,13 +676,15 @@ my $message_only_once = 1;
 sub _parse_ensembl_features {
   my ($self,$features ) = @_;
 
-  my $query_unit = $self->_query_unit();
-  my $hit_unit = $self->_hit_unit();
-
   if (ref($features) ne "ARRAY") {
     throw("features must be an array reference not a [".ref($features)."]");
+  } elsif (scalar(@$features) <= 0) {
```
avullo: I've skipped this last time, but scalar cannot return a value less than zero

lairdm:
Is there somewhere upstream that needs further error checking? ie. how did the situation of an empty set of features transpire?

And is there a regression test that can be added to the test suite to check that an error is properly thrown when the features array is empty.

Thanks.

ens-carlos:
Probably yes there could be another check upstream but I think it's worth adding that check rather than assuming the array won't be empty just before trying to access the first feature in the array later on within the same sub. I think it is a change in line with the current check of ARRAY type.

Sorry I haven't looked into the tests in detail. Could you point me to the file where the ARRAY type exception is checked so I can add a similar test?

Cheers.

avullo:
@ens-carlos have a look here: 
https://github.com/Ensembl/ensembl/blob/release/90/modules/t/transcriptSupportingFeatureAdaptor.t#L42

avullo:
Can you pass the align type so that it doesn't issue a warning which the test doesn't expect? it is failing otherwise.

thanks

avullo:
Great, thanks, finally it can go in! :+1:

ens-carlos:
Thank you!

PR #226, opened 2018-02-05
Alignment storage

premanand17:
## Requirements

## Description

Updates to support ProteinFeature data imported from GIFTS.
 - Added script to import data from GIFTS database in to protein_feature table
- Methods to store cigar_line and align_type to ProteinFeature
- Methods to handle MD Tag string parsing and deducing alignment_length and compute alignment strings
- Updated ProteinFeatureAdapter to extend BaseAlignFeatureAdaptor to support alignment related methods
- Updated ProteinFeature to extend BaseAlignFeature to support alignment related methods
- Added some utility methods to fetch gifts_import data

## Use case

Problem: Current API can only handle 'ensembl' alignment_type and cannot handle mdz string alignment_type.
Also the ProteinFeature object is not 'alignment aware'. The proposed changes, will support handling of other alignment types and also will enable to represent protein-protein alignment info in ProteinFeature object. 

## Benefits

Can import and store gifts data thereby unifying genomic and proteomics data
Can store MD Tag alignment string, which is a better representation than CIGAR or CIGARPLUS

## Possible Drawbacks

If new alignment type has to be added in feature, have to provide methods to handle that alignment string.

## Testing

Yes.

Pass, of course.

Yes.


misc-scripts/gifts_import/populate_protein_feature.pl
```diff
@@ -0,0 +1,350 @@
...
+    host   => $pfhost,
+    port   => $pfport,
+    dbname => $pfdb
+  };
+
+  return ( $giftsDB, $protfeatureDB );
+}
+
+
+populate_protein_feature_db();
```
lairdm: Could this be changed not to intersperse script logic with subroutine declarations please.

misc-scripts/gifts_import/populate_protein_feature.pl
```diff
@@ -0,0 +1,350 @@
...
+  }
+  
+  my $insert_container = build_insert_sql($gifts_dbh, $protfeature_dbh);
+
+  
+  my $rows_inserted = 0;
+  my @failed_stmts = [];
+  unless($dryrun){
+
+  foreach my $insert_stmt (@$insert_container) {
```
lairdm: Just for readability could the indentation be fixed please

lairdm: For future readability, could the indentation of this with respect to the dryrun if statement be fixed please.

misc-scripts/gifts_import/populate_protein_feature.pl
```diff
@@ -0,0 +1,350 @@
...
+
+
+  my $import_sth = $gifts_dbh->prepare($select_sql);
+  $import_sth->execute;
+  my $rows = $import_sth->rows;
+
+  my $start = 0;
+  my $insert_values;
+  my $max_rows_in_insert = 9999;    #Batch size of 10,000
+  # my $max_rows_in_insert = 10;    #Batch size of 10
```
lairdm: If it's permanently commented out let's remove it

misc-scripts/gifts_import/populate_protein_feature.pl
```diff
@@ -0,0 +1,350 @@
...
+  $protfeature_dbh->disconnect();
+  
+}
+
+sub annotate_ref{
+  my $ref = shift;
+  my $ensp_id = $ref->[0];
+  my $ensp = $translation_adaptor->fetch_by_stable_id( $ensp_id ) ;
+  if($ensp){
+  $ref->[0] = $ensp->dbID;
```
lairdm: Again, indentation pickness if this is a new script

misc-scripts/gifts_import/populate_protein_feature.pl
```diff
@@ -0,0 +1,350 @@
...
+  $ref->[1] = 1;
```
lairdm: The discussion you're hearing behind you, we're not sure if this is a good idea, that object coming back from DBI really isn't yours to mess with. It's likely alright if it's not crashing on you so far, but it just feels a little wrong since this whole fetchrow_arrayref lays somewhere between the perl and XS C layer of the driver.

In addition, could we have a bit of inline documentation for the future, why are you just putting '1' here. Why in the next line is is length()-1.

misc-scripts/gifts_import/populate_protein_feature.pl
```diff
@@ -0,0 +1,350 @@
...
+  my $start = 0;
+  my $insert_values;
+  my $max_rows_in_insert = 9999;    #Batch size of 10,000
+  # my $max_rows_in_insert = 10;    #Batch size of 10
+  my $rows_in_insert     = 0;
+
+  my @insert_container = ();
+  my $miss = 1;
+  while ( my $ref = $import_sth->fetchrow_arrayref ) {
+     my $not_ensp =1;
```
lairdm: Not that it seems to be used later, but why is this starting at 1 if it's an incrementer for non-matching lines?

misc-scripts/gifts_import/populate_protein_feature.pl
```diff
@@ -0,0 +1,350 @@
...
+  }
+}
+
+  if ($insert_values) {
+    my $insert_sql_with_values = $insert_sql . $insert_values;
+    push( @insert_container, $insert_sql_with_values );
+  }
+  
+  print "Number of rows fetched from GIFTS $rows \n";
+  return \@insert_container;
```
lairdm: How large is this array going to get over the next few years? I don't see any version in the SELECT that originally populates it. Is there a reason it's structured this way rather than directly selecting then inserting as you iterate over the result set?

In addition, the join(','...) with quote() isn't the safest method. If you are going to return an array of elements to insert, it'd probably be better to return a prepared insert statement with parametarized placeholder for the values plus an array of arrays of the values. Then use DBI to do all the proper quoting and filling in the placeholders in the execute() call by passing this array of values at that time.

misc-scripts/gifts_import/populate_protein_feature.pl
```diff
@@ -0,0 +1,350 @@
+#!/usr/bin/env perl
```
lairdm: A little more inline documentation overall please, just so whoever has to take this up in a few years has a better understanding. It's all pretty self evident, but some things like the SELECT/INSERT, why some values are hardcoded, etc just to make life easier.

modules/Bio/EnsEMBL/BaseAlignFeature.pm
```diff
@@ -513,7 +515,6 @@ sub _parse_ensembl_cigar {
   throw("No cigar string defined in object") if(!defined($string));
 
   my @pieces = ( $string =~ /(\d*[MDI])/g );
-  #print "cigar: ",join ( ",", @pieces ),"\n";
```
lairdm: Remove

modules/Bio/EnsEMBL/BaseAlignFeature.pm
```diff
@@ -270,6 +270,8 @@ sub alignment_length {
 
   if ($self->{'align_type'} eq 'ensembl') {
     return $self->_ensembl_cigar_alignment_length();
+  } elsif ($self->{'align_type'} eq 'mdtag') {
```
lairdm: We have a problem with a lack of comments in our code. As we go through updating things can we take the opportunity to put some inline comments? For example here just briefly explaining what those alignment types are.

modules/Bio/EnsEMBL/BaseAlignFeature.pm
```diff
@@ -1009,7 +1012,303 @@ sub _query_unit {
...
+    $alignment_length = $self->_get_mdz_alignment_length($chunks);
+    $self->{'_alignment_length'} = $alignment_length;
+  }
+  return $self->{'_alignment_length'};
+}
+
+=head2 _get_mdz_chunks
+
+  Arg [1]    : mdz string
+  Description: parses the mdz string and group it according the type
```
lairdm: Sorry dude. Again this sort of thing, if I wasn't at your talk on this, I wouldn't know what the helk an mdz string is. Let's raise the bar from the sins of past ensembl developers :)

modules/Bio/EnsEMBL/BaseAlignFeature.pm
```diff
@@ -1009,7 +1012,303 @@ sub _query_unit {
...
+
+=cut
+
+sub _get_mdz_chunks {
+  my $self = shift;
+
+  my $mdz_string = shift;
+  my $alignment_length = 0;
+  my @chunks;
+    if ( $mdz_string =~ /^MD:Z:(.*)/ ){
```
lairdm: Perhaps that should be ^MD:Z:(.+), because with the * "MD:Z:" would match.

modules/Bio/EnsEMBL/BaseAlignFeature.pm
```diff
@@ -1009,7 +1012,303 @@ sub _query_unit {
...
+      if($mdtag =~/^\d+$/){
+        return [$mdtag];
+      }else{
+        my @char_arr = split('',$mdtag);
+        for(my $i=0; $i< scalar(@char_arr); $i++){
+        my $char = $char_arr[$i];
+
+        #determine the type and next type
+        my $type = $self->_get_mdz_chunk_type($char);
+        my $next_char = $char_arr[++$i];
```
lairdm: What if the string is 1 character long? Array out of bounds.

modules/Bio/EnsEMBL/BaseAlignFeature.pm
```diff
@@ -1009,7 +1012,303 @@ sub _query_unit {
...
+=cut
+
+sub _get_mdz_chunk_type{
+  my $self = shift;
+  my $char = shift;
+
+  my $type;
+  if($char eq '^'){
+    $type = "del";
+  } elsif($char =~ /\d+/){
```
lairdm: You're only passing in single characters, so why /d+?

modules/Bio/EnsEMBL/BaseAlignFeature.pm
```diff
@@ -1009,7 +1012,303 @@ sub _query_unit {
...
+
+        #determine the type and next type
+        my $type = $self->_get_mdz_chunk_type($char);
+        my $next_char = $char_arr[++$i];
+        my $next_type = $self->_get_mdz_chunk_type( $next_char);
+
+        my $chunk = $char;
+
+        #group chars in mdz based on the type, break if it differenc and push to chunks
+        while($type eq $next_type and $i <= $#char_arr){
```
lairdm: I think you might be over thinking this with a lot of ++ and -- index variable. So basically you want to start a new bucket every time the "type" changes (^, digit, alpha). You haven an array with all the character split up. You could go through popping off that array, .= the next character on to the current "chunk" bucket unless the class changes. Otherwise, push the current bucket on to the pile and start another. It would likely be a little less cluttered.

modules/Bio/EnsEMBL/BaseAlignFeature.pm
```diff
@@ -1009,7 +1012,303 @@ sub _query_unit {
...
+  }
+  return $type;
+}
+
+
+
+=head2 alignment_strings
+
+  Arg [1]    : list of string $flags
+  Example    : $pf->alignment_strings or
```
lairdm: or...? :)

modules/Bio/EnsEMBL/BaseAlignFeature.pm
```diff
@@ -1009,7 +1012,303 @@ sub _query_unit {
...
+
+=cut
+
+
+sub alignment_strings {
+  my $self = shift;
+
+  #Translations
+  my $transl_adaptor = $self->adaptor->db->get_TranslationAdaptor();
+  my $transl_object = $transl_adaptor->fetch_by_dbID($self->translation_id);
```
lairdm: translation_id is in ProteinFeature, not BaseAlignFeature, this needs to be moved to ProteinFeature to prevent someone accidentally calling it and the code going bang.

modules/Bio/EnsEMBL/BaseAlignFeature.pm
lairdm: This issue still exists.

modules/Bio/EnsEMBL/BaseAlignFeature.pm
```diff
@@ -1009,7 +1012,266 @@ sub _query_unit {
...
+      my $mdtag = $1;
+
+      #if start to end is a number then all match return as it is
+      if($mdtag =~/^\d+$/){
+        return [$mdtag];
+      }else{
+         my @char_arr = split('',$mdtag);
+         my $cur_char = undef;
+
+         while(defined($cur_char = shift @char_arr)){
```
lairdm: Is the defined needed? Won't shift on an empty array get undef and eval as false?

modules/t/baseAlignFeature.t
```diff
@@ -0,0 +1,150 @@
...
+sub qc_check_sequence{
+  my ($input_seq, $target_seq) = @_;
+  
+  $target_seq =~ s/-//g;
+  return 0 if $input_seq ne $target_seq;
+  
+  return 1;
+
+}
+
```
avullo: How about testing the behaviour with incorrect alignment strings?

modules/t/proteinFeatureAdaptor.t
```diff
@@ -134,9 +135,102 @@ ok(scalar @pfs > 0);
...
+my $test_dbID = 242847;
+my $test_hitname = 'Q86UU9';
+my $transl_id = 21739;
+
+# fetch_all_by_translation_id
+ok($pfa && $pfa->isa('Bio::EnsEMBL::DBSQL::ProteinFeatureAdaptor'));
+my $trl_features = $pfa->fetch_all_by_translation_id($transl_id);
+
+my $trl_feature;
+do{ $trl_feature = shift @$trl_features} until($trl_feature->analysis()->logic_name eq $test_logic_name);
```
avullo: I think you can more efficiently and compactly use grep here.

modules/t/proteinFeatureAdaptor.t
```diff
@@ -134,9 +135,102 @@ ok(scalar @pfs > 0);
...
+ok($trl_feature && $trl_feature->isa('Bio::EnsEMBL::ProteinFeature'));
+ok($trl_feature->analysis()->logic_name eq $test_logic_name, 'Got the right logic name ' .$test_logic_name);
+
+ok($trl_feature->cigar_string eq $test_cigar_string, 'Got the right cigar string ' . $test_cigar_string);
+ok($trl_feature->align_type eq $test_align_type, 'Got the right align type ' . $test_align_type );
+
+# fetch_all_by_translation_id and logic_name
+$trl_features = $pfa->fetch_all_by_translation_id($transl_id, "gifts_import");
+
+do{ $trl_feature = shift @$trl_features} until($trl_feature->analysis()->logic_name eq $test_logic_name);
```
avullo: Same as before.

modules/Bio/EnsEMBL/BaseAlignFeature.pm
avullo: I agree, this should be addressed.

modules/Bio/EnsEMBL/DBSQL/ProteinFeatureAdaptor.pm
```diff
@@ -59,10 +59,11 @@ use strict;
 
 use Bio::EnsEMBL::DBSQL::BaseAdaptor;
 use Bio::EnsEMBL::ProteinFeature;
+use Bio::EnsEMBL::DBSQL::BaseAlignFeatureAdaptor;
 use Bio::EnsEMBL::Utils::Exception qw(throw deprecate warning);
 
 use vars qw(@ISA);
-@ISA = qw(Bio::EnsEMBL::DBSQL::BaseAdaptor);
+@ISA = qw(Bio::EnsEMBL::DBSQL::BaseAlignFeatureAdaptor);
```
tgrego: `use parent` is much more recommended than directly modifying the `@ISA`

modules/Bio/EnsEMBL/ProteinFeature.pm
```diff
@@ -56,10 +56,11 @@ package Bio::EnsEMBL::ProteinFeature;
 use strict;
 
 use Bio::EnsEMBL::FeaturePair;
+use Bio::EnsEMBL::BaseAlignFeature;
 use Bio::EnsEMBL::Utils::Argument qw(rearrange);
 
 use vars qw(@ISA);
-@ISA = qw(Bio::EnsEMBL::FeaturePair);
+@ISA = qw(Bio::EnsEMBL::BaseAlignFeature);
```
tgrego: again, use parent is better.
`use parent qw(Bio::EnsEMBL::BaseAlignFeature);`

instead of
`use vars qw(@ISA);`
`@ISA = qw(Bio::EnsEMBL::BaseAlignFeature);`

modules/Bio/EnsEMBL/BaseAlignFeature.pm
```diff
@@ -1009,7 +1012,259 @@ sub _query_unit {
...
+sub _mdz_alignment_string {
+  my ($self, $input_seq, $mdz_string) = @_;
+
+  my $chunks = $self->_get_mdz_chunks($mdz_string);
+
+  my $target_seq = "";
+  my $query_seq = "";
+
+   #Handle first chunk, which is always a num
+   my $first_chunk = $$chunks[0];
```
tgrego: `$$chunks[0]`
is equivalent to the more modern and pleasant to the eye
`$chunks->[0]`

`$$` gets a bit confusing easily in my opinion, however it is a matter of style and both ways work... 
It's up to you what you want to use, but thought it would be useful to let you know both ways.

modules/Bio/EnsEMBL/ProteinFeature.pm
```diff
@@ -82,21 +83,33 @@ use vars qw(@ISA);
 
 =cut
 
+
 sub new {
   my $proto = shift;
```
tgrego: It's recommended to always unpack `@_` first.
this means:
`
 sub new {
   my ($proto, @args) =   @_;
`
and from there you can use `@args` and leave the `@_` alone.

modules/Bio/EnsEMBL/ProteinFeature.pm
```diff
@@ -213,9 +233,102 @@ sub summary_as_hash {
...
+  if ($self->align_type eq 'mdtag') {
+    if(defined $seq && defined $self->cigar_string){
+      return $self->_mdz_alignment_string($seq,$self->cigar_string);
+    }else{
+      warn "sequence or cigar_line not found for  " .  $self->translation_id;
+    }
+  } else {
+    throw("alignment_strings method not implemented for " . $self->align_type);
+  }
+  return undef;
```
tgrego: don't `return undef;`
just `return;`

modules/Bio/EnsEMBL/ProteinFeature.pm
```diff
@@ -213,9 +233,102 @@ sub summary_as_hash {
...
+}
+
+
+sub transform {
+  my $self = shift;
+
+  $self->throw( "ProteinFeature cant be transformed directly as".
+		" they are not on EnsEMBL coord system" );
+}
```
tgrego: always end a subroutine with `return;`

modules/Bio/EnsEMBL/ProteinFeature.pm
```diff
@@ -213,9 +233,102 @@ sub summary_as_hash {
...
+               1 as the 'unit' used for the hit sequence.
+  Returntype : int
+  Exceptions : none
+  Caller     : Bio::EnsEMBL::BaseAlignFeature
+  Status     : Stable
+
+
+=cut
+
+sub _hit_unit {
```
tgrego: Why not constants instead of _hit_unit and _query_unit?

premanand17: The hit_unit and query_unit varies dynamically depending upon the type of the align feature (dna or protein) and they are not constants theoretically.

PR #228, opened 2018-02-16
Biotype in Core schema

tgrego:
## Requirements

## Description

From ensembl-production pull request #97, the biotype table has been updated to master_biotype and added to the script that copies over data to the core database.
The biotype table is thus created in the core schema.

A BiotypeAdaptor was created to interface with this new table.
A Biotype class was created that allows the creation of biotype objects, containing all the information on a biotype (such as the biotype_group or the so_acc).

Gene and Transcript classes have biotype methods, these have been rewritten to use and retrieve Biotype objects instead of using the string of the biotype name.

Tests have been added in the biotype.t file. 

## Use case

The biotype column in the gene and transcript tables made it easy to get the biotype name of a feature. However getting extra related information, such as the biotype group or SO acc, was not easy.

## Benefits

Having the biotype table in the core schema makes it easy to retrieving this data.
There is no need to query the production database, or to use the SequenceOntologyMapper.

## Possible Drawbacks

The changes made are expected to be compatible with legacy behaviour.

## Testing

Yes.

Pass, of course.

Yes.


nerdstrike:
The schema change commits shouldn't really be in this pull request. They do not belong. It is the result of reusing an existing branch, rather than a clean source. However, the changes _should_ disappear once we can merge in the schema version update.

modules/Bio/EnsEMBL/Gene.pm
```diff
@@ -1527,6 +1507,42 @@ sub havana_gene {
...
 }
 
+=head2 biotype
+
+  Arg [1]    : Arg [1] : (optional) String - the biotype to set
+  Example    : my $biotype = $gene->biotype;
+               my $biotype = $gene->biotype('protin_coding');
+  Description: Returns the Biotype object of this gene.
+               When no biotype exists, defaults to 'protein_coding'.
+  Returntype : Bio::EnsEMBL::Biotype
```
avullo: This is not returning an object in the default case.

modules/t/biotype.t
```diff
@@ -0,0 +1,79 @@
...
+ok($biotype->isa("Bio::EnsEMBL::Biotype"));
+is($biotype->object_type, 'transcript', 'Biotype is from Transcript object');
+is($biotype->name, 'protein_coding', 'Biotype name is protein_coding');
+is($biotype->biotype_group, 'coding', 'Biotype group is coding');
+is($biotype->so_acc, 'SO:0000234', 'Biotype protein_coding refers to SO:0000234');
+
+# test fetch biotypes of object_type gene
+my $biotypes = $biotype_adaptor->fetch_all_by_object_type('gene');
+is(ref $biotypes, 'ARRAY', 'got an array');
+is(scalar @{$biotypes}, '2', 'of size 2');
```
avullo: For an extra, check the array content, e.g. cmp_deeply.

modules/Bio/EnsEMBL/DBSQL/BiotypeAdaptor.pm
```diff
@@ -0,0 +1,190 @@
...
+
+sub fetch_by_name_object_type {
+  my ($self, $name, $object_type) = @_;
+
+  my $constraint = "b.name = ? AND b.object_type = ?";
+  $self->bind_param_generic_fetch($name, SQL_VARCHAR);
+  $self->bind_param_generic_fetch($object_type, SQL_VARCHAR);
+  my ($biotype) = @{$self->generic_fetch($constraint)};
+
+  if (!defined $biotype) {
```
nerdstrike: I think this might be too helpful. It can lead to working with biotypes that aren't in the DB without knowing it. I would prefer an exception and a more wary calling code.

modules/Bio/EnsEMBL/DBSQL/BiotypeAdaptor.pm
```diff
@@ -0,0 +1,190 @@
...
+  Exceptions : none
+
+=cut
+
+sub fetch_all_by_object_type {
+  my ($self, $object_type) = @_;
+
+  my $constraint = "b.object_type = ?";
+  $self->bind_param_generic_fetch($object_type, SQL_VARCHAR);
+  my @biotypes = @{$self->generic_fetch($constraint)};
```
nerdstrike: What happens here if the query returns no rows? Might be better not to deref automatically and either generate an exception or return []

modules/Bio/EnsEMBL/DBSQL/GeneAdaptor.pm
```diff
@@ -1200,7 +1200,7 @@ sub store_alt_alleles {
       if ($gene->slice->is_reference()) {
           $type{'IS_REPRESENTATIVE'} = 1;
       }
-      push @$aa_record, \%type; 
+      push @$aa_record, \%type;
```
nerdstrike: At this point I am now bored of scanning for substantive changes amongst all the deletions of trailing whitespace.

modules/Bio/EnsEMBL/Gene.pm
```diff
@@ -171,8 +170,10 @@ sub new {
   $self->external_status($external_status)
     if ( defined $external_status );
   $self->display_xref($display_xref) if ( defined $display_xref );
-  $self->biotype($type)              if ( defined $type );
```
nerdstrike: This counts as an interface change. In the unlikely event of users setting biotypes on gene objects, you have just broken one of the ways of doing it.

modules/Bio/EnsEMBL/Gene.pm
```diff
@@ -1527,6 +1507,42 @@ sub havana_gene {
...
+
+  Arg [1]    : Arg [1] : (optional) String - the biotype to set
+  Example    : my $biotype = $gene->biotype;
+               my $biotype = $gene->biotype('protin_coding');
+  Description: Returns the Biotype object of this gene.
+               When no biotype exists, defaults to 'protein_coding'.
+  Returntype : Bio::EnsEMBL::Biotype
+  Exceptions : none
+  Caller     : general
+  Status     : Stable
```
nerdstrike: As discussed, status documentation is not meaningful as we have no policy to keep them up to date. Remove

modules/Bio/EnsEMBL/Gene.pm
```diff
@@ -1527,6 +1507,42 @@ sub havana_gene {
...
+    $self->{'biotype'} = 'protein_coding';
+  }
+
+  my $biotype;
+  if( defined $self->adaptor() ) {
+    my $ba = $self->adaptor()->db()->get_BiotypeAdaptor();
+    $biotype = $ba->fetch_by_name_object_type( $self->{'biotype'}, 'gene' );
+  }
+
+  return ( $biotype || $self->{'biotype'} );
```
nerdstrike: This method should short-circuit to the data in the object, rather than querying the database whether it has a biotype property or not. It's a small but valuable efficiency tweak for users.

modules/Bio/EnsEMBL/Transcript.pm
```diff
@@ -3229,6 +3213,42 @@ sub get_Gene {
...
-1;
+  Arg [1]    : Arg [1] : (optional) String - the biotype to set
+  Example    : my $biotype = $transcript->biotype;
+               my $biotype = $transcript->biotype('protin_coding');
+  Description: Returns the Biotype object of this transcript.
+               When no biotype exists, defaults to 'protein_coding'.
+  Returntype : Bio::EnsEMBL::Biotype
+  Exceptions : none
+  Caller     : general
+  Status     : Stable
```
nerdstrike: As before, remove status unless you plan to keep it up to date forever.

modules/Bio/EnsEMBL/Transcript.pm
```diff
@@ -3229,6 +3213,42 @@ sub get_Gene {
...
+    $self->{'biotype'} = 'protein_coding';
+  }
+
+  my $biotype;
+  if( defined $self->adaptor() ) {
+    my $ba = $self->adaptor()->db()->get_BiotypeAdaptor();
+    $biotype = $ba->fetch_by_name_object_type( $self->{'biotype'}, 'transcript' );
+  }
+
+  return ( $biotype || $self->{'biotype'} );
```
nerdstrike: Similarly here, return the cached value before trying to consult the database. Caching is valuable when you're 500 ms from a database server.

modules/t/biotype.t
```diff
@@ -0,0 +1,79 @@
...
+# Get a DBAdaptor to from the test system
+my $multi = Bio::EnsEMBL::Test::MultiTestDB->new;
+ok($multi);
+my $db = $multi->get_DBAdaptor("core");
+ok($db);
+
+# Should get meaningful type back
+debug("get biotype adaptor");
+my $biotype_adaptor = $db->get_BiotypeAdaptor();
+ok($biotype_adaptor->isa("Bio::EnsEMBL::DBSQL::BiotypeAdaptor"));
```
nerdstrike: Annotate your ok( ) tests with a message to aid debugging failures.

modules/t/biotype.t
```diff
@@ -0,0 +1,79 @@
...
+
+# fetch a protein_coding gene object
+debug("fetch gene");
+my $ga = $db->get_GeneAdaptor();
+my $gene = $ga->fetch_by_stable_id("ENSG00000171456");
+ok($gene);
+
+# test biotype related methods on this gene object
+debug("gene biotype");
+is($gene->{'biotype'}, 'protein_coding');
```
nerdstrike: Don't know why you'd test the value of an internal structure when you have a perfectly good getter. That is unless you're testing defaults and haven't mentioned it.

modules/t/biotype.t
```diff
@@ -0,0 +1,79 @@
...
+debug("transcript biotype");
+is($transcript->{'biotype'}, 'protein_coding');
+is($transcript->biotype, 'protein_coding');
+$biotype = $transcript->biotype;
+ok($biotype->isa("Bio::EnsEMBL::Biotype"));
+is($biotype->object_type, 'transcript', 'Biotype is from Transcript object');
+is($biotype->name, 'protein_coding', 'Biotype name is protein_coding');
+is($biotype->biotype_group, 'coding', 'Biotype group is coding');
+is($biotype->so_acc, 'SO:0000234', 'Biotype protein_coding refers to SO:0000234');
+
```
nerdstrike: I'd like to see a non-protein-coding group tested too.

What happens if you change a biotype? Do you then get the new grouping?

modules/t/test-genome-DBs/test_collection/core/SQLite/table.sql
```diff
@@ -131,6 +131,22 @@ CREATE TABLE attrib_type (
 
 CREATE UNIQUE INDEX code_idx ON attrib_type (code);
 
+--
+-- Table: biotype
+--
+CREATE TABLE biotype (
+  biotype_id INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL,
```
nerdstrike: I believe a primary key must be not null by definition.

modules/t/biotype.t
```diff
@@ -0,0 +1,79 @@
...
+# test fetch biotypes of object_type gene
+my $biotypes = $biotype_adaptor->fetch_all_by_object_type('gene');
+is(ref $biotypes, 'ARRAY', 'got an array');
+is(scalar @{$biotypes}, '2', 'of size 2');
+
```
nerdstrike: What happens if you store a Biotype instance? What happens if you store one with missing unique key values?

modules/Bio/EnsEMBL/Gene.pm
tgrego: ok, I'll put it back for now...
might be a good idea to revisit this and clean the code a bit though, I don't think it's useful.

modules/Bio/EnsEMBL/DBSQL/BiotypeAdaptor.pm
tgrego: So it's looks this is required which means I can't raise an exception...
Do you think a warning would be a good compromise?

modules/Bio/EnsEMBL/DBSQL/GeneAdaptor.pm
tgrego: I know, promise from now on I'll do the whitespace removal in separate commits.
But admit it, looks sooo much better (even if you can't see it because it's whitespace!).

modules/Bio/EnsEMBL/Gene.pm
tgrego: missed a couple of them, removed.

modules/t/test-genome-DBs/test_collection/core/SQLite/table.sql
tgrego: this is actually auto generated by the patch_test_databases.pl script...

modules/Bio/EnsEMBL/Gene.pm
tgrego: there is indeed the chance that a string is returned instead of a Biotype object.
That is when adaptor() does not exist in the gene or transcript object.
I'll change it to always return a biotype object, but will need to
use Bio::EnsEMBL::Biotype in the Gene and Trancript classes (to create a Biotype object there bypassing the BiotypeAdaptor).

modules/Bio/EnsEMBL/DBSQL/BiotypeAdaptor.pm
nerdstrike: I suspect warnings would upset genebuild. Document the need instead?

modules/Bio/EnsEMBL/DBSQL/BiotypeAdaptor.pm
tgrego: Added a warning if no rows retrieved.
Empty array returned.

nerdstrike: I think you've got it covered. Would you mind putting in a test to prove that it fails gracefully? A proper exception is better than a warning which may go unnoticed. Does this code need to fail quietly?

tgrego: Added a test to check the warning and return value ([] in case of warning).
I suppose the warning can be an exception instead... Either a full replacement (if empty return value) or just like the exception in Bio::EnsEMBL::Biotype::object_type()
What do you think would be better?

nerdstrike: I prefer my code to fail loudly and early, as that greatly speeds debugging, but hard failures upset downstream users who have relied on the silent failures. It's your call.

modules/Bio/EnsEMBL/Gene.pm
tgrego: Have modified this for efficiency.
if biotype() is never called, the biotype table is never queried.
When biotype() is first called, the Biotype object is retrieved and stored for future biotype() calls.

nerdstrike:
I think you've dealt with all my concerns.

nerdstrike:
@avullo Would you mind giving Tiago's code one more go over?

magaliruffier:
the schema version changes are now on master, can you please remove any changes on your branch that are related to the release version increment?

tgrego:
that should be fixed with the merge of master into the feature branch, let me know if conflicts still exist.

PR #229, opened 2018-02-26
[ENSCORESW-2465] Fix comparisons in Exon::is_coding() to prevent fals…

ens-bwalts:
…e-negatives when transcript and exon boundaries match exactly. Added additional transcript and translation structures to exon test cases.

## Requirements

## Description

Changes comparisons in Exon::is_coding() to use <= and >= so that they catch cases where coding region boundaries exactly match exon boundaries. Also added several new test cases to exon.t to check for correct behaviour when exon boundaries exactly match coding region boundaries in both single- and multi-exon transcripts.

## Use case

When exon boundaries exactly match coding region boundaries, Exon::is_coding() was returning zero, even though the exons were coding (false negative).

## Benefits

Fixes a user-reported bug, and increases test coverage to identify if this (or a similar) bug reoccurs.

## Possible Drawbacks

None

## Testing

Yes, in exon.t

Pass. Checked that they fail with previous version of is_coding()

Yes.

### DIFF
--- a/modules/Bio/EnsEMBL/Exon.pm
+++ b/modules/Bio/EnsEMBL/Exon.pm
@@ -1302,11 +1302,15 @@ sub is_coding {
   my ( $self, $transcript) = @_;
 
   if (!$transcript->translate) { return 0; }
-  if ($transcript->coding_region_start < $self->start && $self->start < $transcript->coding_region_end) { return 1; }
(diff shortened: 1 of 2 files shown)


nerdstrike:
Looks like a good fix to have.

PR #231, opened 2018-03-15
Test for IndelCoordinate first because IndelCoordinate can be a Gap

at7:
This fixes a problem when drawing strain slice sequence against reference sequence in the Resequencing view. Sometimes a IndelCoordinate was treated only as a Gap because Gap is the base class.

### DIFF
--- a/modules/Bio/EnsEMBL/MappedSlice.pm
+++ b/modules/Bio/EnsEMBL/MappedSlice.pm
@@ -574,20 +574,14 @@ sub seq {
       
       # indel / gap
       else {
-        
-        if($ref_coord->isa('Bio::EnsEMBL::Mapper::Gap')) {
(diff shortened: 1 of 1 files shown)


magaliruffier:
Looks reasonably sane. There are no tests at all for MappedSlice at the moment, is this the module you are planning to retire next release?

at7:
I searched the web-code and compara code for use of the module. It looks like only variation uses it. Therefore, I'm happy to move it into the variation API for release/93 and deprecate it in the core API for 93. Could you do me a favour and cherry pick it into release/92 please? It is a postrelease fix and it would be great to see the resequencing view working again.

magaliruffier:
Changes are fine with me. Can you please submit a separate pull request against release/92 to add it there? I'll leave it to a second person of Core to merge the PR.

PR #233, opened 2018-03-16
Feature/biotype update to keep legacy biotype()

tgrego:
## Requirements

## Description

Recent addition of Biotype objects used the old biotype method in gene and transcript.
This proven not to be 100% backwards compatible, so this is to include legacy biotype() dealing with the biotype name string, together with new get_Biotype and set_Biotype methods that deal with biotype objects.

## Use case

`to_json($gene->biotype)` would refuse to do it's job because $gene->biotype was returning an object, even though that object would behave as a string if to_json would make the effort to try it...
Now we bring $gene->biotype back to being a real string, and provide new methods to deal with objects.

## Benefits

Backwards compatibility

## Possible Drawbacks

Might have missed something

## Testing

Yes.

Pass, of course.

Yes.


nerdstrike:
Looks ok, are you expecting to make more changes?

tgrego:
yes, I pushed just now a required schema_version check.

magaliruffier:
I can confirm that the backwards compatibility tests I've executed are now working

lairdm:
Did I miss it, is there a test for this fallback functionality?

magaliruffier:
The test I ran locally is to call biotype on a database that is on schema 92 or older. Not sure how this could be integrated in the test suite, as we typically assume schema and API version are in sync

lairdm:
Ahh, I see, $gene->biotype is testing in the unchanged sections still, that's what I was getting at. Alright, good to go.

PR #234, opened 2018-03-23
Remove variation code

at7:
First part of removing deprecated variation code
## Description
Remove deprecated code StrainSlice and StrainSliceAdaptor.
Update StrainSlice occurrences in core API to use Variation::StrainSlice

Follow up from this PR:
Deprecate AlignStrainSlice and IndividualSlice
Possibly move MappedSliceContainer and MappedSlice to Variation API?


at7:
I will also add tests for MappedSlice and MappedSliceContainer

at7:
I will update test cases and then create a new PR.

PR #235, opened 2018-03-26
fixed cpanfile syntax

warrenson:
## Description
Correct perl syntax required for `cpanfile`. The `;` chars are missing at the end of the lines.

## Use case

Installing the perl requisites requires the `cpanfile` to have correct perl syntax.

## Testing

No

NA

No


avullo:
Thanks, good point. Can you please do the same for master?

PR #236, opened 2018-03-29
Cleanup strain slice

at7:
First part of removing deprecated variation code. Remove StrainSlice related code where possible.

### Description
Remove deprecated code StrainSlice and StrainSliceAdaptor.
Update StrainSlice occurrences in core API to use Variation::StrainSlice
Add tests for remaining use of StrainSlice.

Follow up from this PR:
Deprecate AlignStrainSlice and IndividualSlice
Possibly move MappedSliceContainer and MappedSlice to Variation API?


modules/Bio/EnsEMBL/Slice.pm
```diff
@@ -2461,33 +2460,6 @@ sub get_by_Individual{
 
 }
 
-=head2 get_by_strain
```
avullo: So this should have been removed quite well before.

avullo:
So far, so good to me, much needed move towards decoupling the respective API responsibilities.
Coverage complain is not to be taken seriously in this case. I'm gonna merge after someone else has taken a look.

Thanks

modules/Bio/EnsEMBL/Slice.pm
at7: Yes. Everything Individual related can go. I haven't found any dependencies in the web code. I will deprecate the Individual code in a separate PR.

avullo: That makes sense, thanks.

avullo:
Thanks Anja,, waiting for the "Individual" PR too :+1:

modules/Bio/EnsEMBL/AlignStrainSlice.pm
```diff
@@ -302,7 +302,7 @@ sub get_all_Slices {
     }
   my $indAdaptor = $dbVar->get_IndividualAdaptor();
   my $ref_name =  $indAdaptor->get_reference_strain_name;
-  my $ref_strain = Bio::EnsEMBL::StrainSlice->new(
+  my $ref_strain = Bio::EnsEMBL::Variation::StrainSlice->new(
```
avullo: Here we forgot to import the corresponding module I suspect

PR #238, opened 2018-04-05
Remove var dependency

at7:
Remove StrainSlice dependency from Feature and add feature_Slice call to variation API where it is required. Remove test which was testing that feature_Slice returns a StrainSlice.

### DIFF
--- a/modules/t/strainSlice.t
+++ b/modules/t/strainSlice.t
@@ -43,15 +43,6 @@ my $strain_slice = Bio::EnsEMBL::Variation::StrainSlice->new(
 
 my $vf_adaptor = $vdba->get_VariationFeatureAdaptor;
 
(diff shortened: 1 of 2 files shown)


at7:
Yes. The change is fine for the variation API.

modules/t/strainSlice.t
```diff
@@ -43,15 +43,6 @@ my $strain_slice = Bio::EnsEMBL::Variation::StrainSlice->new(
 
 my $vf_adaptor = $vdba->get_VariationFeatureAdaptor;
```
tgrego: This can be removed as it is not used anymore.

at7: It is used further down line 83.

tgrego: indeed, my bad.

PR #244, opened 2018-06-01
Don't version_check metadata db

james-monkeyshines:
## Description
The metadata database is unversioned, and exists apart from the release cycle. So the version check is not required.

## Use case
Loading the metadata db in the registry gives warnings about a missing db version.

## Benefits
Incorrect warnings are removed.

## Possible Drawbacks
None

## Testing
No new test, it's a minor change. registry.t passes.

### DIFF
--- a/modules/Bio/EnsEMBL/Registry.pm
+++ b/modules/Bio/EnsEMBL/Registry.pm
@@ -2875,6 +2875,9 @@ sub version_check {
     if ( $dba->dbc()->dbname() =~ /^_test_db_/x ) {
       return 1;
     }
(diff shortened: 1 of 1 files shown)


lairdm:
Makes sense.

lairdm:
Fair comment.

james-monkeyshines:
I think all metadata-related things would be inherently unversioned. The most likely case of a variant of the ensembl_metadata db is a WormBase or VectorBase copy, which you might reasonably call something like ensembl_metadata_vb. We wouldn't want a version check on such a db either, which would mean an anchor in the regex would not be desirable.

PR #246, opened 2018-06-06
SequenceOntologyMapper deprecation

tgrego:
## Requirements

## Description

This PR deprecates the SequenceOntologyMapper, providing the missing alternative ways to have all the functionality it provides.
To get the SO acc of a feature:
`$feature->feature_so_acc`
To get the SO acc of a gene/transcript biotype:
`$feature->get_Biotype->so_acc`
To get the SO name of a SO acc:
`$oa = Bio::EnsEMBL::Registry->get_adaptor('multi','ontology','OntologyTerm');`
`$so_name = $oa->fetch_by_accession($so_acc)->name;`

If you want biotype SOs and fallback to feature SOs, implement that logic in your code.

## Use case

SequenceOntologyMapper provides a way to access SO terms for biotypes and feature types, through a series of lookup tables.
The Gene and Transcript biotype SO terms could already be accessed alternatively using the Biotype object that can be accessed by the get_Biotype method on gene and transcript objects.
Feature type SO terms were missing and are now implemented, through the new method feature_so_acc that is inherited and can be overridden by features. 

## Benefits

Removing the SequenceOntologyMapper

## Possible Drawbacks

Although a 'feature' thing, Slice implements feature_so_acc. Slice is however not a feature, but will return either the SO of its coordinate system if available or of a generic region.

Bio::EnsEMBL::Compara::ConstrainedElement was included in the SequenceOntologyMapper. It is however not a feature, and thus will need implementation of feature_so_acc. Attention, this is in ensembl-compara.

GFFSerializer in this repo used the SequenceOntologyMapper and was updated to use the alternative ways, removing the SequenceOntologyMapper requirement.
Other modules in ensembl-production and ensembl-io (and maybe more?) make use of SequenceOntologyMapper and require similar updates.

## Testing

Yes.

Pass, of course.

Yes.

### DIFF
--- a/modules/Bio/EnsEMBL/CDS.pm
+++ b/modules/Bio/EnsEMBL/CDS.pm
@@ -66,6 +66,7 @@ use Scalar::Util qw(weaken isweak);
 
 @ISA = qw(Bio::EnsEMBL::Feature);
 
+use constant SO_ACC => 'SO:0000316';
 
 =head2 new
 
@@ -238,21 +239,6 @@ sub phase {
   my $self = shift;
   $self->{'phase'} = shift if( @_ );
   return ( $self->{'phase'} || 0 );
-}
-
-
-=head2 feature_so_acc
-
-  Example    : print $cds->feature_so_acc;
-  Description: This method returns a string containing the SO accession number of CDS.
-               Overrides Bio::EnsEMBL::Feature::feature_so_acc
-  Returntype : string (Sequence Ontology accession number)
-
-=cut
-
-sub feature_so_acc {
-
-  return 'SO:0000316';
 }
 
 
--- a/modules/Bio/EnsEMBL/CDS.pm
+++ b/modules/Bio/EnsEMBL/CDS.pm
@@ -241,6 +241,21 @@ sub phase {
 }
 
 
+=head2 feature_so_acc
+
+  Example    : print $cds->feature_so_acc;
+  Description: This method returns a string containing the SO accession number of CDS.
+               Overrides Bio::EnsEMBL::Feature::feature_so_acc
+  Returntype : string (Sequence Ontology accession number)
+
+=cut
+
+sub feature_so_acc {
+
+  return 'SO:0000316';
+}
+
+
 =head2 summary_as_hash
 
   Example    : my $hash = $cds->summary_as_hash();
--- a/modules/Bio/EnsEMBL/Feature.pm
+++ b/modules/Bio/EnsEMBL/Feature.pm
@@ -1449,6 +1449,7 @@ sub get_nearest_Gene {
 
   Description: This method returns a string containing the SO accession number of the feature
   Returntype : String (Sequence Ontology accession number)
+  Exceptions : Thrown if no SO accession can be returned for caller object type
 
 =cut
 
@@ -1480,7 +1481,11 @@ sub feature_so_acc {
     'Bio::EnsEMBL::KaryotypeBand'                         => 'SO:0000341', # chromosome_band
   );
 
-  return $feature_so_mapping{ ref $self };
+  my $ref = ref $self;
+
+  return $feature_so_mapping{ $ref } //
+    throw( "Bio::EnsEMBL::Feature::feature_so_acc is not defined for $ref objects. " .
+      "Please update Bio::EnsEMBL::Feature::feature_so_acc or implement ${ref}::feature_so_acc");
 }
 
 =head2 summary_as_hash
--- a/modules/Bio/EnsEMBL/Feature.pm
+++ b/modules/Bio/EnsEMBL/Feature.pm
@@ -82,6 +82,7 @@ use Bio::EnsEMBL::Utils::Argument qw(rearrange);
 use Bio::EnsEMBL::Utils::Exception qw(throw warning);
 use Bio::EnsEMBL::Utils::Scalar qw(check_ref assert_ref);
 use Bio::EnsEMBL::Slice;
+use Try::Tiny;
 use vars qw(@ISA);
 
 use Scalar::Util qw(weaken);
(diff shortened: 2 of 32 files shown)


modules/Bio/EnsEMBL/Slice.pm
```diff
@@ -3759,6 +3759,27 @@ sub add_synonym{
...
+
+  my %region_so_mapping = (
+   'chromosome'  => 'SO:0000340',
+   'supercontig' => 'SO:0000148',
+   'scaffold'    => 'SO:0000148',
+   'contig'      => 'SO:0000149'
+  );
+
+  # return the region SO acc, or Slice acc
+  return %region_so_mapping{$self->coord_system_name} // 'SO:0000001';
```
magaliruffier: should that not be $region_so_mapping{$self->coord_system_name}?
Older versions of Perl do not seem happy with that syntax

modules/Bio/EnsEMBL/UTR.pm
```diff
@@ -220,6 +220,26 @@ sub type {
...
+sub feature_so_acc {
+  my $self = shift;
+
+  my %utr_type_so_mapping = (
+   'five_prime_utr'  => 'SO:0000204',
+   'three_prime_utr' => 'SO:0000205'
+  );
+
+  # return UTR type SO acc, or UTR acc
+  return %utr_type_so_mapping{$self->type} // 'SO:0000203';
```
magaliruffier: review %hash{$elt} syntax

modules/Bio/EnsEMBL/Utils/SequenceOntologyMapper.pm
```diff
@@ -314,6 +315,9 @@ my %feature_so_mapping =
 
 sub new {
   my ($class, $oa) = @_;
+
+  deprecate("new is deprecated and will be removed in e98.");
```
magaliruffier: deprecation warnings should be added to the DEPRECATED.md file

modules/Bio/EnsEMBL/Feature.pm
```diff
@@ -1445,6 +1445,49 @@ sub get_nearest_Gene {
...
+  Exceptions : Thrown if no SO accession can be returned for caller object type
+
+=cut
+
+sub feature_so_acc {
+  my ($self) = @_;
+
+  # Classes that inherit Feature will inherit this method
+  # Slice and UTR override this method for better SO accs based on type
+  my %feature_so_mapping = (
```
nerdstrike: All static lists should be class properties, not dynamically defined. It's faster and saves memory reallocation.

I would have preferred to have this list of SO terms dispersed into the relevant classes as package level "constants" so that each team may maintain their own SO term assignments, but I expect the support path leads back to us either way if a pipeline breaks, so it perhaps does not matter. The exceptional behaviour is still needed for non-Features, but it would have been nice to have a more OO-approach.

modules/Bio/EnsEMBL/Slice.pm
```diff
@@ -3759,6 +3759,27 @@ sub add_synonym{
...
+
+  Example     : $slice_so_acc = $slice->feature_so_acc;
+  Description : This method returns a string containing the SO accession number of the slice, based on the coordinate system name.
+  Returns     : string (Sequence Ontology accession number)
+=cut
+
+sub feature_so_acc {
+  my $self = shift;
+
+  my %region_so_mapping = (
```
nerdstrike: As above, take this outside of the sub definition, so that it is not defined for every invocation.

modules/Bio/EnsEMBL/Feature.pm
```diff
@@ -1445,6 +1445,49 @@ sub get_nearest_Gene {
...
+    'Bio::EnsEMBL::Funcgen::RegulatoryFeature'            => 'SO:0005836', # regulatory_region
+    'Bio::EnsEMBL::DnaDnaAlignFeature'                    => 'SO:0000347', # nucleotide_match
+    'Bio::EnsEMBL::DnaPepAlignFeature'                    => 'SO:0000349', # protein_match
+    'Bio::EnsEMBL::KaryotypeBand'                         => 'SO:0000341', # chromosome_band
+  );
+
+  my $ref = ref $self;
+
+  return $feature_so_mapping{ $ref } //
+    throw( "Bio::EnsEMBL::Feature::feature_so_acc is not defined for $ref objects. " .
```
nerdstrike: I'm pleased to see a decent error message. Is there a way to test this exception? I realise it requires a class that has no entry, but you could bless to something random.

modules/Bio/EnsEMBL/Feature.pm
tgrego: Have ended up refactoring this feature to explicit feature_so_acc methods in the actual classes. If required and not overridden, then will throw an error.
Also took the static lists to class properties where applicable.

modules/Bio/EnsEMBL/Feature.pm
tgrego: it's now being tested in Operon and OperonTranscript features.

modules/Bio/EnsEMBL/Feature.pm
tgrego: When feature has the ok will submit PR to Compara, FuncGen and Variation with the missing class updates.
    'Bio::EnsEMBL::Variation::VariationFeature'  => 'SO:0001060', # sequence variant
    'Bio::EnsEMBL::Variation::StructuralVariationFeature' => 'SO:0001537', # structural variant
    'Bio::EnsEMBL::Compara::ConstrainedElement' => 'SO:0001009', # DNA_constraint_sequence ????
    'Bio::EnsEMBL::Funcgen::RegulatoryFeature' => 'SO:0005836', # regulatory_region

modules/Bio/EnsEMBL/CDS.pm
```diff
@@ -241,6 +241,21 @@ sub phase {
...
+=head2 feature_so_acc
+
+  Example    : print $cds->feature_so_acc;
+  Description: This method returns a string containing the SO accession number of CDS.
+               Overrides Bio::EnsEMBL::Feature::feature_so_acc
+  Returntype : string (Sequence Ontology accession number)
+
+=cut
+
+sub feature_so_acc {
```
nerdstrike: This is a lot of copy-paste. 

I would have gone for:
$self->{so_term} = 'SO:0000316';

and in Bio::EnsEMBL::Feature
```
sub ->feature_so_acc {
  my $self = shift;
  if ( ### one of the exceptional classes ) {
    ...
  } else {
    return $self->{so_term};
  }
}
```
Maybe with some handling of undef etc.

PR #247, opened 2018-06-12
Bugfix/reactome acc

magaliruffier:
## Requirements

## Description

The description field is compared to a regex and if the description does not match, it is skipped

## Use case

Some Reactome descriptions have characters that do not comply with MySQL ASCII. This regex will skip these.
An example of skipped description is 'Biosynthesis of electrophilic ω-3 PUFA oxo-derivatives', because of the greek omega (ω)

## Benefits

The pipeline does not fail for everything because some characters in the Reactome source are not compatible with MySQL

## Possible Drawbacks

The regex is not comprehensive, so it can skip some descriptions we have not yet accounted for.

## Testing

The xref pipeline does not have any tests, so the test suite has not been changed. The parser has been tested on the latest Reactome file though. 1648 entries were skipped out of the 1077438 lines in the file. These entries correspond to two distinct descriptions, 'Biosynthesis of electrophilic ω-3 PUFA oxo-derivatives' and 'Loss of proteins required for interphase microtubule organization from the centrosome'

### DIFF
--- a/misc-scripts/xref_mapping/XrefParser/ReactomeParser.pm
+++ b/misc-scripts/xref_mapping/XrefParser/ReactomeParser.pm
@@ -113,6 +113,7 @@ sub run {
     while (my $line = $reactome_io->getline() ) {
       chomp $line;
       my ($ensembl_stable_id, $reactome_id, $url, $description, $evidence, $species) = split /\t+/,$line;
+      if ($description!~ /^[A-Za-z0-9_,\(\)\/\-\.:\+'&;>\s]+$/) { next; }
   
       $species =~ s/\s//;
       $species = lc($species);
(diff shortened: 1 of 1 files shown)


lairdm:
Ugly but effective, hopefully doesn't accidentally catch any other strings that shouldn't be excluded.

nerdstrike:
I can conceive of a few symbols appearing that are not in the approved set: 
" ' ? / % [ ] 
It's up for debate whether we need them.

magaliruffier:
I am not averse to allowing more symbols.
I tested it on the current Reactome file, so I know it allows most of the descriptions we currently get, but it is not future proof.

lairdm:
So do we want to add more symbols, or shall we wait and see if there's unexpected breakage? Your pull request, your choice.

PR #249, opened 2018-06-13
Bugfix/default aln method

magaliruffier:
## Requirements

## Description

Ensure there is a default align_type for protein_feature and it is ensembl

## Use case

The align_type column in the protein_feature table is an enum. Existing scripts and pipeline that do not use this column should not have to worry about assigning it, hence it is useful to have a default value that makes sense. This is set to 'ensembl' by default. Other

## Benefits

When new entries are added, the align_type will not be set to an empty string.

## Possible Drawbacks

This will be updating the protein_feature table in all existing core databases, so this might take a while to run

## Testing

No additional tests required.

All tests pass, in particular schemaPatches, that ensures the patch file is in sync with table.sql

Yes.

### DIFF
--- a/sql/patch_93_94_b.sql
+++ b/sql/patch_93_94_b.sql
@@ -25,4 +25,4 @@ UPDATE object_xref SET analysis_id = NULL WHERE analysis_id = 0;
 
 # patch identifier
 INSERT INTO meta (species_id, meta_key, meta_value)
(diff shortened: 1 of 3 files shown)


lairdm:
Isn't this a violation of the two separate changes in the same pull request guideline? Fixing a previous patch and adding a new change.

PR #252, opened 2018-06-19
Fix meta entry in patch_93_94_b.sql

mkszuba:
## Description

patch_93_94_b.sql incorrectly identified itself as _c in its entry in the meta table. Fixed.

## Use case

The old behaviour introduced an inconsistency and could perhaps cause problems with automatic patching of databases.

## Benefits

Patch _b is now correctly logged as _b in the meta table.

## Possible Drawbacks

None I can think of.

### DIFF
--- a/sql/patch_93_94_b.sql
+++ b/sql/patch_93_94_b.sql
@@ -25,4 +25,4 @@ UPDATE object_xref SET analysis_id = NULL WHERE analysis_id = 0;
 
 # patch identifier
 INSERT INTO meta (species_id, meta_key, meta_value)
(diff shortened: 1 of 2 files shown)


magaliruffier:
This change has also been fixed in this PR: https://github.com/Ensembl/ensembl/pull/249/commits/d29b8d6633b9e88f704a849b91cdb1c0773d4b59
Sorry for not submitting it as a separate change

magaliruffier:
the patch name also needs to be updated in the test databases, see https://github.com/Ensembl/ensembl/pull/255

PR #253, opened 2018-06-19
Pre release 94 cleanup

at7:
## Requirements
Functionality which is removed from core API is stable and contained in variation API.

## Description
This is part of our effort to remove variation dependencies from the core API.
I removed all deprecated code that is not used in the web code and compara API.
I cannot remove Bio::EnsEMBL::Slice::get_all_VariationFeatures() because in the web code this is called on modules/Bio/EnsEMBL/Compara/AlignSlice/Slice.pm. Not sure if there is a way around this. However, get_all_somatic_VariationFeatures, get_all_somatic_VariationFeatures_by_source, get_all_somatic_VariationFeatures_with_phenotype and get_all_VariationFeatures_by_Population are called on proper core Slice objects in the web code. I will replace those for release/95 which gives me enough time for testing. get_by_strain will be removed in release/95 together with code that still uses get_by_strain at the moment.

## Use case
Code allows to get variation data on slice objects.

## Benefits
remove variation code from core API and therefore reduce possible ways of getting variation data for a slice. Remove code duplications.

## Possible Drawbacks
Possible that I have overseen dependencies which could result in broken web views. But enough testing should prevent that.

## Testing
Lot's of manual testing of web views which use the code.


lairdm:
The tests all seem to still be working, and the functions that aren't being removed until 95 are still passing. I have mild concerns with mixing the removal of the functions with the change in the depreciation warning for different functions. But it's all related in eventually removing this entire sub-section of the repository, so it shouldn't case too much confusion.

PR #254, opened 2018-06-21
ENSCORESW-2742: clean up descriptions from automatic updates

magaliruffier:
## Requirements

## Description

Descriptions assigned by the automatic xref update have a 'Source' field and are removed before adding the newly assigned descriptions.

## Use case

For merged species, there will be a mix of descriptions assigned by the annotators and descriptions assigned by the automatic xref pipeline. While xref updates can be run every release, the annotation is not updated every release. When running xref updates on an annotation that has not changed since the last release, the database will contain two types of descriptions: the manually assigned ones to be conserved and the automatically assigned ones that might become obsolete through the xref update. By deleting the descriptions of type '%Source%', we ensure the latter ones are deleted while the former are conserved.

## Benefits

We do not keep descriptions which are obsolete due to updated mappings.

## Possible Drawbacks

Some genes will lose all description if their xref mapping is lost.

## Testing

The xref pipeline has been tested with the change and the results are identical to the previous run.

NA

NA

### DIFF
--- a/misc-scripts/xref_mapping/XrefMapper/DisplayXrefs.pm
+++ b/misc-scripts/xref_mapping/XrefMapper/DisplayXrefs.pm
@@ -234,7 +234,7 @@ sub set_display_xrefs_from_stable_table{
 
   # Remove descriptions assigned through the xref pipeline, recognisable by the 'Source' field
   # This will maintain any manually added descriptions
-  $reset_sth = $core_dbi->prepare("UPDATE gene SET description = null WHERE description like '%Source%'");
+  $reset_sth = $core_dbi->prepare("UPDATE gene SET description = null WHERE description like '%[Source:%]%'");
   $reset_sth->execute();
   $reset_sth->finish;
 


misc-scripts/xref_mapping/XrefMapper/DisplayXrefs.pm
```diff
@@ -227,7 +227,14 @@ sub set_display_xrefs_from_stable_table{
...
  
-  $reset_sth = $core_dbi->prepare("UPDATE transcript SET display_xref_id = null WHERE biotype NOT IN ('LRG_gene')");
+  # Remove any leftover transcript description, as it is not used anywhere
+  $reset_sth = $core_dbi->prepare("UPDATE transcript SET display_xref_id = null");
+  $reset_sth->execute();
+  $reset_sth->finish;
+
+  # Remove descriptions assigned through the xref pipeline, recognisable by the 'Source' field
+  # This will maintain any manually added descriptions
+  $reset_sth = $core_dbi->prepare("UPDATE gene SET description = null WHERE description like '%Source%'");
```
ens-bwalts: Consider matching '%[Source:%]%' for safety. Granted, in 92 all the human gene descriptions containing the string 'source' appear to be from automated updates. (same count of descriptions containing '%source%' and descriptions containing '%[Source:%]%'). However, it feels a little safer to use a stricter match.

PR #259, opened 2018-07-03
ENSCORESW-2740: fetch correct xref

magaliruffier:
## Requirements

## Description

Allows to re-use duplicated entries with null descriptions.

## Use case

When assigning display_xrefs, it is possible to re-use the same display_xref for the transcript if the gene name is assigned to multiple genes. In this case, it is not possible to store the same entry twice and the existing entry should be re-used. As the description field can be null, it cannot be included in the query, otherwise the existing entry is not found and a new one is attempted to store but fails.

## Benefits

The DisplayXref HC is not failing any more and transcripts get correctly assigned a display_xref.

## Possible Drawbacks

If we have two xrefs with the same display_label and accession but different descriptions, we will not store the description for the second one. Fundamentally, the same display_label should not have two different descriptions.

## Testing

The xref pipeline was run with the SQL query updated on a species which failed DisplayXref in release 93. Running the pipeline on the exact same database, including the change means the HC does not fail.

The pipeline and HCs pass, the test suite does not include xref code.

NA


avullo:
Looks good to go for me, thanks.

PR #262, opened 2018-07-09
Do not call fetchrow_hashref() if the statement has been exhausted

muffato:
## Description

`has_next` in `Utils::Iterator` calls `next`, and expects `next` to repeatedly return _undef_ once the iterator has been exhausted.

`Utils::SqlHelper` can return an iterator based on the sub below, but the sub calls `fetchrow_hashref()` every time it is called, even when the iterator has already reached the end.

So if you build an iterator with SqlHelper, exhaust it, and then call `has_next`, you'll get this error
```
DBD::mysql::st fetchrow_arrayref failed: fetch() without execute() at ensembl/modules/Bio/EnsEMBL/Utils/SqlHelper.pm line 961.
```

My solution is to check whether the handle is still _Active_ as it is set to false when all the rows have been retrieved (or `$sth->finish`) has been called, cf https://metacpan.org/pod/DBI#Active

## Testing

Yes.

Pass, of course.

Yes.

### DIFF
--- a/modules/Bio/EnsEMBL/Utils/SqlHelper.pm
+++ b/modules/Bio/EnsEMBL/Utils/SqlHelper.pm
@@ -948,6 +948,7 @@ sub _execute {
   my $sth_processor;
   if($use_hashrefs) {
     $sth_processor = sub {
+      return unless $sth->{Active};
       while( my $row = $sth->fetchrow_hashref() ) {
         my $v = $callback->($row, $sth);
         return $v if $has_return;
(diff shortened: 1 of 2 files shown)


muffato:
PS: Would you consider having the fix on e94 too ?

nerdstrike:
Wow. That's a gaping hole in iterator behaviour. Thanks for finding that @muffato .

nerdstrike:
@muffato - I will gladly accept another pull request against the e94 branch

PR #263, opened 2018-07-10
Do not call fetchrow_hashref() if the statement has been exhausted

muffato:
Cherry-picked from master. See #262


muffato:
Thank you @nerdstrike !

PR #267, opened 2018-07-24
ENSCORESW-2810: pass RFAMParser even if no xrefs found

magaliruffier:
## Requirements

## Description

The proposed change prevents the RFAMParser from failing if no xrefs are created.

## Use case

The RFAMParser relies on finding transcripts which have RFAM accessions as supporting features. This happens when the ncRNA pipeline has been run for the species in question. Not all species will have that data, but we still want to run the xref pipeline for those species without failing on the RFAMParser. Two species that triggered this change are mus spretus and mus castanus.

## Benefits

The xref pipeline does not fail if no RFAM annotation is available for a species.

## Possible Drawbacks

Most species should have RFAM annotation and we will not be picking this up with the xref pipeline. However, this should probably be spotted by other mechanisms than the xref update.

## Testing

The parser has been run on several species before and after the change. Before the change, species without RFAM annotation fail the parser. With the fix, the parser passes for all species, with or without RFAM annotation.
This is also noted in ENSCORESW-2810

### DIFF
--- a/misc-scripts/xref_mapping/XrefParser/RFAMParser.pm
+++ b/misc-scripts/xref_mapping/XrefParser/RFAMParser.pm
@@ -185,9 +185,6 @@ sub run_script {
   }
 
   print "Added $xref_count RFAM xrefs and $direct_count direct xrefs\n" if($verbose);
-  if ( !$xref_count ) {
-      return 1;    # 1 error
-  }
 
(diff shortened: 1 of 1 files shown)


magaliruffier:
Fair point @nerdstrike I'll add a ticket for that

PR #270, opened 2018-07-25
eHive DBConnections don't allow hash access on the StatementHandle

muffato:
## Description

eHive uses different encapsulation of `DBI::st` (containment rather than inheritance), meaning that the hash interface `$sth->{Active}` doesn't work. Instead we need to call `dbi_sth()` first. Alternatively, attributes like _mysql_use_result_ can be passed to `prepare()`.

## Use case

In Compara we use eHive's DBConnections as 1) they can be constructed with an URL and 2) they are more resilient to database disconnections

## Benefits

With this change, SqlHelper compatible with eHive

## Possible Drawbacks

## Testing

### DIFF
--- a/modules/t/sqlHelper.t
+++ b/modules/t/sqlHelper.t
@@ -27,6 +27,8 @@ use Bio::EnsEMBL::Test::MultiTestDB;
 use Bio::EnsEMBL::Test::TestUtils;
 use Bio::EnsEMBL::Utils::SqlHelper;
 
+use Bio::EnsEMBL::Hive::DBSQL::DBConnection;
+
 #Redefine the WARN sig to note the errors (most are just from transaction retry)
 $SIG{__WARN__} = sub {
   note @_;
@@ -64,6 +66,12 @@ is(
   'Checking count of meta key is right with params'
 );
 
+my $hive_dbc = Bio::EnsEMBL::Hive::DBSQL::DBConnection->new(-dbconn => $dba->dbc);
+is(
+  $hive_dbc->sql_helper->execute_single_result(-SQL => qq{select count(*) from meta where meta_key = '$meta_key'}),
+  1,
+  'SqlHelper::_execute works with eHive DBConnections',
+);
 
 throws_ok { $helper->execute_single_result(-SQL => 'select * from meta') } qr/Too many results/, 'More than 1 row causes an error';
 throws_ok { $helper->execute_single_result(-SQL => 'select * from meta where species_id =?', -PARAMS => [-1]) } qr/No results/, 'Less than 1 row causes an error';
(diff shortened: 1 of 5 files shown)


magaliruffier:
Is this something that could be added to the tests?
Also, does this need to go into the release/94 branch or should it only be on master?

muffato:
I've expanded sqlHelper.t, it behaves as expected locally, but let's see if I've configured it correctly for Travis :)

I would like the change to be applied to both release/94 and master. I'll create the other PR once this one is accepted

magaliruffier:
Thanks for the tests, looks good to me

ens-bwalts:
This is a small change, done safely, that makes Ensembl's SqlHelper more flexible for compatibility with eHive's DBConnection - or other DBConnections for that matter.

Just for the record, I have to note that the changes to  SqlHelper.t and associated Travis files bring in the only dependency on eHive from the Ensembl codebase, and I'm not sure that it's necessary. Ensembl's  concern (it seems to me) is that the changes do not break SqlHelper when going through Ensembl's DBConnection. Whether or not it works correctly with other plugin DBConnections is the concern of the plugins, not Ensembl. Therefore, we don't need to bring in a dependency on eHive code to test that it works with eHive DBConnections - it is sufficient to test that Ensembl DBConnections still work.

AssemblyMapperAdaptor is the only place under DBSQL that invokes mysql_use_result, so no need to re-arrange the invocation in any other classes.

modules/t/sqlHelper.t
```diff
@@ -27,6 +27,8 @@ use Bio::EnsEMBL::Test::MultiTestDB;
 use Bio::EnsEMBL::Test::TestUtils;
 use Bio::EnsEMBL::Utils::SqlHelper;
 
+use Bio::EnsEMBL::Hive::DBSQL::DBConnection;
```
magaliruffier: What happens if a user wants to run this test without having a local checkout of Hive? Would it be possible to run this test only if sHive is available and not fail if it is not?
That way, Travis should still run fine and run this test, but we are not requiring external users to install eHive to be able to run the test suite

muffato: Hi mag,
You mean something like this https://github.com/Ensembl/ensembl-hive/blob/version/2.5/t/05.runnabledb/fastafactory.t#L33-L37 ?

magaliruffier: Exactly, or https://github.com/Ensembl/ensembl/blob/release/93/modules/t/utilsIo.t#L145

muffato:
I unfortunately actually agree with you @nerdstrike and I'm also looking at tieing a hash in eHive's StatementHandle so that all the keys like `Active` could transparently be called on `DBI::st` instead

muffato:
I'm confident it will work. I'll make a PR on ensembl-hive later this afternoon and cc you guys

PR #272, opened 2018-08-03
registry_public_cores.t now uses frozen test database

mkszuba:
## Description

The test _registry_public_cores.t_ now uses frozen test database

## Use case

Using the public registry server in tests is fragile, especially around release time.
See https://www.ebi.ac.uk/panda/jira/browse/ENSCORESW-2325

## Benefits

Registry tests can now work offline and not have to worry about version mismatches between the API and what the public server provides.

## Possible Drawbacks

Had to rewrite the "registry knows more than 10 DBAs" test to "registry knows at least 1 DBA" because by default that is what MultiTestDB provides. _Might make sense to tweak the settings to get at least two so that we can make sure the list  of DBAdaptor is really appended to._

## Testing

It was a test itself that needed changes.

The modified test passes.

Not run, under the assumption that a change in one test suite that does not write anything to the database really shouldn't affect any of the others.

### DIFF
--- a/modules/t/registry_public_cores.t
+++ b/modules/t/registry_public_cores.t
@@ -22,10 +22,24 @@ use Bio::EnsEMBL::Test::MultiTestDB;
 
 my $multi = Bio::EnsEMBL::Test::MultiTestDB->new();
 
-my $db = $multi->get_DBAdaptor('core');
-if ($db->dbc->driver() eq 'SQLite') {
-	plan skip_all => 'No registry support for SQLite yet';
+my $multi_config = $multi->db_conf();
+if (lc($multi_config->{'driver'}) ne 'mysql') {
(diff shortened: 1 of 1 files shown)


ens-bwalts:
Going by the name, it seems like the purpose of this test is to check access to public core DBs. Without testing load_registry_from_url(), we're left with tests that are similar (but not identical to) the tests in registry.t. Does these have enough value to keep this test around?

mkszuba:
Good point. Here is another take on refactoring this test: now we _do_ use load_registry_from_url() but instead of talking to the public DB server, we extract appropriate information from MultiTestDB configuration. Note that since the server in question is not guaranteed to be empty (in fact, it most likely will not be) one can get warnings about duplicate species/group combinations, in the end however the tests should work because there always is at least one homo_sapiens/core database - the frozen one loaded by the test harness.

PR #281, opened 2018-08-22
Update Bio::EnsEMBL::Exon->is_coding():

lairdm:
- Update documentation on required parameters
- Throw an error if a Transcript isn't given
- Add a test for regression

## Requirements

## Description

Updating Bio::EnsEMBL::Exon->is_coding(), documentation is out of date and parameters are properly scrutinized.

## Use case

So people know how to properly use the function, and can't abuse it.

## Benefits

Accurate documentation is good.

## Possible Drawbacks

Code that might not properly handle an exception, but honestly, if they're not passing a transcript in already the $transcript->translation would already blow up.

## Testing

Added a regression test to ensure the parameter is properly scrutinized.

Pass

Yes

### DIFF
--- a/modules/Bio/EnsEMBL/Exon.pm
+++ b/modules/Bio/EnsEMBL/Exon.pm
@@ -1281,8 +1281,7 @@ sub is_constitutive {
 
 =head2 is_coding
 
-  Arg [1]    : Boolean $is_coding
-  Arg [2]    : Bio::EnsEMBL::Transcript
+  Arg [1]    : Bio::EnsEMBL::Transcript
   Example    : $exon->is_coding()
   Description: Says if the exon is within the translation or not
   Returntype : Int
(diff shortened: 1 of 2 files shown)


lairdm:
Since there's no rush on this trivial fix, I'll let you assign a second reviewer next week and merge it, @avullo

modules/Bio/EnsEMBL/Exon.pm
```diff
@@ -1281,8 +1281,7 @@ sub is_constitutive {
 
 =head2 is_coding
 
-  Arg [1]    : Boolean $is_coding
-  Arg [2]    : Bio::EnsEMBL::Transcript
+  Arg [1]    : Bio::EnsEMBL::Transcript
   Example    : $exon->is_coding()
   Description: Says if the exon is within the translation or not
   Returntype : Int
```
tgrego: I know this isn't part of your changes, but shouldn't the return type be a Boolean or something like that?

lairdm: We should, if perl had a boolean type :)

PR #288, opened 2018-09-04
ENSCORESW-2792: separate mouse UCSC to avoid clashes

magaliruffier:
## Requirements

## Description

Mouse UCSC xref parsing uses a separate parser from human UCSC xrefs to avoid re-using IDs between the two sources.

## Use case

As we scale up the number of species, each source can be parsed by any species and the taxon id is extracted from the source. For UCSC xrefs, there is no information about the species in the original data. As a workaround, we use two separate parsers for the two species which use UCSC data.

## Benefits

We will map human xrefs to human and mouse xrefs to mouse, no cross-contamination. This will also prevent core foreign key breakages which happened before.

## Possible Drawbacks

It will not scale well if we wanted UCSC xrefs for more species.

## Testing

The pipeline was run on human and mouse in parallel to check for any possible contamination. 

There are no CoreForeignKey HC failure, the number of xrefs obtained is comparable to previous releases and the data looks sane.

NA

### DIFF
--- a/misc-scripts/xref_mapping/XrefParser/UCSC_mouse_Parser.pm
+++ b/misc-scripts/xref_mapping/XrefParser/UCSC_mouse_Parser.pm
@@ -17,6 +17,9 @@ limitations under the License.
 
 =cut
 
+# This module replicates the generic UCSC parser for mouse specific data
+# This prevents cross-mapping between species by treating each species as a separate source
+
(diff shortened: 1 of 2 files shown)


magaliruffier:
I thought of it but was trying to limit the number of places the code is updated.
If you reckon it makes it more solid longer term, I'll test the change now

PR #289, opened 2018-09-04
ENSCORESW-2744: remove dependent_xref when removing master

magaliruffier:
## Requirements

## Description

Remove all dependent_xref when the main xref is removed

## Use case

Before loading a new set of xrefs into the core database, existing xrefs from updated sources are removed from the core database. This is done source by source, meaning that dependent sources are only deleted if there is an update to that source. As their mapping depends on the master source, they should be removed at the same time.

## Benefits

If a dependent source is not updated but its master is, we do not end up with orphan entries. This also means we remove older mappings more regularly.

## Possible Drawbacks

The query is not optimised and can take a while to run on the server. It also leaves the xref entries in the xref table as they can be dependent of multiple sources.

## Testing

This was tested on gorilla which failed last release due to a single PDB that was left as an orphan. It is left as an orphan before the code change but is cleanly removed with this update. This was also run on a hundred species to check for regression.

The identified edge case is fixed and no new regressions are detected.

NA


magaliruffier:
It is not a bad suggestion. Currently, they are basically treated as unmapped entries, so they do not really harm anything, but it would be cleaner to get rid of them eventually.
I will put a ticket in for this but would rather avoid including it into this change as it would need to be done as a post-processing stage (or run the same query dozen of times as we got through each source)

magaliruffier:
luckily, we got rid of dependent dependents when we removed GO from the xref pipeline. If such a scenario ever happens again, we should spot it quickly as a number of unlinked entries will appear

PR #292, opened 2018-09-05
ENSCORESW-2553: retrieve correct accession for peptides

magaliruffier:
## Requirements

## Description

Retrieve RefSeq accessions for peptide when it is not stored as a stable ID

## Use case

For some species, the refseq_import format changes. The stable id is a randomly generated unique ID and the refseq accession is stored as an xref.
The code was updated to deal with this change but only transcript accessions were correctly retrieved. This change deals with the peptide accessions.

## Benefits

We have RefSeq accessions for peptides.

## Possible Drawbacks

## Testing

This was tested on the latest human refseq import in the new format and correctly stores the accessions and versions.

The results are now as expected

NA

### DIFF
--- a/misc-scripts/xref_mapping/XrefParser/RefSeqCoordinateParser.pm
+++ b/misc-scripts/xref_mapping/XrefParser/RefSeqCoordinateParser.pm
@@ -240,7 +240,7 @@ sub run_script {
 
 # Create a range registry for all the exons of the refseq transcript
       foreach my $transcript_of (sort { $a->start() <=> $b->start() } @$transcripts_of) {
-	my $id;
+	my ($id, $tl_id);
 	# We're moving to RefSeq accessions being stored as xrefs rather than
 	# stable ids. But we also need to maintain backwards compatbility.
 	# If it's the new kind, where there's a display_xref use that,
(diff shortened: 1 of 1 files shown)


misc-scripts/xref_mapping/XrefParser/RefSeqCoordinateParser.pm
```diff
@@ -405,12 +405,17 @@ sub run_script {
 # Also store refseq protein as direct xref for ensembl translation, if translation exists
           if (defined $tl && defined $tl_of) {
             if ($tl_of->seq eq $tl->seq) {
-              ($acc, $version) = split(/\./, $tl_of->stable_id());
+              $tl_id = $tl_of->stable_id();
+              my @xrefs = grep {$_->{dbname} eq 'GenBank'} @{$tl_of->get_all_DBEntries};
+              if(scalar @xrefs == 1) {
```
nerdstrike: From this we may infer that there can only ever be 1 or 0 xrefs in otherfeatures?

magaliruffier: of source GenBank. There are other sources but they do not provide the RefSeq accession

PR #293, opened 2018-09-05
Feature/xref config

magaliruffier:
## Requirements

## Description

Xref configuration file cleaned up by removing deprecated sources and using default settings for each division

## Use case

The current configuration file is hard to maintain. It is also not clear what needs to be added where when sources or species get added. This is the second pass at cleaning it up, mostly focusing on duplicated sources for non-vertebrates.

## Benefits

The file has been simplified, hopefully making it easier to maintain in the future.
There are default settings for most divisions, meaning there is no need to update the file when a new species is included.

## Possible Drawbacks

## Testing

The pipeline was run on all vertebrate species with the complete set of changes. Results were compared with those from a previous run to ensure all sources are still correctly added for all species.

The results are comparable to previous run.

xref_config2sql.pl runs without errors

### DIFF
--- a/misc-scripts/xref_mapping/xref_config.ini
+++ b/misc-scripts/xref_mapping/xref_config.ini
@@ -825,40 +825,6 @@ parser          = CeleraTranscriptParser
 release_uri     =
 data_uri        = file:Celera/Anopheles_Celera_consensus-transcripts_xref.fasta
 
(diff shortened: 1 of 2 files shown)


magaliruffier:
Fair point, I will review the documentation now

PR #295, opened 2018-09-07
ENSCORESW-2850: optimised for single species run

magaliruffier:
## Requirements

## Description

This ensures the xref_parser script works correctly for a single species with a simplified configuration file.

## Use case

The xref code has been largely updated to accommodate a hive-based pipeline with less manual intervention. As part of this, a default configuration for each division has been set up, removing the need to add every new species in the configuration file. This has broken some of the assumptions in the historical parser script. The proposed fix separates clearly the species_id, which is used as a parameter in the various parsers to retrieve the taxon-specific data, and the division_id, which is used to know which sources to run.

## Benefits

The parser can be run as a standalone job on any species and source.

## Possible Drawbacks

The parser cannot be run on more than one species at a time. This should be done with the eHive pipeline instead.

## Testing

The script was run on some sample species.

Before the change, sources get submitted correctly for the division, but any source requiring connection to the species database fail, as it is using the division_id rather than the species_id.
After the change, all sources are submitted and run correctly.

NA

### DIFF
--- a/misc-scripts/xref_mapping/XrefParser/ProcessData.pm
+++ b/misc-scripts/xref_mapping/XrefParser/ProcessData.pm
@@ -55,7 +55,7 @@ sub run {
   my $unzip      = $ref_arg->{unzip};
   my $stats      = $ref_arg->{stats};
   my $cleanup    = $ref_arg->{cleanup};
-  my $rspecies   = $ref_arg->{speciesr};
+  my $species    = $ref_arg->{species};
   my $taxon_id   = $ref_arg->{taxon};
   my $division   = $ref_arg->{division};
   my $sources    = $ref_arg->{sourcesr};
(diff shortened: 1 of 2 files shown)


misc-scripts/xref_mapping/XrefParser/ProcessData.pm
```diff
@@ -101,21 +101,23 @@ DSS
...
-  exit(1) if ( !$self->validate_sources(\@species_ids,$sources, $verbose) );
-  exit(1) if ( !$self->validate_sources(\@species_ids,$notsources, $verbose) );
+  exit(1) if ( !$self->validate_sources(\@species_sources,$sources, $verbose) );
+  exit(1) if ( !$self->validate_sources(\@species_sources,$notsources, $verbose) );
 
   # build SQL
   my $species_sql = "";
-  if (@species_ids) {
+  if (@species_sources) {
     $species_sql .= " AND su.species_id IN (";
```
premanand17: $species_sql initialised to "" at line 115.  If it enters the if block, the resulting $species_sql might not be a valid SQL.

misc-scripts/xref_mapping/XrefParser/ProcessData.pm
```diff
@@ -682,26 +684,21 @@ sub dbi {
...
-    } else {
-      print STDERR "Species $sp is not valid; valid species are:\n";
-      $self->show_valid_species();
-    }
+  my $bind_arg = "^".lc($species).",|^".lc($species)."\$|,[ ]{0,1}".lc($species)."[ ]{0,1},|,[ ]{0,1}".lc($species)."\$";
+  $sth->execute(lc($species), $bind_arg ); 
+  $sth->bind_columns(\$species_id, \$species_name);
+  if (my @row = $sth->fetchrow_array()) {
+    print "Species $species is valid (name = " . $species_name . ", ID = " . $species_id . ")\n" if($verbose);
+  } else {
```
premanand17: If block has just the print statement?!
Where is the $species_id getting initialised?!!
Probably the if block is missing $species_id initialisation?!!!

magaliruffier: the fetchrow_array method populates the species_id and species_name values

PR #296, opened 2018-09-10
C. elegans references use WormBase mapping to INSDC protein ids

wbazant:
### Changes affecting vertebrates run of the pipeline
*None. I didn't even change anything apart from one method extracted from a block in RefSeqGPFFParser. The rest of the new behaviour was added by subclassing and inheritance. *

### All changes
- rewrite WormbaseDirectParser
- WormbaseDirectParser populates protein_ids
- superclass method to make dependent protein_ids as parent
- tap into UniProtParser
  + also skip EMBL scaffold ids (we can't reliably assign them)
- tap into RefSeqGPFFParser
  + extract a method
- tests for new stuff
  + add %args to parametrise test_parser

## Use case

RefSeq proteins have coordinates as part of their identity, so we
can't reliably sequence match them, we will also pick up all paralogs.

## Benefits

For Ensembl:
The method extracted in RefSeqGPFFParser is a very minor change for the better. Test coverage for UniProtParser and RefSeqGPFFParser.

Organisational:
Easier to maintain if WormBase specific behaviour is inside the code base as long as we're responsible. Ensembl can refactor without breaking our code (and a test will fail if it breaks).

Benefits - RefSeqGPFFParser:
This change fixes this spurious mapping of paralogs.

Benefits - UniProtParser:
Not the above: UniProt entries are not tied to coordinates so all
paralogs map to the same entry. We can handle versioning and updates
a bit better: if WormBase updates an entry and a protein id changes but
UniProt doesn't reflect this yet, with the change we will still pick up
the UniProt entry although we can't sequence match any more.

## Possible Drawbacks

The repo will take a bit longer to download. The test suite will run a bit longer.

## Testing

I unit tested relevant old code and new code. I did a run for C. elegans, and the results looked vaguely okay.


wbazant:
The newest DBD::mysql is broken. I've raised the issue in https://github.com/perl5-dbi/DBD-mysql/issues/263. My changes to the cpanfile fix this build, but maybe you had a reason to `require` rather than `test_require` DBD::mysql?

magaliruffier:
DBD::mysql should be a requirement, our whole API depends on it
We are looking into it, so you can ignore this error for the moment.
Your tests fail with the latest Perl 5.26 though

magaliruffier:
Overall, the changes look fine to me, it is also nice to get more tests.
I would like to do a test run on a handful of vertebrate species to ensure we have not missed any dependencies, this will not happen before the 1st week of October.
Are you OK with keeping the PR opened until then or should we restart the conversation once I'm back?

wbazant:
@magaliruffier I have adjusted the commit, changing the names of the files. Can you review again?

The files that say "Wormbase" at the front are meant to be specific to the WormBase run of the pipeline. Reviewing it to search for bugs and giving @khowe and me tips on good software engineering techniques is appreciated, but perhaps above and beyond necessary: a reviewer might satisfy themselves with searching for the string "Wormbase" in the Ensembl codebase to verify that the files are not referred to by the vertebrate run of the pipeline.

I have also changed a test: there's no CI at the moment so I can only assert that the tests pass at least on the Ensembl's installation of Perl in the EBI.

The only change in this PR that doesn't say "Wormbase" at the front or is a test file is in RefSeqGPFFParser. It is of relatively simple nature - it extracts a method and changes block indentation - and perhaps a reviewer could just read it. I am happy to justify it more or engage into a discussion about what it might affect or be done better.

I am not sure which bit of this PR you are unsure about, that completing a run on a handful of vertebrate species could help clear out. It is highly unfortunate if this were necessary, could only be done by you (can I do this run? what should I pay attention to that might be made different by this change? ) and could only happen in a month.

wbazant:
@magaliruffier I've reused the PR for some more work on this. I've done minor changes to RefSeqGPFFParser that don't intend to change behaviour. 

For C. elegans in our run:
- UniProtParser: entries linking to multiple EMBL proteins create multiple xrefs
- new cross-reference to wormbase_cds
- our subclassed RefSeqGPFFParser can do both protein and transcript entries
- use wormbase_cds to link mRNA and ncRNA RefSeq entries

Can you have a look? The CI fails but I think it's unrelated.

magaliruffier:
Overall, the changes look fine to me. I should have some time this week to do a test run, then it should all be good to go.
Travis fails because of some discrepancy between the release versions, if you rebase against the latest master, that should fix it.

wbazant:
I only needed to rebase to fix the build. Now 1 successful check has passed. @magaliruffier would you also be able to arrange at least 1 approving review by reviewers with write access?

magaliruffier:
I was doing a test run to compare with an existing run on the same data but without the code change.
Unfortunately, it is still running when it should have finished by now. My pipeline is stuck in the parsing stage, parsing RefSeq GPFF files. According to eHive, the average takes over 1h when it took 4 minute in the previous run.
It could still be that there is a problem with the database load or the file system that is independent of the code changes, so I will try again and let you know.
Otherwise, it could be that calling the sub method every time eats up more time than expected

wbazant:
@magaliruffier That's super interesting! :) at the beginning of  line 209 (that sub `xref_from_record` I extracted) I'm copying a bunch of hashes. It's intentional  - I didn't think it will be very wasteful and I wanted to change fewer things - but it's the culprit, it's easy to fix.

wbazant:
I'm running the parser under a profiler now, I will see what shows up.

wbazant:
[prof-after.txt](https://github.com/Ensembl/ensembl/files/2491523/prof-after.txt)
[prof-before.txt](https://github.com/Ensembl/ensembl/files/2491524/prof-before.txt)
I'm not sure why the if statement block shows up as taking a lot of time before, and is gone after. Thank you for catching that! My pipeline runs also took ages, but I didn't really reflect on it :)

magaliruffier:
OK, let me try that again ;-)

magaliruffier:
Run times are comparable now, average 5 min per job. Still a bit more than the previous run, but that's more in the line of just general variability due to resource contention
I'll let the pipeline run to completion to compare the overall result and will let you know how it goes

magaliruffier:
I finally managed to get two comparable runs of the pipeline, once with and once without your changes.
I can confirm that the results are mostly the same. There are some small differences which I believe are due to consecutive runs rather than code changes.
I'll let someone else from Core have a look at the code but otherwise I think it is good to be merged in

misc-scripts/xref_mapping/XrefParser/WormbaseCElegansBase.pm
```diff
@@ -0,0 +1,57 @@
...
+  }
+  my @result;
+  for my $matching_source_id_dependent (@matching_source_id_dependents) {
+    my $source_id = $matching_source_id_dependent->{SOURCE_ID};
+    my $xref_as_dependent_here = {
+      %$xref, INFO_TYPE => "DEPENDENT",
+      LINKAGE_SOURCE_ID => $source_id,
+      DEPENDENT_XREFS => undef,
+    };
+    my @other_dependents_as_dependents_here = map {{%$_,INFO_TYPE => "DEPENDENT", LINKAGE_SOURCE_ID => $source_id}} @other_dependents;
```
tgrego: Perhaps better written as:
`my @other_dependents_as_dependents_here = map {`
`	$_->{INFO_TYPE} = "DEPENDENT";`
`	$_->{LINKAGE_SOURCE_ID} => $source_id;`
`	$_`
`} @other_dependents;`

This is what I believe you are trying to do, but I might be wrong as it is very tricky to understand...
consider refactoring and maybe even replace with a for as suggested by perlcritic (See page 113 of PBP)

misc-scripts/xref_mapping/XrefParser/WormbaseCElegansRefSeqGPFFParser.pm
```diff
@@ -0,0 +1,67 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+=cut
+
+package XrefParser::WormbaseCElegansRefSeqGPFFParser;
+use strict;
```
tgrego: also `use warnings;`

misc-scripts/xref_mapping/XrefParser/WormbaseCElegansRefSeqGPFFParser.pm
```diff
@@ -0,0 +1,67 @@
...
+   );
+}
+
+sub modify_xref_with_dependent {
+   my ($source_ids, $entry, $xref, $pattern) = @_;
+   return unless $xref;
+   return unless $entry =~ $pattern;
+   return unless $1;
+   $xref->{DEPENDENT_XREFS} //= [];
+   push @{$xref->{DEPENDENT_XREFS}}, map {{ACCESSION => $1, SOURCE_ID=>$_}} @$source_ids;
```
tgrego: Capture variable used outside conditional. See page 253 of PBP.
Basically do the `$entry =~ $pattern` in a conditional that assigns the match to a properly named variable, you can then return if that variable (say `$acc`?) in undef or use it in the map.

misc-scripts/xref_mapping/XrefParser/WormbaseCElegansUniProtParser.pm
```diff
@@ -0,0 +1,43 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+=cut
+
+package XrefParser::WormbaseCElegansUniProtParser;
+
```
tgrego: use strict and warnings

misc-scripts/xref_mapping/XrefParser/WormbaseCElegansUniProtParser.pm
```diff
@@ -0,0 +1,43 @@
...
+# INSDC entries which we get from somewhere else
+# Attempt to find the parent (has to already be present in the xref table)
+# INSDC and UniProt entries have the same protein sequence, and
+# UniProt lists INSDC as a parent. We get INSDC entries from somewhere else,
+# so make UniProt entries dependent on INSDC entries.
+# Note:
+# INSDC entries have coordinates, and UniProt entries don't.
+# So for perfect homologs, there can be many INSDC entries per UniProt.
+
+use parent XrefParser::WormbaseCElegansBase, XrefParser::UniProtParser;
```
tgrego: safer to qw() the parents

misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
```diff
@@ -164,30 +164,48 @@ sub create_xrefs {
...
-  if ($file =~ /protein/) {
+  my $type = $self->type_from_file($file);
+  return unless $type;
 
-    $type = 'peptide';
-
-  } elsif ($file =~ /rna/) {
-
-    $type = 'dna';
+  while ( $_ = $refseq_io->getline() ) {
```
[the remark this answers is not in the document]
wbazant: Nice one! In the previous code $_ was referred to in a method body , even though the code also did  `my $entry = $_;`. I only saw it because it stopped referring to anything when I applied your remark and I got a warning. So I think the behaviour is still the same, but more futureproof when the previous version will start doing something else when $_ gets clobbered.

misc-scripts/xref_mapping/XrefParser/WormbaseCElegansBase.pm
```diff
@@ -0,0 +1,57 @@
...
+
+package XrefParser::WormbaseCElegansBase;
+
+sub swap_dependency {
+  my ($self, $source_ids, $dbi, $xref, @source_ids_skip) = @_;
+  $source_ids = [$source_ids] unless ref $source_ids eq 'ARRAY';
+
+  my @matching_source_id_dependents;
+  my @other_dependents;
+  for my $dependent_xref (@{$xref->{DEPENDENT_XREFS} || []}){
```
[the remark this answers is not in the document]
wbazant: It's the same as
```
$xref->{DEPENDENT_XREFS} ? @{$xref->{DEPENDENT_XREFS}} : ()
```
i.e. don't iterate if there's nothing.

misc-scripts/xref_mapping/XrefParser/WormbaseCElegansBase.pm
wbazant: You understood it correctly! I don't want to clobber  entries in `@other_dependents` because I use it multiple times. Although I overwrite the same fields every time, so maybe it doesn't matter. I rewrote it with a for loop, maybe it's better now.

wbazant:
@tgrego thank you for the review! See if I addressed everything. If you apply this PR, refactor RefSeqGPFFParser on top of that, and the unit tests still pass, WormBase stuff will fairly likely still be good but I'll follow the developments!

PR #297, opened 2018-09-11
ENSCORESW-2856: explicitly add DBD dependency

magaliruffier:
## Requirements

## Description

Import Devel::CheckLib with the other dependencies

## Use case

DBD::mysql has a dependency on Devel::CheckLib which is not correctly imported by most versions of cpan. This causes Travis to fail when setting up a new environment. The proposed fix imports the library explicitly

## Benefits

All dependencies are correctly loaded in the tested environments. In particular, the test suite passes, removing noise when checking code changes.

## Possible Drawbacks

We add one more dependency that is not necessarily required for Ensembl and might disappear in the future.

## Testing

No changes required.

NA

Yes

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -4,6 +4,7 @@ requires 'HTTP::Tiny';
 requires 'IO::Compress::Gzip';
 requires 'URI::Escape';
 requires 'Config::IniFiles';
+requires 'Devel::CheckLib';
 
 test_requires 'Test::Warnings';
 test_requires 'Test::Differences';
--- a/cpanfile
+++ b/cpanfile
@@ -1,10 +1,10 @@
+requires 'Devel::CheckLib';
 requires 'DBI';
 requires 'DBD::mysql';
 requires 'HTTP::Tiny';
 requires 'IO::Compress::Gzip';
 requires 'URI::Escape';
 requires 'Config::IniFiles';
-requires 'Devel::CheckLib';
 
 test_requires 'Test::Warnings';
 test_requires 'Test::Differences';


lairdm:
Le sigh. Not my preferred solution, I wish DBD::mysql could address the dependency more elegantly. However reading the Checklib docs a little, the fact it's in Makefile.PL does make it difficult for them to address. Of course this will still have ripple effects on other repos that need DBD::mysql but not Ensembl (eg ensembl-hive), so this prairie dog will pop up a few more times before this is over.

lairdm:
Oops, order might matter.

wbazant:
I made an issue and now get all the emails! It looks like the DBD::mysql know what the problem is and will just fix it. Maybe you'd like to just wait for that? https://github.com/perl5-dbi/DBD-mysql/issues/263#issuecomment-420180077

wbazant:
@lairdm it could be worse than "order matters" : https://github.com/miyagawa/cpanfile/issues/42 says these things are put in a hash and then the order is undefined :$

magaliruffier:
I'm OK waiting until later today to see if that will be fixed on the cpan side of things
If it goes on for longer though, a temporary solution that allows us to check our code rather than external dependencies seems like the lesser of two evils

magaliruffier:
The original error in the library has not been fixed, I can see 3 solutions:

- keep waiting for the fix. As all our test builds will fail in the mean time, we should probably not merge any more PRs until then
- include the dependency explicitly as in this PR. There might still be issues with the order in which dependencies are loaded
- fix the DBD::mysql version to 4.046

Any thoughts?

magaliruffier:
I'll try the set version

PR #306, opened 2018-10-18
EntrezGene/HPA parsers update

avullo:
## Requirements

## Description

Refactoring the parsers to consider:
- consistent error handling
- code compression, clarity
- NULL fields where applicable without touching BaseParser at the moment, i.e. forcing NULL description when adding xrefs in HPA parser

## Use case

Xref pipeline for species with EntrezGene/HPA sources

## Benefits

Code quality improvement

## Possible Drawbacks

According to the guidelines, not fully there yet. Need to change the BaseParser and schema to force NULL info_text instead of empty string.

## Testing

No unit tests at the moment I'm afraid. Run the xref_parser script with the current version and proposed update and found no difference except attribute description is now NULL for xrefs from HPA (formerly '').

### DIFF
--- a/misc-scripts/xref_mapping/XrefParser/Database.pm
+++ b/misc-scripts/xref_mapping/XrefParser/Database.pm
@@ -1,3 +1,4 @@
+
 =head1 LICENSE
 
 Copyright [1999-2015] Wellcome Trust Sanger Institute and the EMBL-European Bioinformatics Institute
@@ -29,40 +30,41 @@ use IO::File;
 use English;
 
 sub new {
-  my ($proto, $arg_ref) = @_;
+  my ( $proto, $arg_ref ) = @_;
 
   my $class = ref $proto || $proto;
-  my $self =  bless {}, $class;
-  
-  $self->dbc(ref $arg_ref eq 'Bio::EnsEMBL::DBSQL::DBConnection' 
-    ? $arg_ref : Bio::EnsEMBL::DBSQL::DBConnection->new(
-    -HOST => $arg_ref->{host},
-    -DBNAME => $arg_ref->{dbname},
-    -USER => $arg_ref->{user},
-    -PASS => $arg_ref->{pass} || '',
-    -PORT => $arg_ref->{port} || '3306'
-  ));
-  $self->verbose($arg_ref->{verbose});
+  my $self = bless {}, $class;
+
+  $self->dbc(
+        ref $arg_ref eq 'Bio::EnsEMBL::DBSQL::DBConnection' ? $arg_ref :
+          Bio::EnsEMBL::DBSQL::DBConnection->new(
+                                     -HOST   => $arg_ref->{host},
+                                     -DBNAME => $arg_ref->{dbname},
+                                     -USER   => $arg_ref->{user},
+                                     -PASS   => $arg_ref->{pass} || '',
+                                     -PORT => $arg_ref->{port} || '3306'
+          ) );
+  $self->verbose( $arg_ref->{verbose} );
 
   return $self;
 }
 
 sub verbose {
-  my ($self, $arg) = @_;
+  my ( $self, $arg ) = @_;
 
-  (defined $arg) &&
-    ($self->{_verbose} = $arg );
+  ( defined $arg ) && ( $self->{_verbose} = $arg );
   return $self->{_verbose};
 }
 
 sub dbc {
   my $self = shift;
 
-  if(@_) {
+  if (@_) {
     my $arg = shift;
 
-    if(defined($arg)) {
-      croak "$arg is not a DBConnection" unless $arg->isa('Bio::EnsEMBL::DBSQL::DBConnection');
+    if ( defined($arg) ) {
+      croak "$arg is not a DBConnection"
+        unless $arg->isa('Bio::EnsEMBL::DBSQL::DBConnection');
       $self->{_dbc} = $arg;
     }
   }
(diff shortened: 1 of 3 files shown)


misc-scripts/xref_mapping/XrefParser/Database.pm
```diff
@@ -57,11 +56,21 @@ sub verbose {
...
-  (defined $arg) &&
-    ($self->{_dbc} = $arg );
+  my $self = shift;
+
+  if(@_) {
+    my $arg = shift;
+
+    if(defined($arg)) {
+      croak "$arg is not a DBConnection" 
+	unless $arg->isa('Bio::EnsEMBL::DBSQL::DBConnection');
```
nerdstrike: I thought this was an indent issue, but it's a tab. Please remove the tabs

misc-scripts/xref_mapping/XrefParser/EntrezGeneParser.pm
```diff
@@ -50,74 +50,29 @@ sub run {
   
 
   my $eg_io = $self->get_filehandle($file);
-  if ( !defined $eg_io ) {
-    print STDERR "ERROR: Could not open $file\n";
-    return 1;    # 1 is an error
-  }
+  croak "ERROR: Could not open $file\n" unless defined $eg_io;
```
nerdstrike: Errors are self-evidently errors when the croak fires. Is croak correct for this?

misc-scripts/xref_mapping/XrefParser/EntrezGeneParser.pm
```diff
@@ -50,74 +50,29 @@ sub run {
...
 
-  my %seen;
+  my $input_file = Text::CSV->new({ sep_char           => "\t",
+				    empty_is_undef     => 1,
+				    allow_loose_quotes => 1 # turns out file has unescaped quotes
```
nerdstrike: Tabs tabs tabs...

misc-scripts/xref_mapping/XrefParser/EntrezGeneParser.pm
```diff
@@ -50,74 +50,29 @@ sub run {
...
   my $eg_io = $self->get_filehandle($file);
-  if ( !defined $eg_io ) {
-    print STDERR "ERROR: Could not open $file\n";
-    return 1;    # 1 is an error
-  }
+  croak "ERROR: Could not open $file\n" unless defined $eg_io;
 
-  my %seen;
+  my $input_file = Text::CSV->new({ sep_char           => "\t",
+				    empty_is_undef     => 1,
```
nerdstrike: Could you turn on strict here instead of checking for errors later?

misc-scripts/xref_mapping/XrefParser/EntrezGeneParser.pm
```diff
@@ -50,74 +50,29 @@ sub run {
...
+				    allow_loose_quotes => 1 # turns out file has unescaped quotes
+				  }) or croak "Cannot use file $file: ".Text::CSV->error_diag ();
 
+  # process header
+  $input_file->column_names( @{ $input_file->getline( $eg_io ) } );
```
nerdstrike: $csv->header($eg_io) might suffice for this, depending on the first line of your file.

misc-scripts/xref_mapping/XrefParser/HPAParser.pm
```diff
@@ -39,70 +42,55 @@ sub run {
...
-
-  while ( defined( my $line = $file_io->getline() ) ) {
-
-    $line =~ s/\s*$//;
-		
-
-    my ( $antibody, $antibody_id, $ensembl_peptide_id, $link) = split( /,/, $line );
+  my $parsed_count = 0;
+  while ( my $row = $input_file->getline( $file_io ) ) {
+    map { $_ =~ s/\s*$// } @{$row};
```
nerdstrike: maps with no return are a perlcritic warning. You could just feed these into the variables defined below if you must map{}.

misc-scripts/xref_mapping/XrefParser/HPAParser.pm
```diff
@@ -39,70 +42,55 @@ sub run {
...
+    my $version = '1';
 
     my $xref_id = $self->get_xref( $antibody_id, $source_id, $species_id, $dbi );
+    $xref_id = $self->add_xref({ acc        => $antibody_id,
+				 version    => $version,
+				 label      => $label,
+				 source_id  => $source_id,
+				 species_id => $species_id,
+				 dbi        => $dbi,
+				 info_type  => "DIRECT"} ) unless defined $xref_id or $xref_id ne '';
```
nerdstrike: Complex postfix condition is important, I would prefer it up front as this is multi-line and multi-term.

misc-scripts/xref_mapping/XrefParser/EntrezGeneParser.pm
avullo: From the docs: "If this attribute (strict) is set to 1, any row that parses to a different number of fields than the previous row will cause the parser to throw error 2014."

I am not checking for errors later, other than checking parsing has reached EOF. This does not seem to apply to the above.

misc-scripts/xref_mapping/XrefParser/HPAParser.pm
avullo: They would all be set to 1, which is what the expression inside map evaluates to. I have to be boring by removing map and make explicit statements for each one of the declared variables.

misc-scripts/xref_mapping/XrefParser/HPAParser.pm
avullo: It's a matter of preference, but I can make you happy :)

misc-scripts/xref_mapping/XrefParser/EntrezGeneParser.pm
avullo: Or maybe you meant that for the HPA parser?!

misc-scripts/xref_mapping/XrefParser/EntrezGeneParser.pm
```diff
@@ -20,10 +20,13 @@ limitations under the License.
 package XrefParser::EntrezGeneParser;
 
 use strict;
+
+use Carp;
 use POSIX qw(strftime);
 use File::Basename;
```
mkszuba: Are these two modules actually used in the parser?

misc-scripts/xref_mapping/XrefParser/EntrezGeneParser.pm
```diff
@@ -32,14 +35,11 @@ sub run {
   my $species_id   = $ref_arg->{species_id};
   my $species_name = $ref_arg->{species};
   my $files        = $ref_arg->{files};
-  my $verbose      = $ref_arg->{verbose};
-  my $dbi          = $ref_arg->{dbi};
-  $dbi = $self->dbi unless defined $dbi;
+  my $verbose      = $ref_arg->{verbose} || 0;
+  my $dbi          = $ref_arg->{dbi} || $self->dbi;
```
mkszuba: I would suggest checking definedness rather than truthiness (_i.e._ using // rather than ||) here, in case of DBI it is probably harmless but unless I am mistaken setting verbosity will no longer work correctly with || if the default value gets changed to non-zero and someone tries switching it back to 0 at run time.

misc-scripts/xref_mapping/XrefParser/EntrezGeneParser.pm
```diff
@@ -50,74 +50,27 @@ sub run {
...
 
   my $eg_io = $self->get_filehandle($file);
-  if ( !defined $eg_io ) {
-    print STDERR "ERROR: Could not open $file\n";
-    return 1;    # 1 is an error
-  }
+  croak "Could not open $file\n" unless defined $eg_io;
 
-  my %seen;
+  my $input_file = Text::CSV->new({ sep_char => "\t", empty_is_undef => 1, allow_loose_quotes => 1 })
```
mkszuba: I would split this into multiple lines, both for better readability and to keep line length under control.

misc-scripts/xref_mapping/XrefParser/HPAParser.pm
```diff
@@ -39,70 +42,61 @@ sub run {
...
-    if ( !defined($xref_id) || $xref_id eq '' ) {
-      $xref_id = $self->add_xref({ acc        => $antibody_id,
-				   version    => $version,
-				   label      => $label,
-				   desc       => $description,
-				   source_id  => $source_id,
-				   species_id => $species_id,
-                                   dbi        => $dbi,
-				   info_type  => "DIRECT"} );
+    unless(defined $xref_id or $xref_id ne '') {
```
mkszuba: Leaving aside the fact that I am really not a fan of using unless for anything more complicated than "unless defined $foo", especially with double negatives (the "ne" bit), I am quite sure this condition is in fact different than what it was before the change - (NOT a) OR (NOT b) <=> NOT (a **AND** b). That would imply either the original version was wrong or your current one is. Could you double-check, please?

Oh, and one way or another I would suggest using q{} to indicate an empty string. Or just replace the "string defined + non-empty" condition with a truthiness check, I am pretty sure MySQL does not like 0 in auto-increment numeric columns so we needn't worry about the string "0".

misc-scripts/xref_mapping/XrefParser/Database.pm
avullo: Done.

misc-scripts/xref_mapping/XrefParser/HPAParser.pm
avullo: Now with explicit statements.

PR #308, opened 2018-10-18
Bugfix/delint DBASS parser

mkszuba:
## Description

Fixes bugs observed (so far) in DBASSParser during the xref sprint, add additional checks, and delint the code to facilitate further refactoring. See ENSCORESW-2886.

## Use case

Part of the efforts to improve the xref pipeline. Moreover, one of the observed bugs actually prevents current versions of DBASS3 data from being fully parsed.

## Benefits

DBASS3 input can now be parsed properly. Some future-proofing. Code (hopefully) easier to maintain. Passes PerlCritic inspection at level 3 and the only complaints at level 2 are missing certain POD sections and $VERSION. Use a standardised rather than hand-rolled CSV parser, with potential for a performance increase if compiled rather than native-Perl version of the parser is used.

## Possible Drawbacks

None I can think of.

## Testing

No.

N/A

N/A. However, I have run the parser itself on both current DBASS data and some intentionally malformed input, and it appears to work correctly.

### DIFF
--- a/misc-scripts/xref_mapping/XrefParser/DBASSParser.pm
+++ b/misc-scripts/xref_mapping/XrefParser/DBASSParser.pm
@@ -72,14 +72,19 @@ sub run {
     # assumes actual column values will never contain commas, which while
     # true at present brings into question why bother quoting them in the
     # first place.
-    # FIXME: add a sanity check for that
     $line =~ s/"//gx;
 
-    my ( $dbass_gene_id, $dbass_gene_name, $ensembl_id ) =
+    my ( $dbass_gene_id, $dbass_gene_name, $ensembl_id, @split_overflow ) =
       split( /,/x, $line );
 
     if ( !defined($dbass_gene_id) || !defined($ensembl_id) ) {
       printf STDERR ( "Line %d has fewer than three columns.\n",
+                      1 + $parsed_count );
+      print STDERR ("The parsing failed\n");
+      return 1;
+    }
+    elsif ( scalar @split_overflow > 0 ) {
+      printf STDERR ( "Line %d has more than three columns.\n",
                       1 + $parsed_count );
       print STDERR ("The parsing failed\n");
       return 1;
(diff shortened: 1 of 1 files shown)


misc-scripts/xref_mapping/XrefParser/DBASSParser.pm
```diff
@@ -20,124 +20,185 @@ limitations under the License.
...
+  # match to begin with
+  if ( scalar @{ $header } != $EXPECTED_NUMBER_OF_COLUMNS ) {
+    return 0;
+  }
+
+  my @fields_ok;
+
+  my ( $dbass_end ) = ( $header->[0] =~ m{ DBASS (3|5) GeneID }msx );
+  push @fields_ok, defined $dbass_end;
+
```
premanand17: Why the variable is named as $dbass_end?  As it is capturing the header (DBASS3GeneID/DBASS5GeneID), could be named as $dbass_geneid or $dbass_geneid_ok for consistency with other variable names below.

Also 'defined' check is done only for the first field and not for other fields?!

premanand17: Good. Well explained.

magaliruffier:
Current linkout is done using the gene ID, which I would assume is stored in the accession field
See http://www.ensembl.org/Homo_sapiens/Gene/Matches?g=ENSG00000139618;r=13:32315474-32400266 and http://www.dbass.soton.ac.uk/DBASS3/viewlist.aspx?filter=gene&id=15

misc-scripts/xref_mapping/XrefParser/DBASSParser.pm
```diff
@@ -20,124 +20,182 @@ limitations under the License.
...
+                                    (.*)
+                                    \s?\/\s?  # typically no ws here but just in case
+                                    (.*)
+                                  }msx ) ||
+           ( $dbass_gene_name =~ m{
+                                    (.*)
+                                    \s?  # typically there IS a space before the ( here
+                                    [(] (.*) [)]
+                                  }msx ) ) {
+        $first_gene_name  = $1;
```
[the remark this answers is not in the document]
mkszuba: Makes sense. Added a comment block with examples.

mkszuba:
> Current linkout is done using the gene ID, which I would assume is stored in the accession field

Correct. Gene ID is the accession, primary gene name is the label, second name (if any) is a synonym.

PR #310, opened 2018-10-18
VGNC parser update

tgrego:
## Description

Update to the VGNC parser during the first stage of the xref sprint. See ENSCORESW-2886.
Updates were as agreed... no $_ variable handling, croak on error, null instead of '', linting in general...
Parser now using Text::CSV.
UPDATED: ENSCORESW-2835 also dealt with.

## Testing

no unit tests.
Manually tested with chimpanzee.
UPDATED:
Results are consistent.

OUTDATED:
I have got 16101 xrefs with old code and 16105 with new...
Xtra Xrefs are 
'9556', 'VGNC:57780', '0', 'VIPR1', 'vasoactive intestinal peptide receptor 1', '146', '9598', 'DIRECT', '', NULL
'9564', 'VGNC:57781', '0', 'WDR82', 'WD repeat domain 82', '146', '9598', 'DIRECT', '', NULL
'9566', 'VGNC:57782', '0', 'WNT7A', 'Wnt family member 7A', '146', '9598', 'DIRECT', '', NULL
'9540', 'VGNC:57783', '0', 'UBXN7', 'UBX domain protein 7', '146', '9598', 'DIRECT', '', NULL

### DIFF
--- a/misc-scripts/xref_mapping/XrefParser/VGNCParser.pm
+++ b/misc-scripts/xref_mapping/XrefParser/VGNCParser.pm
@@ -62,7 +62,27 @@ sub run {
     empty_is_undef => 1
   }) or croak "Cannot use file $file: ".Text::CSV->error_diag ();
 
-  $input_file->column_names( @{ $input_file->getline( $file_io ) } );
+  # header must contain these columns
+  my @required_columns = qw(
+    taxon_id
+    ensembl_gene_id
+    vgnc_id symbol
+    name
+    alias_symbol
+    prev_symbol
+  );
+
+  # get header columns
+  my @columns = @{ $input_file->getline( $file_io ) };
+
+  # die if some required_column is not in columns
+  foreach my $colname (@required_columns) {
+    if ( !grep { /$colname/xms } @columns ) {
+      croak "ERROR: Can't find required column \"$colname\" in VGNC file $file\n";
+    }
+  }
+
+  $input_file->column_names( @columns );
 
   while ( my $data = $input_file->getline_hr( $file_io ) ) {
 
--- a/misc-scripts/xref_mapping/XrefParser/VGNCParser.pm
+++ b/misc-scripts/xref_mapping/XrefParser/VGNCParser.pm
@@ -37,7 +37,7 @@ sub run {
 
 
   if ( (!defined $source_id) or (!defined $species_id) or (!defined $files) ) {
-    croak 'Need to pass source_id, species_id and files';
+    croak "Need to pass source_id, species_id and files as pairs";
   }
 
   my $file = shift @{$files};
@@ -47,7 +47,7 @@ sub run {
   my $file_io = $self->get_filehandle($file);
 
   if ( !defined $file_io ) {
-    croak "ERROR: Can't open VGNC file $file\n";
+    croak "Can't open VGNC file $file\n";
   }
 
   my $source_name = $self->get_source_name_for_source_id($source_id, $dbi);
@@ -79,7 +79,7 @@ sub run {
   # die if some required_column is not in columns
   foreach my $colname (@required_columns) {
     if ( !grep { /$colname/xms } @columns ) {
-      croak "ERROR: Can't find required column \"$colname\" in VGNC file $file\n";
+      croak "Can't find required column $colname in VGNC file $file\n";
     }
   }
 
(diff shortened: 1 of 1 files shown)


magaliruffier:
I think those 4 xrefs were simply added to VGNC between your two runs
They parse correctly using the unchanged parser but they are not in the file I downloaded just two days ago

tgrego:
thanks @magaliruffier, I can confirm that is correct.
Rerun both version, and results are consistent at 16124 xrefs, so an extra 19 from yesterday.
Good to know the sources are updated regularly!

misc-scripts/xref_mapping/XrefParser/VGNCParser.pm
```diff
@@ -21,93 +21,109 @@ package XrefParser::VGNCParser;
...
-  my $dbi          = $ref_arg->{dbi};
-  $dbi = $self->dbi unless defined $dbi;
+  my $verbose      = $ref_arg->{verbose} // 0;
+  my $dbi          = $ref_arg->{dbi} // $self->dbi;
+
 
-  if((!defined $source_id) or (!defined $species_id) or (!defined $files) ){
-    croak "Need to pass source_id, species_id, files and rel_file as pairs";
+  if ( (!defined $source_id) or (!defined $species_id) or (!defined $files) ) {
+    croak 'Need to pass source_id, species_id and files';
```
nerdstrike: Check that croak is giving you the error you wanted.

misc-scripts/xref_mapping/XrefParser/VGNCParser.pm
```diff
@@ -21,93 +21,109 @@ package XrefParser::VGNCParser;
...
-				   species_id => $species_id} );
-
-      $self->add_synonyms_for_hgnc( {source_id  => $source_id,
-                                     name       => $acc,
-                                     species_id => $species_id,
-                                     dbi        => $dbi,
-                                     dead       => $previous_symbols,
-                                     alias      => $synonyms});
+  my $input_file = Text::CSV->new({
+    sep_char       => "\t",
```
nerdstrike: Can you turn strict on here?

misc-scripts/xref_mapping/XrefParser/VGNCParser.pm
magaliruffier: I would use double quote to be consistent with other lines

misc-scripts/xref_mapping/XrefParser/VGNCParser.pm
```diff
@@ -21,93 +21,109 @@ package XrefParser::VGNCParser;
...
+  my $file = shift @{$files};
 
   my $count = 0;
 
   my $file_io = $self->get_filehandle($file);
 
   if ( !defined $file_io ) {
-    print "ERROR: Can't open VGNC file $file\n";
-    return 1;
+    croak "ERROR: Can't open VGNC file $file\n";
```
magaliruffier: As stated in other places, maybe remove the 'ERROR' from the error message

misc-scripts/xref_mapping/XrefParser/VGNCParser.pm
```diff
@@ -21,93 +21,109 @@ package XrefParser::VGNCParser;
...
+    prev_symbol
+  );
+
+  # get header columns
+  my @columns = @{ $input_file->getline( $file_io ) };
+
+  # die if some required_column is not in columns
+  foreach my $colname (@required_columns) {
+    if ( !grep { /$colname/xms } @columns ) {
+      croak "ERROR: Can't find required column \"$colname\" in VGNC file $file\n";
```
magaliruffier: Same comment as above, 'ERROR' might be overkill
Does the $colname need to be escape quoted?

misc-scripts/xref_mapping/XrefParser/VGNCParser.pm
tgrego: no I can't...
last "column" in the source file is uniprot_ids, and that can be any number of columns (usually just one) containing a uniprot_id each.
The parser ignores the uniprot ids so it doesn't really matter (all previous fields are consistent), but turning strict on will complain about inconsistent number of fields.

nerdstrike: Eww

tgrego:
I've updated based on the reviews, let me know if more changes are required.

mkszuba:
Are you happy with the current state of this parser, @nerdstrike ?

PR #311, opened 2018-10-18
Bugfix/delint MIM parser

mkszuba:
## Description

Fixes bugs observed (so far) in MIMParser during the xref sprint, add additional checks, and delint the code to facilitate further refactoring. See ENSCORESW-2880.

## Use case

Part of the efforts to improve the xref pipeline.

## Benefits

Multi-line descriptions no longer end up with words adjacent to line breaks erronenously concatenated. Improved performance thanks to reduced amount of redundant text processing. Partly generalised xref-insertion code. Some future-proofing. Code (hopefully) easier to maintain. PerlCritic at level 3 complains only about still rather high complexity of run() and the only additional complaints introduced by level 2 are missing certain POD sections and $VERSION.

## Possible Drawbacks

None I can think of.

## Testing

No.

N/A

N/A. However, I have run the parser itself on current MIM data set and it appears to work correctly.

### DIFF
--- a/misc-scripts/xref_mapping/XrefParser/MIMParser.pm
+++ b/misc-scripts/xref_mapping/XrefParser/MIMParser.pm
@@ -255,7 +255,8 @@ sub run {
 
   if ($verbose) {
     print "$gene genemap and $phenotype phenotype MIM xrefs added\n"
-      . "added $syn_count synonyms (defined by MOVED TO)\n";
+      . "added $syn_count synonyms (defined by MOVED TO)\n"
+      . "$removed_count entries removed\n";
   }
 
   return 0;
--- a/misc-scripts/xref_mapping/XrefParser/MIMParser.pm
+++ b/misc-scripts/xref_mapping/XrefParser/MIMParser.pm
@@ -194,12 +194,10 @@ sub run {
                              info_type  => "DEPENDENT" } );
         }
         elsif ( $type eq q{^} ) {
-          if ( $_ =~ m{
-                        [*]FIELD[*]\sTI\n
-                        \N{CARET}
-                        \d+\sMOVED\sTO\s
-                        (\d+)
-                    }msx ) {
+          if ( $long_desc =~ m{
+                                MOVED\sTO\s
+                                (\d+)
+                            }msx ) {
             if ( $1 eq $number ) { next; }
             $old_to_new{$number} = $1;
           }
--- a/misc-scripts/xref_mapping/XrefParser/MIMParser.pm
+++ b/misc-scripts/xref_mapping/XrefParser/MIMParser.pm
@@ -81,7 +81,6 @@ sub run {
 
   my %old_to_new;
   my %removed;
-  my $source_id;
   my @sources;
 
   push @sources, $general_source_id;
@@ -113,13 +112,11 @@ sub run {
 
   while ( $_ = $mim_io->getline() ) {
     #get the MIM number
-    my $is_morbid = 0;
     my ( $number ) = ( $_ =~ m{
                                 [*]FIELD[*]\s+NO\n
                                 (\d+)
                             }msx );
     if ( defined $number ) {
-      $source_id = $gene_source_id;
 
       my ( $ti ) = ( $_ =~ m{
                               [*]FIELD[*]\sTI\n
--- a/misc-scripts/xref_mapping/XrefParser/MIMParser.pm
+++ b/misc-scripts/xref_mapping/XrefParser/MIMParser.pm
@@ -64,7 +64,7 @@ sub run {
 
   print "sources are:- " . join( ", ", @sources ) . "\n" if ($verbose);
 
-  local $/ = "*RECORD*";
+  IO::Handle->input_record_separator('*RECORD*');
 
   my $mim_io = $self->get_filehandle($file);
 
--- a/misc-scripts/xref_mapping/XrefParser/MIMParser.pm
+++ b/misc-scripts/xref_mapping/XrefParser/MIMParser.pm
@@ -195,12 +195,41 @@ sub run {
 
   $mim_io->close();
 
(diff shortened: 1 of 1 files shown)


misc-scripts/xref_mapping/XrefParser/MIMParser.pm
```diff
@@ -21,162 +21,245 @@ package XrefParser::MIMParser;
...
+use Readonly;
+use POSIX qw(strftime);
+
+use parent qw( XrefParser::BaseParser );
+
+# This parser will read xrefs from a record file downloaded from the
+# OMIM Web site. They should be assigned to two different xref
+# sources: MIM_GENE and MIM_MORBID. MIM xrefs are linked to EntrezGene
+# entries so the parser does not match them to Ensembl; this will be
+# taken care of when EntrezGene antries are matched.
```
nerdstrike:
```suggestion
# taken care of when EntrezGene entries are matched.
```

misc-scripts/xref_mapping/XrefParser/MIMParser.pm
```diff
@@ -21,162 +21,245 @@ package XrefParser::MIMParser;
...
+  my $files             = $ref_arg->{files};
+  my $verbose           = $ref_arg->{verbose};
+  my $dbi               = $ref_arg->{dbi};
 
-  if((!defined $general_source_id) or (!defined $species_id) or (!defined $files) ){
+  if ( ( !defined $general_source_id ) or
+       ( !defined $species_id ) or
+       ( !defined $files ) )
+  {
     croak "Need to pass source_id, species_id and files as pairs";
```
nerdstrike: I don't think croak is giving us the errors we desire. For me it was "Error in run() in file", rather than "Error on line n"

misc-scripts/xref_mapping/XrefParser/MIMParser.pm
```diff
@@ -21,162 +21,245 @@ package XrefParser::MIMParser;
...
 
-  my $syn_count =0;
-  foreach my $mim (keys %old_to_new){
-    my $old= $mim;
-    my $new= $old_to_new{$old};
-    while(defined($old_to_new{$new})){
+  foreach my $mim ( keys %old_to_new ) {
+    my $old = $mim;
+    my $new = $old_to_new{$old};
+    while ( defined( $old_to_new{$new} ) ) {
```
nerdstrike: This seems like a strange sort of while loop

misc-scripts/xref_mapping/XrefParser/MIMParser.pm
mkszuba: Fixed, thanks.

misc-scripts/xref_mapping/XrefParser/MIMParser.pm
mkszuba: Hmm, all the four parsers I have been assigned handle this particular bit identically _i.e._ with croak()... Would you have any suggestions regarding how to improve this? By the way, this test never gets triggered if _standaloneJob.pl_ is used, one gets _e.g._

` WORKER_ERROR : ParamError: value for param_required('source') is required and has to be defined`

instead.

magaliruffier: The ParseSource runnable checks for required parameters, see https://github.com/Ensembl/ensembl-production/blob/release/94/modules/Bio/EnsEMBL/Production/Pipeline/Xrefs/ParseSource.pm#L30
It is probably good to leave the croak for backwards compatibility with the xref_parser.pl script, but also comforting to know the eHive pipeline does a better job at checking for input

misc-scripts/xref_mapping/XrefParser/MIMParser.pm
mkszuba: Having had a very careful look at this, it seems to work what it is supposed to do - which is making all synonyms point to their respective final entries regardless of how many time an entry might have been moved. Not entirely optimal performance-wise but overall, I am impressed! I have just added extensive comments to that bit of code, hopefully all will be clear now.

misc-scripts/xref_mapping/XrefParser/MIMParser.pm
```diff
@@ -21,162 +21,274 @@ package XrefParser::MIMParser;
 
 use strict;
 use warnings;
+
 use Carp;
-use POSIX qw(strftime);
 use File::Basename;
```
tgrego: File::Basename can probably be safely removed

misc-scripts/xref_mapping/XrefParser/MIMParser.pm
```diff
@@ -21,162 +21,274 @@ package XrefParser::MIMParser;
...
+       ( !defined $species_id ) or
+       ( !defined $files ) )
+  {
     croak "Need to pass source_id, species_id and files as pairs";
   }
-  $verbose |=0;
-
-  my $file = @{$files}[0];
+  $verbose //= 0;
+  $dbi //= $self->dbi;
```
tgrego: definedness check and default values can be done at the same time of the assignment, just to save a couple of lines.

misc-scripts/xref_mapping/XrefParser/MIMParser.pm
mkszuba: Done...

misc-scripts/xref_mapping/XrefParser/MIMParser.pm
mkszuba: ...and done.

misc-scripts/xref_mapping/XrefParser/MIMParser.pm
mkszuba: OK

tgrego:
@nerdstrike are you happy with the updates?

PR #312, opened 2018-10-18
ZFINDesc parser update

tgrego:
## Description

Update to the ZFINDesc parser during the first stage of the xref sprint. See ENSCORESW-2877.
Updates were as agreed... no direct file handling, croak on error, null instead of '', linting in general...
Parser now using Text::CSV.

## Testing

no unit tests.
Manually tested with zebrafish.
Results were consistent.

### DIFF
--- a/misc-scripts/xref_mapping/XrefParser/ZFINDescParser.pm
+++ b/misc-scripts/xref_mapping/XrefParser/ZFINDescParser.pm
@@ -36,7 +36,7 @@ sub run {
   my $verbose      = $ref_arg->{verbose} // 0;
 
   if ( (!defined $source_id) or (!defined $species_id) or (!defined $files) ) {
-    croak 'Need to pass source_id, species_id and files';
+    croak "Need to pass source_id, species_id and files as pairs";
   }
 
   my $file = shift @{$files};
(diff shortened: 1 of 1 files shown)


misc-scripts/xref_mapping/XrefParser/ZFINDescParser.pm
```diff
@@ -22,58 +22,73 @@ package XrefParser::ZFINDescParser;
...
   my $source_id    = $ref_arg->{source_id};
   my $species_id   = $ref_arg->{species_id};
   my $files        = $ref_arg->{files};
-  my $verbose      = $ref_arg->{verbose};
+  my $verbose      = $ref_arg->{verbose} // 0;
 
-  if((!defined $source_id) or (!defined $species_id) or (!defined $files) ){
-    croak "Need to pass source_id, species_id and files as pairs";
+  if ( (!defined $source_id) or (!defined $species_id) or (!defined $files) ) {
+    croak 'Need to pass source_id, species_id and files';
```
magaliruffier: why was the double quote replaced by a single quote?
double quotes are usually safer and it is used that way further down the code (line 56 for example)

tgrego: no good reason really, it's only because there is no variable interpolation happening and I got the habit of double quoting only when I need to...
will change it back.

PR #313, opened 2018-10-19
Feature/jgi

nerdstrike:
## Description

Clean up of JGI parser, removing redundant version 2 content, and creating fatal errors instead of warnings

## Use case

Ciona xrefs

## Benefits

Errors will not be hidden, code is a little simpler

## Possible Drawbacks

Errors will not be hidden.

## Testing

No unit tests. Produces the same output statistics

### DIFF
--- a/misc-scripts/xref_mapping/XrefParser/JGI_Parser.pm
+++ b/misc-scripts/xref_mapping/XrefParser/JGI_Parser.pm
@@ -28,7 +28,6 @@ use base qw( XrefParser::BaseParser );
 
 # JGI protein file with gene predictons  - FASTA FORMAT
 #
-#
 # This is the parser that provides most functionality, subclasses
 # (JGI_ProteinParser) just set sequence type)
 
@@ -80,8 +79,7 @@ sub run {
   my $file_io = $self->get_filehandle($file);
 
   if ( !defined $file_io ) {
-    print STDERR "ERROR: Could not open $file\n";
-    return 1;    # 1 is an error
+    croak "ERROR: Could not open $file\n";
   }
 
   $file_io->input_record_separator( "\n>");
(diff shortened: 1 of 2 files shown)


misc-scripts/xref_mapping/XrefParser/JGI_Parser.pm
```diff
@@ -76,89 +74,60 @@ sub run {
...
-#    }
-#
+
     push @xrefs, $xref;
 
   }
 
   $file_io->close();
-
+  IO::Handle->input_record_separator( "\n");
```
avullo: This is probably not needed, but it's good anyway to explicitly reset to default.

misc-scripts/xref_mapping/XrefParser/JGI_Parser.pm
```diff
@@ -76,89 +74,60 @@ sub run {
...
-      print STDERR "WARNING : The source-name specified ($source_name) in the populate_metatable.sql file is\n" .
-        "WARNING : not matching the differnt cases specified in JGI_Parser.pm - plese\n".
-          "WARNING : edit the parser \n" ; 
-      return 1;
-    } 
-    #print "ACCESSION $acession\n" ;  
+      # JGI 1.0
+
+    } else {
+      croak "WARNING : The source-name specified ($source_name) in the populate_metatable.sql file is\n" .
```
avullo: I find it's a little bit misleading to emit a warning when dying: I'd rather say: "ERROR"

avullo: I find it's a little bit misleading to emit a warning when dying: I'd rather say: "ERROR"

misc-scripts/xref_mapping/XrefParser/JGI_Parser.pm
avullo: This is probably not needed, but good anyway to explicitly reset to default.

misc-scripts/xref_mapping/XrefParser/JGI_Parser.pm
```diff
@@ -38,7 +37,6 @@ sub run {
   my $source_id    = $ref_arg->{source_id};
   my $species_id   = $ref_arg->{species_id};
   my $files        = $ref_arg->{files};
-  my $release_file = $ref_arg->{rel_file};
   my $verbose      = $ref_arg->{verbose};
   my $dbi          = $ref_arg->{dbi};
   $dbi = $self->dbi unless defined $dbi;
```
premanand17: Two lines could be merged to one
my $dbi = $ref_arg->{dbi} || $self->dbi;

misc-scripts/xref_mapping/XrefParser/JGI_Parser.pm
```diff
@@ -76,89 +74,60 @@ sub run {
...
-      # we want  146277
-      ($acession = $header)  =~s/\w{6}//;  
-      $version = "JGI 1.0" ;  
-    
-    } elsif ($source_name=~m/cint_aniseed_.*v1/) { 
+      # JGI 1.0
+      # we want accession 146277 from above
+      ($accession = $header) =~s/\w{6}//;
+
+    } elsif ($source_name=~m/cint_aniseed_.*v1/) {
```
premanand17: What is this  block ( elsif ($source_name=~m/cint_aniseed_.*v1/) { )  doing?!!

It's an empty block?!!  Do we need this?!

misc-scripts/xref_mapping/XrefParser/JGI_Parser.pm
nerdstrike: I'm slightly nervous about a worker running another parser after this one and remembering the IO::Handle package state.

misc-scripts/xref_mapping/XrefParser/JGI_Parser.pm
nerdstrike: Funnily enough, that's how this always was. I'll see what happens if I run it with the same regex as the previous block.

PR #316, opened 2018-10-22
Bugfix/delint Mim2Gene parser

mkszuba:
**Warning:** as of 2018-10-30, this PR is expected to fail Travis builds due to being dependent on #323 and #324 . 

## Description

Fixes bugs observed (so far) in Mim2GeneParser during the xref sprint, implements support for direct MIM xrefs, add additional checks, and delint the code to facilitate further refactoring. See ENSCORESW-2891.

## Use case

Part of the efforts to improve the xref pipeline. Moreover, one of the observed bugs actually prevents current versions of mim2gene data from being parsed at all.

## Benefits

The parser can now handle recent versions of mim2gene input. Direct xrefs are now produced wherever possible, with dependent ones only used for entries lacking Ensembl ID but with EntrezGene ID present. Use BaseParser methods for inserting dependent xrefs into the database, which in addition to avoiding hand-rolled DBI code will, once pull request #314 has been approved, prevent Mim2GeneParser from inserting duplicate entries upon re-runs with the same input. Some future-proofing. Code (hopefully) easier to maintain. Most complaints of PerlCritic levels 3 and 2 taken care of. Use a standardised rather than hand-rolled CSV parser, with potential for a performance increase if compiled rather than native-Perl version of the parser is used.

## Possible Drawbacks

Output is less straightforward than it used to be because it now includes both direct (the vast majority) and dependent (around 10 percent as of today) xrefs.

## Testing

No.

N/A

N/A. However, I have run the parser itself on both current DBASS data and some intentionally malformed input, and it appears to work correctly.

### DIFF
--- a/misc-scripts/xref_mapping/XrefParser/Mim2GeneParser.pm
+++ b/misc-scripts/xref_mapping/XrefParser/Mim2GeneParser.pm
@@ -172,6 +172,7 @@ sub run {
                                        'source_id'      => $mim_gene_source_id,
                                        'species_id'     => $species_id,
                                        'linkage'        => $entrez_source_id,
+                                       'dbi'            => $dbi,
                                      });
           }
         }
@@ -186,6 +187,7 @@ sub run {
                                        'source_id'      => $mim_morbid_source_id,
                                        'species_id'     => $species_id,
                                        'linkage'        => $entrez_source_id,
+                                       'dbi'            => $dbi,
                                      });
           }
         }
@@ -201,6 +203,7 @@ sub run {
                                        'source_id'      => $mim_morbid_source_id,
                                        'species_id'     => $species_id,
                                        'linkage'        => $entrez_source_id,
+                                       'dbi'            => $dbi,
                                      });
           }
         }
@@ -215,6 +218,7 @@ sub run {
                                        'source_id'      => $mim_gene_source_id,
                                        'species_id'     => $species_id,
                                        'linkage'        => $entrez_source_id,
+                                       'dbi'            => $dbi,
                                      });
           }
         }
--- a/misc-scripts/xref_mapping/XrefParser/Mim2GeneParser.pm
+++ b/misc-scripts/xref_mapping/XrefParser/Mim2GeneParser.pm
@@ -27,6 +27,7 @@ use warnings;
 
 use Carp;
 use File::Basename;
+use List::Util;
 use POSIX qw(strftime);
 use Readonly;
 use Text::CSV;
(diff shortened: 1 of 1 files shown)


misc-scripts/xref_mapping/XrefParser/Mim2GeneParser.pm
```diff
@@ -19,120 +19,281 @@ limitations under the License.
...
+      }
+      next RECORD;
     }
-    
-    if((!defined $mim_gene{$omim_id} ) and (!defined $mim_morbid{$omim_id} ) ){
-      $missed_omim++;
-      next;
+
+    if ( scalar @{ $line } != $EXPECTED_NUMBER_OF_COLUMNS ) {
+      croak ' Line ' . $csv->record_number()
```
nerdstrike: I'd have used sprintf here, but ymmv.

misc-scripts/xref_mapping/XrefParser/Mim2GeneParser.pm
```diff
@@ -19,120 +19,281 @@ limitations under the License.
...
-	  foreach my $mim_id (@{$mim_gene{$omim_id}}){
-	    $add_dependent_xref_sth->execute($ent_id, $mim_id);
-	  }
-	}
-	#	$add_dependent_xref_sth->execute($entrez{$entrez_id}, $mim_gene{$omim_id});	
-      }
+
+    # ...or no Ensembl ID or EntrezGene xref to match it to
+    # FIXME: this number might be underestimated because it doesn't
+    # check if Ensembl IDs from the input file actually exist
```
nerdstrike: And indeed, some of them may not, because of the separation between Ensembl releases (and stable ID updates) and OMIM updates.

misc-scripts/xref_mapping/XrefParser/Mim2GeneParser.pm
```diff
@@ -19,120 +19,281 @@ limitations under the License.
...
+    # An unknown type might indicate the change of input format,
+    # therefore make sure the user notices it. That said, do not
+    # bother we do not have an xref this entry would operate on anyway
+    # - which is why we only check this after the preceding two
+    # presence checks.
+    if ( ( $type ne 'gene')
+         && ( $type ne 'gene/phenotype' )
+         && ( $type ne 'predominantly phenotypes' )
+         && ( $type ne 'phenotype' ) ) {
+      croak "Unknown type $type";
```
nerdstrike: Could this error include a line/record number, or just the MIM ID? It would help with debugging.

mkszuba: ...or both, because why not.

mkszuba:
No changes since Kieron's review,  I have only rebased this branch against the current state of _feature/xref_sprint_.

PR #318, opened 2018-10-24
Standardise table.sql file

james-monkeyshines:
## Description and Use case
The new datacheck for validating a db schema uses the table.sql file.
This is mostly consistent already, both within the file and between instances of the files in different repositories. But there are a few simple changes to terminology and formatting that would further standardise the files. These sorts of things can be handled in the datacheck code, but having fewer, and less complicated, exceptions makes the code more robust and easier to understand.

## Benefits
Datacheck is simpler. Standardisation and consistency are generally good things to aim for.

## Possible Drawbacks
None I can think of.

## Testing
Unit tests not relevant.
I created databases before and after the changes, and the results were identical.

### DIFF
--- a/sql/table.sql
+++ b/sql/table.sql
@@ -177,9 +177,9 @@ CREATE TABLE data_file (
   file_type         ENUM('BAM','BAMCOV','BIGBED','BIGWIG','VCF'),
 
   PRIMARY KEY (data_file_id),
-  UNIQUE KEY df_unq_idx(coord_system_id, analysis_id, name, file_type),
-  KEY df_name_idx(name),
-  KEY df_analysis_idx(analysis_id)
+  UNIQUE KEY df_unq_idx (coord_system_id, analysis_id, name, file_type),
+  KEY df_name_idx (name),
+  KEY df_analysis_idx (analysis_id)
 ) COLLATE=latin1_swedish_ci ENGINE=MyISAM;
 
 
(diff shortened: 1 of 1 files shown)


james-monkeyshines:
I don't think I could write a patch file for these changes; ignoring the whitespace ones, they amount to replacing MySQL synonyms with the 'canonical' keywords. But when you load a database, that replacement happens automatically; so there is nothing to change in an existing database.

magaliruffier:
How do we update all the table.sql files living in modules/t/test-genome-DBs then? For example
https://github.com/Ensembl/ensembl/blob/release/94/modules/t/test-genome-DBs/homo_sapiens/core/table.sql

james-monkeyshines:
There's no need to update those files; they've already been through MySQL and out the other side, so the synonym mapping has already happened there. E.g. in the example you give, there are no BOOLEANs, just TINYINT(1)s. So those files are already consistent with my proposal.

magaliruffier:
Fair enough

james-monkeyshines:
I'm not making changes for the sake of brevity, but to use the standard terms rather than synonyms. I guess it's a matter of preference, but I do not like using BOOLEAN in MySQL; because that's not what it is, it's an integer in the range [-128, 127].

The "BOOLEAN" is lost as soon as you create a database, i.e. doing a 'show create table' or mysqldump will return "TINYINT(1)". So it would only be visible for people seeking to understand the schema by reading this file (or anything derived from this file), and the column names in these cases make the nature of the data clear, I think.

Note also that this is not just about standardising within this file; I've also been looking at funcgen, variation and production table.sql files, and they already all use TINYINT(1) in preference to BOOLEAN.

tgrego:
indeed mysql does not have a true boolean type and uses TINYINT(1) to hold that type of data, and defines the alias to use it. the alias exists only to make people's life easier, of course the engine does not care and uses the internal real data type.
But tables.sql is written by people, and I believe less people will know instantly what a TINYINT(1) means is than a BOOLEAN.
But I agree is just a matter of preference and not really important.

PR #322, opened 2018-10-30
Remove mgi xref mappings via ccds and uniprot 

premanand17:
## Requirements

## Description

Removed the sections from the ini file corresponding to MGI ccds and uniprot sources.
Removed the call to get the source id from 'MGI' source name and uniprot as it is not applicable anymore.

## Use case

For mouse, we use MGI symbols as the official naming source. To increase the coverage, we use third-party mappings to MGI, including CCDS and UniProt mappings. As of the latest update, we have 14 MGI xrefs obtained via CCDS mappings and another 52 via UniProt. These updates will  remove those additional mappings to get the mappings from a single source ie MGI.

## Benefits

We get the mappings only from single source (MGI)

## Possible Drawbacks

We may miss the mappings for the some minimal ids. (14-ccds, 52-uniprot)

## Testing

Tested it by running the xref_parser script and checked the _xref database for row counts. 
modules/t/xref_parser.t also passes

### DIFF
--- a/misc-scripts/xref_mapping/XrefParser/UniProtParser.pm
+++ b/misc-scripts/xref_mapping/XrefParser/UniProtParser.pm
@@ -141,45 +141,6 @@ sub create_xrefs {
   my @tax_ids = @{$species2tax{$species_id}};
   my %taxonomy2species_id = map{ $_=>$species_id } @tax_ids;
 
-
-#
-# MGI data needed---------------------------------------------------------
-#
-  my %mgi_acc_to_desc;
-  my %mgi_acc_to_label;
-  my %mgi_label_to_desc;
-  my %mgi_label_to_acc;
-
-  my $sth = $dbi->prepare("SELECT x.accession, x.label, x.description from xref x, source s where x.source_id = s.source_id and s.name like 'MGI' and s.priority_description like 'descriptions'");
-  
-  $sth->execute() or croak( $dbi->errstr() );
-  while ( my @row = $sth->fetchrow_array() ) {
-    $mgi_acc_to_desc{$row[0]}   = $row[2];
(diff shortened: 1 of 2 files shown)


misc-scripts/xref_mapping/XrefParser/UniProtParser.pm
```diff
@@ -123,10 +123,6 @@ sub create_xrefs {
 
   my %dependent_sources = $self->get_xref_sources($dbi);
 
-  if(defined($dependent_sources{'MGI'})){
-    $dependent_sources{'MGI'} = $self->get_source_id_for_source_name("MGI","uniprot", $dbi);
-  }
-
```
mkszuba: Is this really everything that we need to remove from UniProtParser? There are relatively sizeable chunks of MGI-related code elsewhere in it, moreover it seems that we might need something like _if ( $source=~ "MGI" ) { next; }_ around line 450 in order to truly stop processing this source.

premanand17: This along with removing the source from theini file was enough to prevent the population of the mgi-uniprot xrefs. You are right, have to do a proper removal of untouched code. Will update.

magaliruffier: The UniProtParser is strange in that it will add any new xrefs as long as it can find a matching source for it. As we are explicitly asking for the MGI source of priority_description 'uniprot', this is normally enough to stop it adding the xrefs automatically if the source is not in the xref_config.ini file
However, I agree that it would be safer to add an explicit ignore MGI in the UniProtParser.
In the past, we would suddenly end up adding FlyBase xrefs to human because they exist in UniProt and we added a source Flybase in the config file.

PR #324, opened 2018-10-30
BaseParser: fast path for insertion of dependent xrefs

mkszuba:
**Warning:** this PR is now dependent on #326

## Description

Factor insertion of _dependent_xref_ entries out of add_dependent_xref() so that it can be used independently.

## Use case

Parsers inserting dependent-xref mappings which do not insert xrefs themselves, e.g. Mim2GeneParser.

## Benefits

Parsers which are not supposed to insert xrefs themselves will now no longer insert stub xrefs upon e.g. encountering a typo. In the event of a single accession pointing to multiple xref_ids dependent mappings can now be correctly assigned to those IDs, whereas using add_dependent_xref() would assign them all to the first matching xref_id returned by the database. Considerable performance gain thanks to fewer conversations with the database. API behaviour for inserting dependent xrefs is now consistent with that for inserting direct ones, albeit due to a name collision functions doing the same thing have different names for the two.

## Possible Drawbacks

When the fast path is used validation of xref IDs is left up to the programmer, therefore it may now be easier to insert mappings pointing to xrefs which do not exist.

## Testing

No

N/A

N/A but I have tested the changes by running a parser which uses the relevant bits of code (i.e. my delinted version of Mim2GeneParser) and the changes seem to do what they should.

### DIFF
--- a/misc-scripts/xref_mapping/XrefParser/BaseParser.pm
+++ b/misc-scripts/xref_mapping/XrefParser/BaseParser.pm
@@ -1024,6 +1024,7 @@ sub add_identity_xref {
 
 ###################################################################
 # Create new xref if needed and add as a direct xref to a stable_id
+# Note that a corresponding method for dependent xrefs is called add_dependent_xref()
 ###################################################################
 sub add_to_direct_xrefs{
   my ($self, $arg_ref) = @_;
@@ -1073,6 +1074,7 @@ AXX
 ##################################################################
 # Add a single record to the direct_xref table.
 # Note that an xref must already have been added to the xref table
+# Note that a corresponding method for dependent xrefs is called add_dependent_xref_maponly()
 ##################################################################
 sub add_direct_xref {
   my ($self, $general_xref_id, $ensembl_stable_id, $ensembl_type, $linkage_type, $dbi) = @_;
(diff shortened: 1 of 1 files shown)


misc-scripts/xref_mapping/XrefParser/BaseParser.pm
```diff
@@ -1133,30 +1130,66 @@ ADX
...
+}
+
+
+##################################################################
+# Add a single record to the dependent_xref table.
+# Note that an xref must already have been added to the xref table
+# Note that a corresponding method for direct xrefs is called add_direct_xref()
+##################################################################
+
+sub add_dependent_xref_maponly {
```
tgrego: Naming is a bit strange, all of it...
add_direct_xref links in function to add_dependent_xref_maponly
add_to_direct_xrefs links in function to add_dependent_xref

I'm sure we could make it more clear, maybe in the design refactoring.

ens-bwalts: Agree - add_dependent_xref() would be more meaningfully named add_xref_and_dependency(), and add_dependent_xref_maponly() maybe set_dependency_for_xref()

but calls to add_dependent_xref() are already everywhere in the parsers...

mkszuba: Indeed, the naming is very much due for refactoring.

PR #325, opened 2018-10-31
Xref parser testing framework

nerdstrike:
## Description

Adds a single unit test for a parser, and a database test harness to facilitate that. Small opening in BaseParser made to remove tight dependency on XrefParser::Database

## Use case

Testing parsers is hard. They require a database handle as argument, and the testable output appears directly in the database.

## Benefits

A test harness without needing to use Bio::EnsEMBL::Test. ORM functionality should make query-based testing easier, and allow schema upgrades to occur without undue suffering by way of DBIC.

## Possible Drawbacks

DBIC dependencies, foreign key conflicts in constraint-free xref code, no Travis harness, but we should be adding that to the new repo instead of this one. Generated ORM is now diverged from original schema - there are changes that are required for DBIC to work with the schema.

## Testing

I wrote the only tests that exist. No Travis yet.


nerdstrike:
I should add the ORM schema is incomplete. We could probably add more relationships, but I don't understand the original schema enough to follow through.

PR #329, opened 2018-11-02
RefSeqCoordinateParser update

tgrego:
## Description

Update to the RefSeqCoordinateParser as part of the efforts of the xref sprint. See ENSCORESW-2894.
Many updates however logic was mostly kept.
There is logic that is likely to be reused by other parsers, so might move in a future pull request.

## Testing

No unit tests.
Tested with rat and results look consistent.


misc-scripts/xref_mapping/XrefParser/RefSeqCoordinateParser.pm
```diff
@@ -23,405 +23,300 @@ use strict;
...
+    RefSeq_peptide
+    RefSeq_mRNA
+    RefSeq_ncRNA
+    RefSeq_peptide_predicted
+    RefSeq_mRNA_predicted
+    RefSeq_ncRNA_predicted
+  ) );
+
+  my $source_ids; 
+  map { $source_ids->{$_} = $self->get_source_id_for_source_name( $_, 'otherfeatures', $dbi ) } @source_names;
```
nerdstrike: Damian Conway frowns on maps that do not return anything.

misc-scripts/xref_mapping/XrefParser/RefSeqCoordinateParser.pm
```diff
@@ -23,405 +23,300 @@ use strict;
...
-  } else {
-      die("Missing or unsupported project value. Supported values: ensembl, ensemblgenomes");
   }
 
-## Not all species have an otherfeatures database, skip if not found
+  # Not all species have an otherfeatures database, skip if not found
   if (!$otherf_dba) {
-    print STDERR "No otherfeatures database for $species_name, skipping import for refseq_import data\n";
-    return;
+    croak "No otherfeatures database for $species_name, skipping import for refseq_import data\n";
```
nerdstrike: Technically this is not skipping.

misc-scripts/xref_mapping/XrefParser/RefSeqCoordinateParser.pm
```diff
@@ -23,405 +23,300 @@ use strict;
...
 use DBI;
-
-use base qw( XrefParser::BaseParser );
+use Readonly;
 use Bio::EnsEMBL::Registry;
 
-sub run_script {
+use parent qw( XrefParser::BaseParser );
+
+# set the score thresholds
```
nerdstrike: Can we get something a bit more descriptive here? What is the penalty for? Why is there a threshold?

misc-scripts/xref_mapping/XrefParser/RefSeqCoordinateParser.pm
```diff
@@ -23,405 +23,300 @@ use strict;
...
+  # Not all species have an otherfeatures database, skip if not found
   if (!$otherf_dba) {
-    print STDERR "No otherfeatures database for $species_name, skipping import for refseq_import data\n";
-    return;
+    croak "No otherfeatures database for $species_name, skipping import for refseq_import data\n";
   }
 
-## Add link to EntrezGene IDs where available
-  my (%entrez_ids) = %{ $self->get_valid_codes("EntrezGene", $species_id, $dbi) };
+  # Add link to EntrezGene IDs where available
```
nerdstrike: This code does not add EntrezGene IDs. It only caches them.

misc-scripts/xref_mapping/XrefParser/RefSeqCoordinateParser.pm
```diff
@@ -23,405 +23,300 @@ use strict;
...
-    }
-  }
-## Not all species have refseq_import data, skip if not found
-  if (!defined $logic_name) {
-    print STDERR "No data found for RefSeq_import, skipping import\n";;
-    return;
+
+  # Not all species have refseq_import data, skip if not found
+  if (!defined $aa_of->fetch_by_logic_name('refseq_import')->logic_name) {
+    croak "No data found for RefSeq_import, skipping import\n";
```
nerdstrike: This doesn't skip anything either.

misc-scripts/xref_mapping/XrefParser/RefSeqCoordinateParser.pm
```diff
@@ -23,405 +23,300 @@ use strict;
...
   }
 
-  foreach my $chromosome_of (@$chromosomes_of) {
+  # start main loop
```
nerdstrike: Would you annotate this "Iterate over chromosomes in otherfeatures database" or something similar? It may not be obvious what '_of' stands for.

misc-scripts/xref_mapping/XrefParser/RefSeqCoordinateParser.pm
```diff
@@ -23,405 +23,300 @@ use strict;
...
-	}
-        if ($id !~ /^[NXMR]{2}_[0-9]+/)  { next; }
-        my %transcript_result;
-        my %tl_transcript_result;
-        if (!defined $id) { next; }
+      # Create a range registry for all the exons of the refseq transcript
+      foreach my $transcript_of (sort { $a->start <=> $b->start } @{$transcripts_of}) {
+        my $id;
+        # We're moving to RefSeq accessions being stored as xrefs rather than
+        # stable ids. But we also need to maintain backwards compatbility.
```
nerdstrike: Sentences should not generally begin with 'But'. In general this comment suggests you understand what is going on, but it could be made clearer to another reader. What 'new kind' do you refer to here?

misc-scripts/xref_mapping/XrefParser/RefSeqCoordinateParser.pm
```diff
@@ -23,405 +23,300 @@ use strict;
...
         my $rr1 = Bio::EnsEMBL::Mapper::RangeRegistry->new();
         my $tl_exons_of = $transcript_of->get_all_translateable_Exons();
         my $rr3 = Bio::EnsEMBL::Mapper::RangeRegistry->new();
 
-        foreach my $exon_of (@$exons_of) {
-          my $start_of = $exon_of->seq_region_start();
-          my $end_of = $exon_of->seq_region_end();
-          $rr1->check_and_register( 'exon', $start_of, $end_of );
-        }
+        # register $exons_of on $rr1
```
nerdstrike: Perhaps rr1,2,3,4 and could be named more functionally, now you've cleaned up the mess? I reckon a longer variable name could be forgiven here where there is so much room for confusion

misc-scripts/xref_mapping/XrefParser/RefSeqCoordinateParser.pm
```diff
@@ -23,405 +23,300 @@ use strict;
...
-                                          species_id => $species_id,
-                                          dbi => $dbi,
-                                          info_type => 'DIRECT' });
-          $self->add_direct_xref($xref_id, $best_id, "Transcript", "", $dbi);
-
-	  my $t_of = $transcript_of;
-          my $g_of = $t_of->get_Gene();
-          my $entrez_id = $g_of->stable_id();
-          my $tl_of = $t_of->translation();
+          my ($acc, $version) = split(/\./xms, $id);
```
nerdstrike: I think perhaps observing the fully annotate regex options for 'split on period' is going overboard.

misc-scripts/xref_mapping/XrefParser/RefSeqCoordinateParser.pm
```diff
@@ -431,4 +326,107 @@ sub run_script {
...
+  my $best_tl_score = 0;
+  my $best_id;
+
+  # Comparing the scores based on coding exon overlap
+  # If there is a stale mate, chose best exon overlap score
+  foreach my $tid (sort { $transcript_result->{$b} <=> $transcript_result->{$a} } keys(%{$transcript_result})) {
+    my $score = $transcript_result->{$tid};
+    my $tl_score = $tl_transcript_result->{$tid};
+    if ($score > $TRANSCRIPT_SCORE_THRESHOLD || $tl_score > $TL_TRANSCRIPT_SCORE_THRESHOLD) {
+      if ($tl_score >= $best_tl_score) {
```
nerdstrike: This code looks familiar - The outer if block is redundant with the if/elsif within.

misc-scripts/xref_mapping/XrefParser/RefSeqCoordinateParser.pm
```diff
@@ -431,4 +326,107 @@ sub run_script {
...
+  return ($best_id, $best_score, $best_tl_score);
+}
+
+# gets the source id for a given refseq acc.
+# requires as args the refseq acc and a hash mapping source names to ids
+# Needs to be moved elsewhere because other parsers benefit from using it
+sub source_id_for_acc {
+  my ($self, $acc, $source_ids) = @_;
+
+  my $source = {
```
nerdstrike: I'm not at all convinced by this hash creation and disposal. You could at least stop passing source_ids around and stick them in the instance. Barring a print statement, source_ids is only used here. Maybe you could burn the RefSeq prefixes into the source_id cache directly (and earlier), rather than in every ID mapping?

misc-scripts/xref_mapping/XrefParser/RefSeqCoordinateParser.pm
```diff
@@ -431,4 +326,107 @@ sub run_script {
...
+    NM => $source_ids->{'RefSeq_mRNA'},
+    NR => $source_ids->{'RefSeq_ncRNA'},
+    XM => $source_ids->{'RefSeq_mRNA_predicted'},
+    XR => $source_ids->{'RefSeq_ncRNA_predicted'},
+    NP => $source_ids->{'RefSeq_peptide'},
+    XP => $source_ids->{'RefSeq_peptide_predicted'}
+  };
+
+  my ($prefix) = $acc =~ /\A(\w\w)_/xms;
```
nerdstrike: substr _would_ be faster...

tgrego:
@nerdstrike thanks for the comments, hopefully was able to address everything

misc-scripts/xref_mapping/XrefParser/RefSeqCoordinateParser.pm
```diff
@@ -23,405 +23,302 @@ use strict;
...
-    }
-  }
-## Not all species have refseq_import data, skip if not found
-  if (!defined $logic_name) {
-    print STDERR "No data found for RefSeq_import, skipping import\n";;
-    return;
+
+  # Not all species have refseq_import data, exit if not found
+  if (!defined $aa_of->fetch_by_logic_name('refseq_import')->logic_name) {
+    croak "No data found for RefSeq_import\n";
```
magaliruffier: I don't think we want the code to fail here if the analysis is not found.
There are species which will have an otherfeatures database without having the refseq_import data. In those cases, we want to skip the rest of the parsers but not fail the whole pipeline.

misc-scripts/xref_mapping/XrefParser/RefSeqCoordinateParser.pm
```diff
@@ -23,405 +23,302 @@ use strict;
...
-  } else {
-      die("Missing or unsupported project value. Supported values: ensembl, ensemblgenomes");
   }
 
-## Not all species have an otherfeatures database, skip if not found
+  # Not all species have an otherfeatures database, error if not found
   if (!$otherf_dba) {
-    print STDERR "No otherfeatures database for $species_name, skipping import for refseq_import data\n";
-    return;
+    croak "No otherfeatures database found for $species_name\n";
```
magaliruffier: Similar comment here as for a missing refseq_import analysis
Some species will not have an otherfeatures database, in which case this parser should not be run but the rest of the pipeline should run without failing.
In the eHive pipeline, the parsing job should not be created if no otherfeatures database is available, so hopefully we would not hit this case. If we want to maintain backwards compatibility with the xref_parser.pl script though, we need to allow the parser to exit without error if no otherfeatures database is found.

misc-scripts/xref_mapping/XrefParser/RefSeqCoordinateParser.pm
```diff
@@ -23,405 +23,302 @@ use strict;
...
-          if (defined $entrez_ids{$entrez_id} ) {
-            foreach my $dependent_xref_id (@{$entrez_ids{$entrez_id}}) {
-              $add_dependent_xref_sth->execute($xref_id, $dependent_xref_id);
+          # Add link between Ensembl gene and EntrezGene
+          if (defined $entrez_ids->{$entrez_id} ) {
+            foreach my $dependent_xref_id (@{$entrez_ids->{$entrez_id}}) {
+              $self->add_dependent_xref({
+                master_xref_id => $xref_id,
+                acc            => $dependent_xref_id,
+                version        => $version,
```
tgrego: Self comment...
Don't feel this version is correct, might need removing.

misc-scripts/xref_mapping/XrefParser/RefSeqCoordinateParser.pm
```diff
@@ -23,405 +23,302 @@ use strict;
...
-  if($file =~ /host[=][>](\S+?)[,]/){
-    $host = $1;
-  }
-  if($file =~ /port[=][>](\S+?)[,]/){
-    $port =  $1;
-  }
-  if($file =~ /dbname[=][>](\S+?)[,]/){
-    $dbname = $1;
+  # project or db param validation
+  if ( $file_params->{project} ne ('ensembl' || 'ensemblgenomes') && !defined $db ) {
```
mkszuba: This doesn't really do what it is supposed to, does it? Correct me if I'm wrong but ( 'ensembl' || 'ensemblgenomes' ) should always evaluate to simply 'ensembl'.

misc-scripts/xref_mapping/XrefParser/RefSeqCoordinateParser.pm
```diff
@@ -23,405 +23,302 @@ use strict;
...
-      $otherf_dba = $registry->get_DBAdaptor($species_name, 'otherfeatures') if !defined($ofhost);
-      if (defined $otherf_dba) { $otherf_dba->dnadb($core_dba); }
+  $species_name //= shift @{$id2name{$species_id}};
+
+  # prepare registry and core/otherfeatures dba
+  my $registry = "Bio::EnsEMBL::Registry";
+  my ($core_dba, $otherf_dba);
+
+  # for ensembl project, use provided connection details or default to staging
+  if ( $file_params->{project} eq ('ensembl') ) {
```
mkszuba: What is the reason for brackets around 'ensembl' here? I'd need someone with considerably better Perl knowledge than mine to clarify but to me, it looks like it makes the second argument to _eq_ a list.

misc-scripts/xref_mapping/XrefParser/RefSeqCoordinateParser.pm
```diff
@@ -431,4 +328,85 @@ sub run_script {
...
+
+
+# parses the input string $file into an hash
+# string $file is in the format as the example:
+# script:project=>ensembl,host=>ens-staging1,dbname=>homo_sapiens_core_70_37,ofhost=>ens-staging1,...
+# string until : is ignored, hash is built with keys=>values provided
+sub parse_file_string {
+  my ($self, $file_string) = @_;
+
+  $file_string =~ s/\A\w+://xms;
```
mkszuba: Considering this regex has got the /x flag set, it really would be nice to throw in some whitespace here to make the pattern more readable.

misc-scripts/xref_mapping/XrefParser/RefSeqCoordinateParser.pm
```diff
@@ -431,4 +328,85 @@ sub run_script {
...
+
+  my @param_pairs = split( /,/x, $file_string );
+
+  my $params;
+
+  # set provided values
+  foreach my $pair ( @param_pairs ) {
+    my ($key, $value) = split( /=>/x, $pair );
```
mkszuba: I would have the split pattern include optional whitespace on either side of =>, to be absolutely certain we do not end up slurping whitespace into keys or values in the event of it having suddenly begun to appear in the file string.

misc-scripts/xref_mapping/XrefParser/RefSeqCoordinateParser.pm
```diff
@@ -431,4 +328,85 @@ sub run_script {
...
+
+  my $exon_match = 0;
+
+  foreach my $exon (@{$params->{exons}}) {
+    if (defined $params->{check_and_register}) {
+      $params->{check_and_register}->check_and_register( 'exon', $exon->seq_region_start, $exon->seq_region_end );
+    }
+    if (defined $params->{overlap}) {
+      my $overlap = $params->{overlap}->overlap_size('exon', $exon->seq_region_start, $exon->seq_region_end);
+      $exon_match += $overlap/($exon->seq_region_end - $exon->seq_region_start + 1);
```
mkszuba: Spaces around the / operator too, perhaps?

misc-scripts/xref_mapping/XrefParser/RefSeqCoordinateParser.pm
```diff
@@ -23,405 +23,302 @@ use strict;
...
-	} else {
-	  # Skip non conventional accessions
-	  next;
-	}
-        if ($id !~ /^[NXMR]{2}_[0-9]+/)  { next; }
-        my %transcript_result;
-        my %tl_transcript_result;
-        if (!defined $id) { next; }
+      # Create a range registry for all the exons of the refseq transcript
+      foreach my $transcript_of (sort { $a->start <=> $b->start } @{$transcripts_of}) {
```
mkszuba: This is probably my C/C++ background showing but I would really prefer to have names of getters like _start()_ here, display_xref() a bit further down and several others in the whole chromosome loop include () at the end. Especially given those have already been used in quite a few other parts of the code.

misc-scripts/xref_mapping/XrefParser/RefSeqCoordinateParser.pm
```diff
@@ -23,405 +23,302 @@ use strict;
...
+        # stable ids as it used to be in the past. This means 
+        # priority is given to the display_id, and fall back to stable_id
+        # for backwards compatibility.
+        if (defined $transcript_of->display_xref ) {
+          $id = $transcript_of->display_xref->display_id;
+        } elsif (defined $transcript_of->stable_id) {
+          $id = $transcript_of->stable_id;
+        }
+        # Skip non conventional and missing accessions
+        unless ( $id =~ /\A[NXMR]{2}_\d+\.?\d+\z/xms ) {
```
mkszuba: This regex is in dire need of having some whitespace added to it for readability, possibly comments as well. Which we could now do given the /x flag.

misc-scripts/xref_mapping/XrefParser/RefSeqCoordinateParser.pm
```diff
@@ -23,405 +23,302 @@ use strict;
...
-                                          species_id => $species_id,
-                                          dbi => $dbi,
-                                          info_type => 'DIRECT' });
-          $self->add_direct_xref($xref_id, $best_id, "Transcript", "", $dbi);
-
-	  my $t_of = $transcript_of;
-          my $g_of = $t_of->get_Gene();
-          my $entrez_id = $g_of->stable_id();
-          my $tl_of = $t_of->translation();
+          my ($acc, $version) = split(/\./xms, $id);
```
mkszuba: I would use a character class for the dot instead of escaping it but that is a bit of a personal preference.

misc-scripts/xref_mapping/XrefParser/RefSeqCoordinateParser.pm
```diff
@@ -23,405 +23,302 @@ use strict;
...
-  if($file =~ /ofport[=][>](\S+?)[,]/){
-    $ofport =  $1;
+  # set default values
+  $file_params->{user}   //= 'ensro';
+  $file_params->{port}   //= '3306';
+  $file_params->{ofuser} //= 'ensro';
+  $file_params->{ofport} //= '3306';
+
+  # set the list of source names
+  my $source_names = {
```
mkszuba: This could be made read-only, just in case (it's my C++ background showing again... const, const, const).

misc-scripts/xref_mapping/XrefParser/RefSeqCoordinateParser.pm
```diff
@@ -23,405 +23,302 @@ use strict;
...
+    NR => 'RefSeq_ncRNA',
+    XM => 'RefSeq_mRNA_predicted',
+    XR => 'RefSeq_ncRNA_predicted',
+    NP => 'RefSeq_peptide',
+    XP => 'RefSeq_peptide_predicted',
+  };
+
+  my $source_ids; 
+  while (my ($source_prefix, $source_name) = each %{$source_names}) {
+    $source_ids->{$source_prefix} = $self->get_source_id_for_source_name( $source_name, 'otherfeatures', $dbi )
```
mkszuba: You probably should make sure this doesn't return -1 for any source. Or perhaps one of us should finally bite the bullet and patch BaseParser to have this method throw upon a failed lookup.

misc-scripts/xref_mapping/XrefParser/RefSeqCoordinateParser.pm
```diff
@@ -23,405 +23,302 @@ use strict;
...
+    }
+    $registry->load_registry_from_db(
+      '-host' => $file_params->{ofhost},
+      '-port' => $file_params->{ofport},
+      '-user' => $file_params->{ofuser},
+    );
+    $otherf_dba = $registry->get_DBAdaptor($species_name, 'otherfeatures');
+    $otherf_dba->dnadb($core_dba);
+  # for ensemblgenomes project, use staging and ignore any connection details provided
+  } elsif ( $file_params->{project} eq ('ensemblgenomes') ) {
```
mkszuba: See above.

misc-scripts/xref_mapping/XrefParser/RefSeqCoordinateParser.pm
```diff
@@ -23,405 +23,302 @@ use strict;
...
+        # priority is given to the display_id, and fall back to stable_id
+        # for backwards compatibility.
+        if (defined $transcript_of->display_xref ) {
+          $id = $transcript_of->display_xref->display_id;
+        } elsif (defined $transcript_of->stable_id) {
+          $id = $transcript_of->stable_id;
+        }
+        # Skip non conventional and missing accessions
+        unless ( $id =~ /\A[NXMR]{2}_\d+\.?\d+\z/xms ) {
+          next;
```
mkszuba: Given we have got multiple levels of nested loops in this function, I would **strongly** recommend labelling all of them and using labels with each call to _next_, _last_ _etc._

misc-scripts/xref_mapping/XrefParser/RefSeqCoordinateParser.pm
```diff
@@ -23,405 +23,302 @@ use strict;
...
-# Fetch slice in core database which overlaps refseq transcript
+        # Fetch slice in core database which overlaps refseq transcript
         my $chromosome = $sa->fetch_by_region('toplevel', $chr_name, $transcript_of->seq_region_start, $transcript_of->seq_region_end);
         my $transcripts = $chromosome->get_all_Transcripts(1);
 
-# Create a range registry for all the exons of the ensembl transcript
-        foreach my $transcript(@$transcripts) {
+        # Create a range registry for all the exons of the ensembl transcript
+        foreach my $transcript(@{$transcripts}) {
           if ($transcript->strand != $transcript_of->strand) { next; }
```
mkszuba: Could you split this into multiple lines, please?

misc-scripts/xref_mapping/XrefParser/RefSeqCoordinateParser.pm
```diff
@@ -23,405 +23,302 @@ use strict;
...
-                                          info_type => 'DIRECT' });
-          $self->add_direct_xref($xref_id, $best_id, "Transcript", "", $dbi);
-
-	  my $t_of = $transcript_of;
-          my $g_of = $t_of->get_Gene();
-          my $entrez_id = $g_of->stable_id();
-          my $tl_of = $t_of->translation();
+          my ($acc, $version) = split(/\./xms, $id);
+
+          my $source = $source_ids->{substr $acc, 0, 2} // next;
```
mkszuba: I really, really do not like this sort of mixing of flow control and data manipulation. Couldn't we just have an if statement around _next_ instead?

misc-scripts/xref_mapping/XrefParser/RefSeqCoordinateParser.pm
```diff
@@ -23,405 +23,302 @@ use strict;
...
-                                              label => $tl_id,
-                                              desc => undef,
-                                              source_id => $source_id,
-                                              species_id => $species_id,
-                                              dbi => $dbi,
-                                              info_type => 'DIRECT' });
-              $self->add_direct_xref($tl_xref_id, $tl->stable_id(), "Translation", "", $dbi);
+              my ($tl_acc, $tl_version) = split(/\./xms, $tl_id);
+
+              my $tl_source = $source_ids->{substr $tl_acc, 0, 2} // next;
```
mkszuba: See my previous comment on mixing control flow and data manipulation.

misc-scripts/xref_mapping/XrefParser/RefSeqCoordinateParser.pm
```diff
@@ -23,405 +23,302 @@ use strict;
...
-              my $tl_xref_id = $self->add_xref({ acc => $acc,
-                                              version => $version,
-                                              label => $tl_id,
-                                              desc => undef,
-                                              source_id => $source_id,
-                                              species_id => $species_id,
-                                              dbi => $dbi,
-                                              info_type => 'DIRECT' });
-              $self->add_direct_xref($tl_xref_id, $tl->stable_id(), "Translation", "", $dbi);
+              my ($tl_acc, $tl_version) = split(/\./xms, $tl_id);
```
mkszuba: Again, I personally would use a character class instead of escaping.

misc-scripts/xref_mapping/XrefParser/RefSeqCoordinateParser.pm
tgrego: this is unnecessary because if any whitespace is there FetchFiles.pm will scream before it gets here

tgrego:
Pull request https://github.com/Ensembl/ensembl/pull/336 has been closed due to the performance hit of using ensembl-io.
A new pull request has been submitted without the ensembl-io dependency, thus this is no longer required.

misc-scripts/xref_mapping/XrefParser/RefSeqCoordinateParser.pm
tgrego: thanks for noticing. fixed

misc-scripts/xref_mapping/XrefParser/RefSeqCoordinateParser.pm
tgrego: it's harmless but unnecessary, so removed it

tgrego:
Let me know if there's something left that needs changing.

misc-scripts/xref_mapping/XrefParser/RefSeqCoordinateParser.pm
```diff
@@ -22,406 +22,317 @@ package XrefParser::RefSeqCoordinateParser;
...
+            version    => $version,
+            label      => $id,
+            source_id  => $source,
+            species_id => $species_id,
+            dbi        => $dbi,
+            info_type  => 'DIRECT'
+          });
+          $self->add_direct_xref($xref_id, $best_id, 'Transcript', undef, $dbi);
+
+          my $entrez_id = $gene_of->stable_id;
```
magaliruffier: We need the same check here as for the transcript->stable_id
In some cases, the stable_id will be a randomly generated ID and the NCBIGene accession is stored as the display_xref
For example in homo_sapiens_otherfeatures_95_38
select stable_id, x.dbprimary_acc from gene g, xref x where display_xref_id = xref_id and analysis_id = 8360
+-----------+---------------+
| stable_id | dbprimary_acc |
+-----------+---------------+
| gene58752 | 3139          |
| gene58319 | 5555          |
| gene58242 | 6046          |
+-----------+---------------+

PR #330, opened 2018-11-02
Xref parser zfin

ens-bwalts:
## Requirements

## Description

Change parser to utilize Text::CSV. Change parser to use ZFIN files that now contain four columns.
Rename functions for clarity

Restrict Uniprot source to only use SwissProt

## Use case

Parser is updated to meet new xref standards (Text::CSV). Cleanup for readability. Should now work with ZFIN source files

## Benefits

See under Use Case

## Possible Drawbacks

Adds dependency to Text::CSV.

## Testing

Parser works and adds xrefs in the context of the xref pipeline as it stands in early November.

### DIFF
--- a/misc-scripts/xref_mapping/XrefParser/ZFINParser.pm
+++ b/misc-scripts/xref_mapping/XrefParser/ZFINParser.pm
@@ -46,7 +46,7 @@ sub run {
   my $file = @{$files}[0];
   my $dir = dirname($file);
 
-  my (%swiss) = %{$self->get_valid_codes("uniprot/",$species_id, $dbi)};
+  my (%swiss) = %{$self->get_valid_codes("uniprot/swissprot",$species_id, $dbi)};
   my (%refseq) = %{$self->get_valid_codes("refseq",$species_id, $dbi)};
 
   my $swissprot_io =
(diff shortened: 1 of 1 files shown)


misc-scripts/xref_mapping/XrefParser/ZFINParser.pm
```diff
@@ -45,68 +46,78 @@ sub run {
...
+
+  $swissprot_csv->column_names([ 'zfin',
+                                 'so',
+                                 'label',
+                                 'acc',
+  ]);
 
   my %description;
 
   my $sql = "insert ignore into synonym (xref_id, synonym) values (?, ?)";
```
tgrego: the custom sql should be dropped and methods from the (here) base parser used.
In this case BaseParser::add_to_syn() provides this functionality.

misc-scripts/xref_mapping/XrefParser/ZFINParser.pm
```diff
@@ -123,17 +134,29 @@ sub run {
   my $refseq_io = $self->get_filehandle( catfile( $dir, 'refseq.txt' ) );
 
   if ( !defined $refseq_io ) {
-    print STDERR "ERROR: Could not open " . catfile( $dir, 'refseq.txt' ),"\n" ;
+    croak "ERROR: Could not open " . catfile( $dir, 'refseq.txt' ),"\n" ;
     return 1;
```
tgrego: no need to return anymore

misc-scripts/xref_mapping/XrefParser/ZFINParser.pm
```diff
@@ -160,12 +183,25 @@ sub run {
   my $zfin_io = $self->get_filehandle( catfile( $dir, 'aliases.txt' ) );
 
   if ( !defined $zfin_io ) {
-    print STDERR  "ERROR: Could not open " . catfile( $dir, 'aliases.txt' ), "\n" ;
+    croak "ERROR: Could not open " . catfile( $dir, 'aliases.txt' ), "\n" ;
     return 1;
```
tgrego: return

misc-scripts/xref_mapping/XrefParser/ZFINParser.pm
ens-bwalts: This statement handle never gets executed anywhere, as it turns out. I suspect it never got cleaned up when the add_to_syn_for_mult_sources was added further down the file()

ens-bwalts:
Planning to defer replacing some of the SQL until after porting this into ensembl-xref. No sense putting methods into BaseParser only to have to move them into BaseAdaptor.

PR #331, opened 2018-11-06
Use seq_region_start and seq_region_end when updating gene coordinate…

thibauthourlier:
…oordinates in the database

## Requirements

## Description

Use seq_region_start and seq_region_end when updating the coordinates of a gene in a database.

## Use case

If the gene is on a sub Slice and start and end are called, they return the start and end relative to the slice. When start and end are used in update_coords, it will then insert the wrong coordinates of the gene in the database.

## Benefits

It will store the correct coordinates of a gene when updating the coordinates in the database.

## Possible Drawbacks

## Testing

Yes
Pass

### DIFF
--- a/modules/t/gene.t
+++ b/modules/t/gene.t
@@ -474,15 +474,18 @@ cmp_ok($new_gene->start(), '==', 30735607, 'Updated gene start');
 cmp_ok($new_gene->end(), '==', 30815178, 'Updated gene end');
 
 # test update_coords method when working on sub Slice
-my $update_slice = $db->get_SliceAdaptor()->fetch_by_region('chromosome', '20', 30730000, 30815178);
+# to avoid setting the start and end relative to a sub Slice
+# correct coords: 30735607 - 30815178
+# wrong coords: 1 - 79572
+my $update_slice = $db->get_SliceAdaptor()->fetch_by_gene_stable_id('ENSG00000171456');
+# Make sure that the slice of the gene is a sub Slice
 my $update_genes = $update_slice->get_all_Genes();
-cmp_ok(scalar(@$update_genes), '>=', 1, 'Check the region has at least one gene');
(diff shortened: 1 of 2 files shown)


modules/t/gene.t
```diff
@@ -473,6 +473,20 @@ my $new_gene = $ga->fetch_by_stable_id("ENSG00000171456");
 cmp_ok($new_gene->start(), '==', 30735607, 'Updated gene start');
 cmp_ok($new_gene->end(), '==', 30815178, 'Updated gene end');
 
+# test update_coords method when working on sub Slice
+my $update_slice = $db->get_SliceAdaptor()->fetch_by_region('chromosome', '20', 30730000, 30815178);
```
nerdstrike: I feel like this could have been achieved less cryptically with SliceAdaptor->fetch_by_gene_stable_id() or just Gene->feature_Slice(). Then you would set that slice back to the corresponding gene. That way it doesn't matter what genes are in the region. You'd have to double-check it really does expose the same bug though.

thibauthourlier: I'm happy to use fetch_by_gene_stable_id but I don't really see why it makes it less cryptic. If I simply set the sub slice to the gene, it would not expose the same bug unless I add the sub slice to transcripts,...
I will remove the stable_id check in the loop as it is useless

nerdstrike: My request is that the test should self-evidently prove something. The approach you took is a bit obscure as to what it's really doing, what with the loop, the ID test, the know stable ID but not choosing to fetch it via a slice instead... it could be cleaner or better commented.

thibauthourlier: The snipper of code was showing how I fetch the slice, next time I'll know it means the whole commit. As I commented previously, yes the stable id check in the loop is not needed.
I would be interested to know how I can fetch a gene using its stable id and slice.

PR #334, opened 2018-11-14
Fixes in BaseParser::upload_xref_object_graphs()

mkszuba:
## Description

BaseParser::upload_xref_object_graphs() can insert duplicate dependent xrefs on subsequent reruns of parsers using it on the same input, and inserted empty strings rather than NULLs as descriptions of dependent xrefs unless explicitly overridden by arguments.

## Use case

_BaseParser::upload_xref_object_graphs()_ is used by (at least) UniProtParser.

## Benefits

Calls to upload_xref_object_graphs() should now be replay-safe. Fewer empty strings.

## Possible Drawbacks

Parser output will likely change a lot, mostly due to the empty-space-to-null transition - which will make validation challenging.

## Testing

No.

N/A

I have run _xref_parser.t_, which AFAIK is the only part of the test suite dealing with xref parsers. All tests still pass.
Moreover, I have run the new UniProtParser (with a small subset of input extracted from the middle of current Swiss-Prot file) before and after the change to compare results and they have changed as expected.

### DIFF
--- a/misc-scripts/xref_mapping/XrefParser/BaseParser.pm
+++ b/misc-scripts/xref_mapping/XrefParser/BaseParser.pm
@@ -931,11 +931,11 @@ sub add_xref {
   my $acc         = $arg_ref->{acc}        || croak 'add_xref needs aa acc';
   my $source_id   = $arg_ref->{source_id}  || croak 'add_xref needs a source_id';
   my $species_id  = $arg_ref->{species_id} || croak 'add_xref needs a species_id';
-  my $label       = $arg_ref->{label}      || $acc;
+  my $label       = $arg_ref->{label}      // $acc;
   my $description = $arg_ref->{desc};
-  my $version     = $arg_ref->{version}    || 0;
-  my $info_type   = $arg_ref->{info_type}  || 'MISC';
-  my $info_text   = $arg_ref->{info_text}  || '';
+  my $version     = $arg_ref->{version}    // 0;
+  my $info_type   = $arg_ref->{info_type}  // 'MISC';
+  my $info_text   = $arg_ref->{info_text}  // q{};
   my $dbi         = $arg_ref->{dbi};
 
   $dbi = $self->dbi unless defined $dbi;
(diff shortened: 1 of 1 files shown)


misc-scripts/xref_mapping/XrefParser/BaseParser.pm
```diff
@@ -566,42 +565,34 @@ sub upload_xref_object_graphs {
...
-	 $xref_id_sth->execute(
-                   $dep{ACCESSION},
-                   $dep{SOURCE_ID},
-                   $xref->{SPECIES_ID} );
-         my $dep_xref_id = ($xref_id_sth->fetchrow_array())[0];
-
-	 if(!(defined $dep_xref_id) || $dep_xref_id ==0 ){
-	   print STDERR "acc = $dep{ACCESSION} \nlink = $dep{LINKAGE_SOURCE_ID} \n".$dbi->err."\n";
-	   print STDERR "source = $dep{SOURCE_ID}\n";
+         my $dep_xref_id = $xref_sth->{'mysql_insertid'};
```
nerdstrike: As long as this remains a DBI statement handle, perhaps you could use DBI's last_insert_id() instead of the driver method? I'd like for us to become less backend specific, as the core API is.

misc-scripts/xref_mapping/XrefParser/BaseParser.pm
```diff
@@ -566,42 +565,34 @@ sub upload_xref_object_graphs {
...
 	 }
 
 	 #
 	 # Add the linkage_annotation and source id it came from
 	 #
-	 $dep_sth->execute( $xref_id, $dep_xref_id,
-			    $dep{LINKAGE_ANNOTATION},
-			    $dep{LINKAGE_SOURCE_ID} )
-	   or croak( $dbi->errstr() );
+         $self->add_dependent_xref_maponly( $dep_xref_id,
```
nerdstrike: Is this only present in ensembl-xref? I don't think it's valid here.

misc-scripts/xref_mapping/XrefParser/BaseParser.pm
mkszuba: I've used the _mysql_insertid_ way because that's what _add_xref()_ uses - but I am all for being database-agnostic.

misc-scripts/xref_mapping/XrefParser/BaseParser.pm
mkszuba: No, I am positive (and have just checked to be sure) this has already been merged into _feature/xref_sprint_.

misc-scripts/xref_mapping/XrefParser/BaseParser.pm
```diff
@@ -580,20 +581,9 @@ sub upload_xref_object_graphs {
...
-                   $dep{ACCESSION},
-                   $dep{SOURCE_ID},
-                   $xref->{SPECIES_ID} );
-         my $dep_xref_id = ($xref_id_sth->fetchrow_array())[0];
-
-	 if(!(defined $dep_xref_id) || $dep_xref_id ==0 ){
-	   print STDERR "acc = $dep{ACCESSION} \nlink = $dep{LINKAGE_SOURCE_ID} \n".$dbi->err."\n";
-	   print STDERR "source = $dep{SOURCE_ID}\n";
+         my $dep_xref_id = $dbi->last_insert_id();
+	 if( ! $dep_xref_id ) {
```
nerdstrike: Smells like a tab!

mkszuba: It is, that whole block has been indented with a mixture of tabs and spaces - with my additions being the only lines there only indented with the latter.

nerdstrike: I will forgive you eventually.

PR #335, opened 2018-11-14
Rewritten UniProtParser

mkszuba:
**Edited: DO NOT MERGE at this point, please see the "Testing" section below for details.**

## Description

The old UniProtParser is very complex, includes a lot of legacy code, contains several bugs (see _e.g._ ENSCORESW-2837) and would be difficult to adapt to follow the Extract-Transform-Load process of operation.

## Use case

UniProt-KB xrefs are a crucial component of every Ensembl release.

## Benefits

Hopefully cleaner, more modular and ETL-friendly code base which might be easier to transplant to the refactored xref pipeline or extend (_e.g._ to process XML rather than text dumps or to query the UniProt database directly) in the future.

## Possible Drawbacks

Differences in the output of the old and the new parser due to bug fixes, which will be fairly challenging to validate given the size of the input.

## Testing

No.

N/A

1. I have confirmed that the new parser works fine on and produces expexted results from a small subset of both Swiss-Prot and TrEMBL data - **but only if the input is uncompressed**, gzipped files cause a "broken pipe" error in zcat.
2. xref_parser.t, which got added to the repository by PR #286 and which tests, among others, UniProtParser, **fails**. Problems pertaining directly to UniProtParser have been traced down to bugs in the test suite and subsequently fixed, unfortunately it turns out WormbaseCElegansUniProtParser inherits UniProtParser and will therefore have to be refactored as well in order to become compatible with the ETL version.

### DIFF
--- a/misc-scripts/xref_mapping/XrefParser/UniProtParser.pm
+++ b/misc-scripts/xref_mapping/XrefParser/UniProtParser.pm
@@ -1,7 +1,7 @@
 =head1 LICENSE
 
-Copyright [1999-2015] Wellcome Trust Sanger Institute and the EMBL-European Bioinformatics Institute
-Copyright [2016-2018] EMBL-European Bioinformatics Institute
+See the NOTICE file distributed with this work for additional
+information regarding copyright ownership.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


misc-scripts/xref_mapping/XrefParser/UniProtParser/Extractor.pm
```diff
@@ -0,0 +1,645 @@
...
+sub _get_quality {
+  my ( $self ) = @_;
+
+  Readonly my $id_status_field
+    => qr{
+           (?: Unreviewed )
+         | (?: Reviewed )
+       }msx;
+
+  # These is only one ID line
```
nerdstrike: Argument error

misc-scripts/xref_mapping/XrefParser/UniProtParser/Extractor.pm
```diff
@@ -0,0 +1,645 @@
...
+  # current SwissProt data file doesn't contain any records which
+  # violate this. On the other hand we haven't checked this in the
+  # TrEMBL file because of its size and the old UniProtParser did have
+  # support for synonyms (albeit using different syntax -
+  # "db=id1, id2, ...;" rather than "db1=id1; db2=id2; ...").
+  # The code below assumes there might be multiple entries present, if
+  # you want to force it to only ever look for one simply drop the /g
+  # modifier from the regex match.
+
+  Readonly my $taxon_db_entry
```
nerdstrike: As per my earlier comment about seemingly defining in every iteration.

misc-scripts/xref_mapping/XrefParser/UniProtParser/Extractor.pm
```diff
@@ -0,0 +1,645 @@
...
+
+  return \@extracted_taxon_codes;
+}
+
+
+sub _record_has_all_needed_fields {
+  my ( $self ) = @_;
+
+  # Only check mandatory fields
+  my @needed_fields = grep { $prefixes_of_interest{$_} } %prefixes_of_interest;
```
nerdstrike: I think you could use a `keys` here or It'll do interesting things with every other element.

misc-scripts/xref_mapping/XrefParser/UniProtParser/Loader.pm
```diff
@@ -0,0 +1,114 @@
...
+use Carp;
+
+
+sub new {
+  my ( $proto, $arg_ref ) = @_;
+
+  my $self = {
+              'batch_size'  => $arg_ref->{'batch_size'} // 1,
+              'dbh'         => $arg_ref->{'dbh'},
+              '_baseParser' => $arg_ref->{'baseParser'},
```
nerdstrike: Another camelCase

misc-scripts/xref_mapping/XrefParser/UniProtParser/Loader.pm
```diff
@@ -0,0 +1,114 @@
...
+
+
+sub load {
+  my ( $self, $transformed_data ) = @_;
+
+  if ( ! defined $transformed_data ) {
+    return;
+  }
+
+  $self->_add_to_send_buffer( $transformed_data );
```
nerdstrike: I think this buffering should ultimately live in the DB abstraction class as yet to be realised, so that all parsers may buffer up their loads. I'm also not totally convinced you won't have an unflushed buffer at the end with n > 1 batch_size.

misc-scripts/xref_mapping/XrefParser/UniProtParser/Transformer.pm
```diff
@@ -0,0 +1,464 @@
...
+}
+
+
+# Returns true if the current record describes an entry derived from
+# Ensembl, false otherwise.
+sub _entry_is_from_ensembl {
+  my ( $self ) = @_;
+
+  # As of end of October 2018, the old parser's way of identifying
+  # proteins derived from Ensembl by searching for a comment topic
```
nerdstrike: @magaliruffier - does this sound right to you?

misc-scripts/xref_mapping/XrefParser/UniProtParser/Extractor.pm
```diff
@@ -0,0 +1,645 @@
...
+# FIXME: at the moment the extractor only handles one input file at a
+# time for backwards compatibility with the old UniProtParser, that
+# said there is no reason for it not to be able to handle multiple
+# files. Should we want to support this, we should stop opening the
+# filehandle in the constructor because it would no longer be a fixed
+# property of the object.
+sub new {
+  my ( $proto, $arg_ref ) = @_;
+  my $file_names = $arg_ref->{'file_names'};
+  my $baseParserInstance = $arg_ref->{'baseParser'};
```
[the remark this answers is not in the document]
mkszuba: This is intentional in order to draw attention to the passing of BaseParser from the parser itself, which should be terminated with extreme prejudice as soon as BaseAdaptor and friends have become a thing.

misc-scripts/xref_mapping/XrefParser/UniProtParser.pm
```diff
@@ -1,7 +1,7 @@
 =head1 LICENSE
 
-Copyright [1999-2015] Wellcome Trust Sanger Institute and the EMBL-European Bioinformatics Institute
-Copyright [2016-2018] EMBL-European Bioinformatics Institute
+See the NOTICE file distributed with this work for additional
```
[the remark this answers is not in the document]
mkszuba: Fixed.

misc-scripts/xref_mapping/XrefParser/UniProtParser/Loader.pm
mkszuba: See the previousComment.

misc-scripts/xref_mapping/XrefParser/UniProtParser/Loader.pm
mkszuba: Not by itself, no. UniProtParser::run() takes care of that by calling $loader->flush() after the loop.

misc-scripts/xref_mapping/XrefParser/UniProtParser/Transformer.pm
mkszuba: This is what Mag told me last week after having checked.

magaliruffier: Correct

misc-scripts/xref_mapping/XrefParser/UniProtParser/Extractor.pm
mkszuba: Fixed as well, along with all the other occurences of regex constants in Extractor.

misc-scripts/xref_mapping/XrefParser/UniProtParser/Extractor.pm
mkszuba: D'oh. Harmless in this context but you are of course right, grep simply takes a list.

misc-scripts/xref_mapping/XrefParser/UniProtParser/Extractor.pm
mkszuba: Meesa embarrassed!

misc-scripts/xref_mapping/XrefParser/UniProtParser/Loader.pm
mkszuba: Added a call to flush() to the destructor, just in case.

modules/t/xref_parser.t
```diff
@@ -40,6 +40,11 @@ $database->populate(
   "with force please",
 );
 
+
+# UniProt species taxon codes happen to match species ids of the core database
+my $SPECIES_ID = 6239;
+my $SPECIES_NAME = "Caenorhabditis elegans";
```
wbazant: @mkszuba the test database created here is for homo sapiens. You'd probably want to change the test to use a C. elegans core database if you want it to both use a species id 6239 and remain useful - I'm a fan of the second part more.

modules/t/xref_parser.t
```diff
@@ -180,7 +184,7 @@ my @recognised_sources = (
  "MEROPS; C26.956; -.",
 );
 for my $l (@recognised_sources) {
-  (my $uniprot_elegans_record_extra_line = $uniprot_elegans_record) =~ s/DR(.*?)\n/DR$1\nDR  $l/;
+  (my $uniprot_elegans_record_extra_line = $uniprot_elegans_record) =~ s/DR(.*?)\n/DR$1\nDR   $l/;
```
wbazant: I don't understand this change. Was the regex wrong in the first place, or did the UniProt format change?

Perhaps \s+ would be better than adjusting the amount of spaces.

misc-scripts/xref_mapping/XrefParser/UniProtParser/Loader.pm
```diff
@@ -0,0 +1,122 @@
...
+  return;
+}
+
+
+sub _add_to_send_buffer {
+  my ( $self, $entry ) = @_;
+
+  # Play it safe and assume that parser-private fields *could* pollute
+  # the output unless removed from it
+  my ( $multiplicity ) = delete $entry->{'_multiplicity'};
```
wbazant: I know a single UniProt record can generate multiple xrefs, because there can be a few or no dependent xrefs, and that this varies per record, but I'm not sure why you care about smoothing it out enough to add this construct. Batching at a level higher, pushing e.g. a 1000 xrefs and all their dependent xrefs, will probably work just as well.

mkszuba:
The test suite **has** been updated and all UniProtParser tests pass now. In fact, the only test in this whole suite which fails is XrefParser::WormbaseCElegansUniProtParser Correct xrefs and dependent xrefs - which is to be expected because like I said in the e-mail to you, the way it extends UniProtParser is not compatible with the ETL structure.

misc-scripts/xref_mapping/XrefParser/UniProtParser/Loader.pm
mkszuba: Looking at the species_id-taxonomy_id map in the Ensembl database, for certain species the number of generated xrefs can be several times that of the number of parsed UniProt records. I could live with the loader inserting e.g. 1200 xrefs instead of 1000 (the batch size is still a thousand) but expecting 1000 and inserting _e.g._ 3000 would be a fairly sizeable discrepancy. Therefore, with multiplicity counting having been implemented in the old UniProtParser I saw no reason not to implement it in the new version.

wbazant:
Great! XrefParser::WormbaseCElegansUniProtParser  should be simple to fix too. Please fix it if you can!
One benefit of keeping these parsers together is that we don't break each other's stuff inadvertently (though it clearly also happens that we break each other's stuff intentionally)

modules/t/xref_parser.t
mkszuba: The regex was wrong. Quoting UniProtKB User Manual (https://web.expasy.org/docs/userman.html#entrystruc, at the very end of the section): "The two-character line-type code that begins each line is always followed by three blanks, so that the actual information begins with the sixth character."

modules/t/xref_parser.t
mkszuba: I can change the ID to 9606 if you prefer. As long as it is a valid value, I do not mind.

wbazant: I remember I had to make them match and the default test db I created was for human. If the test passes it doesn't matter what it is.

misc-scripts/xref_mapping/XrefParser/UniProtParser/Loader.pm
wbazant: You're now getting it from `$self->_recognised_taxon_ids();` in the Transformer. It's some per-species value that I think doesn't achieve what you want.

If this field is meant to be about evening out the payload, you could count the number of dependent xrefs in the loader - 1 plus however many dependent xrefs it has.

misc-scripts/xref_mapping/XrefParser/UniProtParser.pm
```diff
@@ -17,521 +17,163 @@ limitations under the License.
...
+  }
 
-    #Use normal source only if it is PE levels 1 & 2
-      if (defined($protein_evidence_code) && $protein_evidence_code < 3) {
-          $xref->{SOURCE_ID} = $sptr_source_id;
-          $num_sptr++;
-      } else {
-          $xref->{SOURCE_ID} = $sptr_non_display_source_id;
-          $num_sptr_non_display++;	  
+  my $extractor = XrefParser::UniProtParser::Extractor->new({
```
wbazant: @mkszuba @nerdstrike I've just realised that I won't be able to fix `WormbaseCElegansUniProtParser` if you do this unless I mess around with %INC to substitute my own `XrefParser::UniProtParser::Transformer` or something like that.

I think separating E, T, and L is a good idea into logical blocks and it's good that you've identified breaks in the previous monolith that separate it into more reasonable chunks.

I don't like where you've drawn a division between E and T. In other parsers (e.g. Tiago's new RefSeqGPFFParser) extracting means separating chunks of text at a time. Then transforming is taking the data's original format, and returning a maybe-record in an application's internal format. You do a lot of transforming in your extractor, and a transformer also filters (skipping unreviewed entries and stuff). Secondly, `XrefParser::UniProtParser::Extractor` requiring a `baseParserInstance` is also apparent enough that it's inconvenient to make it its own object.

Consider keeping extracting in the `XrefParser::UniProtParser`, where you can draw from `XrefParser::BaseParser`. @nerdstrike mentioned you'll be making a `XrefParser::BaseAdaptor` instead or something - maybe you won't even need much or any UniProt-specific behaviour to get a clean chunk of code responsible for knowing where the file is, and then opening a stream of text chunks.

I recommend you move the logic of examining the record to build Ensembl-specific things from it contained in the T part. Make it a self contained module taking a piece of text (perhaps a ref to a scalar to avoid copying) and returning either nothing - if the entry should be rejected - or an Ensembl object. It's a fantastic opportunity for encapsulation. It will be maintainable, and will let you do robust unit testing - get some canned records, say what Ensembl objects they should make, compare with `is_deeply`, and you can very easily get to 100% test coverage for it.

Finally I think your Loader should become a utility class, say a `RecordBatchingQueue`. This is the crux of your Loader really, and it's not specific to UniProt, it will work on any xrefs with dependent xrefs that it can count. Counting xrefs should not be tied to also calling the parent. The behaviour you've designed is very general and possibly so useful that it's a shame if all the other modules have to batching in their own way. Move it somewhere, think how to make it general (Java's iterators constructed from other iterators, and registering a callback that processes a full batch, are two patterns that come to my mind), add some unit tests for it, and you can solve the problem of parsed xrefs overloading the database for all the other 25981 parsers too.

If you think the current design with three objects is superior, and also you don't want to create your E, T, and L using dependency injection, could you at least put the transformer in a `sub transform`? If there is a separate sub that creates a XrefParser::UniProtParser::Transformer and applies it to the arguments, I will be able to rewrite `WormbaseCElegansUniProtParser` to inherit from it - but I won't be able to use this code in its current form.

mkszuba:
Thank you for the comments, everyone. Seeing as it is presently not possible to merge this without affecting the Wormbase parser, I'll close this PR for now.

PR #336, opened 2018-11-14
RefSeqGPFFParser update

tgrego:
## Description

Update to the RefSeqGPFFParserr as part of the efforts of the xref sprint.
See ENSCORESW-2898.
Genbank parser from ensembl-io is now used to parse the source files instead of a custom parser.
This will require https://github.com/Ensembl/ensembl-io/pull/69 to be merged.
There are a few differences introduced from the original parser. For instance, only refseq ids with the prefixes defined in $refseq_sources are considered. Previously for peptide files all other possible types were considered and treated as RefSeq_peptide.

## Testing
No unit tests.
Tested with subset of rat, however related xrefs were absent.
Ongoing testing with full dataset.

### DIFF
--- a/misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
+++ b/misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
@@ -17,25 +17,19 @@ limitations under the License.
 
 =cut
 
-# Parse RefSeq GPFF files to create xrefs.
-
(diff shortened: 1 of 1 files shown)


nerdstrike:
Your build requires ensembl-io as a dependency

tgrego:
Due to performance issues related to the usage of ensembl-io, https://github.com/Ensembl/ensembl/pull/338 has been submitted with the refactoring changes without the ensembl-io dependency.

PR #338, opened 2018-11-20
RefSeqGPFFParser update, no ensembl-io

tgrego:
## Description

Update to the RefSeqGPFFParser as part of the efforts of the xref sprint.
See ENSCORESW-2898.

This pull request is a follow up on https://github.com/Ensembl/ensembl/pull/336
keeping the refactoring, but removing the ensembl-io dependency.

WormbaseCElegansRefSeqGPFFParser depends on RefSeqGPFFParser and is non-functional with this pull request, thus requiring changes.

## Testing

No unit tests for this parser directly.
Tested with subset of rat, however related xrefs were absent.
Ongoing testing with full dataset.
WormbaseCElegansRefSeqGPFFParser has tests in modules/t/xref_parser.t
these will be temporarily silenced until changes are committed to WormbaseCElegansRefSeqGPFFParser

### DIFF
--- a/misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
+++ b/misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
@@ -23,10 +23,19 @@ use strict;
 use warnings;
 use Carp;
 use List::MoreUtils qw(uniq);
-use Bio::EnsEMBL::IO::Parser::Genbank;
+use Readonly;
 
 use parent qw( XrefParser::BaseParser );
 
+# Refseq sources to consider. Prefixes not in this list will be ignored
+Readonly my $refseq_sources => {
+    NM => 'RefSeq_mRNA',
+    NR => 'RefSeq_ncRNA',
+    XM => 'RefSeq_mRNA_predicted',
+    XR => 'RefSeq_ncRNA_predicted',
+    NP => 'RefSeq_peptide',
+    XP => 'RefSeq_peptide_predicted',
+};
 
 
 sub run {
@@ -44,48 +53,40 @@ sub run {
     croak "Need to pass source_id, species_id, files and rel_file as pairs";
   }
 
-  # set the list of RefSeq source names
-  my $refseq_sources = {
-    NM => 'RefSeq_mRNA',
-    NR => 'RefSeq_ncRNA',
-    XM => 'RefSeq_mRNA_predicted',
-    XR => 'RefSeq_ncRNA_predicted',
-    NP => 'RefSeq_peptide',
-    XP => 'RefSeq_peptide_predicted',
-  };
+  $self->{species_id} = $species_id;
+  $self->{verbose} = $verbose;
 
   # get RefSeq source ids
-  my $source_ids = {};
   while (my ($source_prefix, $source_name) = each %{$refseq_sources}) {
-    $source_ids->{$source_name} = $self->get_source_id_for_source_name( $source_name, undef, $dbi )
+    $self->{source_ids}->{$source_name} = $self->get_source_id_for_source_name( $source_name, undef, $dbi )
   }
 
   # get extra source ids
-  $source_ids->{EntrezGene} = $self->get_source_id_for_source_name('EntrezGene', undef, $dbi);
-  $source_ids->{WikiGene} = $self->get_source_id_for_source_name('WikiGene', undef, $dbi);
+  $self->{source_ids}->{EntrezGene} = $self->get_source_id_for_source_name('EntrezGene', undef, $dbi);
+  $self->{source_ids}->{WikiGene} = $self->get_source_id_for_source_name('WikiGene', undef, $dbi);
 
   # Retrieve existing RefSeq mRNA
-  my $refseq_ids = { %{$self->get_valid_codes('RefSeq_mRNA', $species_id, $dbi)},
+  $self->{refseq_ids} = { %{$self->get_valid_codes('RefSeq_mRNA', $species_id, $dbi)},
       %{$self->get_valid_codes('RefSeq_mRNA_predicted', $species_id, $dbi)} };
-  my $entrez_ids = $self->get_valid_codes('EntrezGene', $species_id, $dbi);
-  my $wiki_ids = $self->get_valid_codes('WikiGene', $species_id, $dbi);
+  $self->{entrez_ids} = $self->get_valid_codes('EntrezGene', $species_id, $dbi);
+  $self->{wiki_ids} = $self->get_valid_codes('WikiGene', $species_id, $dbi);
 
   if ($verbose) {
     for my $source_name (sort values %{$refseq_sources}) {
-      print "$source_name source ID = $source_ids->{$source_name}\n";
+      print "$source_name source ID = $self->{source_ids}->{$source_name}\n";
     }
   }
 
   # populate entrez gene id => label hash
-  my $entrez = $self->get_acc_to_label("EntrezGene", $species_id, undef, $dbi);
+  $self->{entrez} = $self->get_acc_to_label('EntrezGene', $species_id, undef, $dbi);
 
   # get the species name, prepare species related data checks
   my %species2name = $self->species_id2name($dbi);
   $species_name //= shift @{$species2name{$species_id}};
 
-  my %name2species_id = map{ $_=>$species_id } @{$species2name{$species_id}};
+  $self->{name2species_id} = { map{ $_=>$species_id } @{$species2name{$species_id}} };
   my %species2tax  = $self->species_id2taxonomy($dbi);
-  my %taxonomy2species_id = map{ $_=>$species_id } @{$species2tax{$species_id}};
+  $self->{taxonomy2species_id} = { map{ $_=>$species_id } @{$species2tax{$species_id}} };
 
   # process the source files
   foreach my $file (@{$files}) {
@@ -101,119 +102,22 @@ sub run {
       return;
     }
 
-    # instantiate ensembl-io genbank parser
-    my $parser = Bio::EnsEMBL::IO::Parser::Genbank->open($refseq_fh);
-
     # this will hold the array of xrefs to bulk insert
     my $xrefs;
 
-    # For each record in the file
-    while ( $parser->next ) {
+    do {
+      local $/ = "\/\/\n";
+      while ( my $item = $refseq_fh->getline() ) {
+        my $xref = $self->get_xref_from_record({
+          record => $item,
+          type   => $type
+        });
 
-      # Get the record species
-      my $record_species = lc $parser->get_organism;
-      $record_species =~ s/\s+/_/xg;
-
-      my $record_taxon = $parser->get_taxon_id;
-
-      # get the record species id from the record name, if not found try from the record taxon_id
-      my $record_species_id = $name2species_id{$record_species} // $taxonomy2species_id{$record_taxon};
-
-      # skip if species is not the required
-      next unless ( defined $record_species_id && ($record_species_id eq $species_id) );
-
-      # get the acc
-      my $acc = $parser->get_accession;
-
-      # get description and remove the [species] at the end
-      my $description = $parser->get_description;
-      $description =~ s/\s*\[[^]]*\]\.?\z//x;
-
-      # get refseq pair if available
-      my $refseq_pair = $parser->get_dbsource_acc;
-
-      # get the source_id for this acc type, warning and skip if not found
-      my $prefix = substr($acc, 0, 2);
-      if (!exists $refseq_sources->{$prefix}) {
-        warn "WARNING: can't get source ID for $type $acc. Skipping\n";
-        next;
-      }
-      my $acc_source_id = $source_ids->{$refseq_sources->{$prefix}};
-
-      # the "pair" is the id from coded_by, or the dbsource_acc if that does not exist
-      my $pair = pop @{$parser->get_coded_by_list};
-      $pair //= $parser->get_dbsource_acc;
-
-      # set up the direct xref
-      my $xref = {
-        ACCESSION     => $acc,
-        LABEL         => $parser->get_sequence_name,
-        VERSION       => $parser->get_sequence_version,
-        DESCRIPTION   => $description,
-        INFO_TYPE     => 'SEQUENCE_MATCH',
-        PAIR          => $pair,
-        SEQUENCE      => $parser->get_sequence,
-        SEQUENCE_TYPE => $type,
-        SOURCE_ID     => $acc_source_id,
-        SPECIES_ID    => $species_id,
-        PROTEIN       => pop @{$parser->get_protein_id_list},
-      };
-
-      # retrieve and remove duplicates from referenced entrez GeneID
-      my $entrez_gene_ids = $parser->get_db_xref_list_for_type('GeneID');
-      my @gene_ids = uniq( @{$entrez_gene_ids}); 
-
-      # process existing entrez_gene_ids as dependent xrefs
-      foreach my $gene_id (@gene_ids) {
-
-        next unless (defined $entrez->{$gene_id});
-
-        push @{$xref->{DEPENDENT_XREFS}}, {
-            SOURCE_ID         => $source_ids->{'EntrezGene'},
-            LINKAGE_SOURCE_ID => $acc_source_id,
-            ACCESSION         => $gene_id,
-            LABEL             => $entrez->{$gene_id}
-        };
-
-        push @{$xref->{DEPENDENT_XREFS}}, {
-            SOURCE_ID         => $source_ids->{'WikiGene'},
-            LINKAGE_SOURCE_ID => $acc_source_id,
-            ACCESSION         => $gene_id,
-            LABEL             => $entrez->{$gene_id}
-        };
-
-        next unless (defined $refseq_pair);
-
-        # discard the version number
-        $refseq_pair =~ s/\.\d+//x;
-
-        # Add xrefs for RefSeq mRNA as well where available
-        foreach my $refseq_id (@{ $refseq_ids->{$refseq_pair} }) {
-          foreach my $entrez_id (@{ $entrez_ids->{$gene_id} }) {
-            $self->add_dependent_xref_maponly(
-                $entrez_id,
-                $source_ids->{'EntrezGene'},
-                $refseq_id,
-                $source_ids->{$refseq_sources->{substr($refseq_id, 0, 2)}},
-                $dbi
-            );
-          }
-          foreach my $wiki_id (@{ $wiki_ids->{$gene_id} }) {
-            $self->add_dependent_xref_maponly(
-                $wiki_id,
-                $source_ids->{'WikiGene'},
-                $refseq_id,
-                $source_ids->{$refseq_sources->{substr($refseq_id, 0, 2)}},
-                $dbi
-            );
-          }
+        if (defined $xref) {
+          push @{$xrefs}, $xref;
         }
-
       }
-
-      push @{$xrefs}, $xref;
-
-    }
+    };
 
     $refseq_fh->close();
 
@@ -244,7 +148,7 @@ sub run {
       # set release info
       $self->set_release( $source_id, $release_string, $dbi );
       for my $source_name (sort values %{$refseq_sources}) {
-        $self->set_release( $source_ids->{$source_name}, $release_string, $dbi );
+        $self->set_release( $self->{source_ids}->{$source_name}, $release_string, $dbi );
       }
       if ($verbose) {
         print "RefSeq release: '$release_string'\n";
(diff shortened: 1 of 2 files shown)


tgrego:
@wbazant This brings changes that affect Wormbase code.
Please keep an eye and coordinate with me required changes to the affected parser tests.

wbazant:
@tgrego can I push to this branch?

tgrego:
@wbazant I did an effort to keep the overridden method xref_from_record(), although arguments have changed.
Also the used type_from_file() has been kept, although a lot stricter (and perhaps I'll make it even stricter to make sure it returns only peptide or dna).
you're welcome to push to this branch.

modules/t/xref_parser.t
```diff
@@ -364,8 +364,14 @@ sub test_refseq {
...
   }, "$type RefSeq entries hang off INSDC entries", tmp_file_name => $file);
 }
-test_refseq($refseq_protein_elegans_record, "protein");
-test_refseq($refseq_mrna_elegans_record, "mrna");
+
+# FIXME
+# silenced due to refactoring of XrefParser::RefSeqCoordinateParser
+# pending updates to XrefParser::WormbaseCElegansRefSeqGPFFParser
+
+# test_refseq($refseq_protein_elegans_record, "protein");
```
wbazant: The tests pass locally if I merge https://github.com/Ensembl/ensembl/pull/337/commits/7c80847980e51243cf0ed409c404193ed0410be0 and do this:
```
sub test_refseq {
   my ($record, $type) = @_;
-  my $file = "something_that_says_$type";
+  my $file = "RefSeq_$type";
   test_parser("XrefParser::RefSeqGPFFParser",$record, {
```
Thanks to you keeping the interfaces I think it should all still work!

misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
```diff
@@ -17,340 +17,341 @@ limitations under the License.
...
 }
 
-# --------------------------------------------------------------------------------
+# get type from filename path. this includes the source name and that's enough to extract it
+sub type_from_file {
+  my ($self, $file) = @_;
+
+  my ($type) = $file =~ /RefSeq_(\w+)\//x;
+
+  warn "WARNING: Could not work out sequence type for '$file'\n" unless $type;
```
wbazant: As far as I can see, after your refactor the only place `$type` is used is in
```
    SEQUENCE_TYPE => $params->{type},
```
(apart from `WormbaseCElegansRefSeqGPFFParser`). 

If that's the case, it can only be a few things that are later accepted, I think in the sequence match part of the pipeline. You mentioned restricting it to just mrna / protein - do that! Maybe even with a big comment explaining the dependency and what this is tied to, referring to constants in the module where the sequence match code is, or something like that.

misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
```diff
@@ -17,340 +17,341 @@ limitations under the License.
...
+    # get the file handler
+    my $refseq_fh = $self->get_filehandle($file);
+
+    if ( !defined $refseq_fh ) {
+      warn "WARNING: Can't open RefSeqGPFF file $file\n";
+      return;
+    }
+
+    # this will hold the array of xrefs to bulk insert
+    my $xrefs;
```
wbazant: `$xrefs` instead of `@xrefs` puzzled me at the beginning. You do you and it's of course equivalent, but I think "different things should look different" and I try to use @ and % wherever I can. At first I thought the code is e.g. passing `$xrefs` into a subroutine and pushing to it elsewhere - which would require using a ref - but happily it doesn't.

Also the comment about what the variable contains seems necessary, you could instead make the variable longer or make it say `@result` or something.

misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
```diff
@@ -17,340 +17,341 @@ limitations under the License.
...
-
-        $self->set_release( $source_id,              $release, $dbi );
-        $self->set_release( $peptide_source_id,      $release, $dbi );
-        $self->set_release( $mrna_source_id,         $release, $dbi );
-        $self->set_release( $ncrna_source_id,        $release, $dbi );
-        $self->set_release( $pred_mrna_source_id,    $release, $dbi );
-        $self->set_release( $pred_ncrna_source_id,   $release, $dbi );
-        $self->set_release( $pred_peptide_source_id, $release, $dbi );
+    # upload the xrefs
+    $self->upload_xref_object_graphs( $xrefs, $dbi );
```
wbazant: See above comment - you could
```
if(@xrefs){
   $self->upload_xref_object_graphs( \@xrefs, $dbi );
}
```

misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
```diff
@@ -17,340 +17,341 @@ limitations under the License.
...
+
+  }
+
+
+  # process the release file
+  if ( defined $release_file ) {
+    # get filehandle
+    my $release_io = $self->get_filehandle($release_file);
```
wbazant: I wonder if this filehandle ever gets closed.

misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
```diff
@@ -17,340 +17,341 @@ limitations under the License.
...
-        $self->set_release( $pred_mrna_source_id,    $release, $dbi );
-        $self->set_release( $pred_ncrna_source_id,   $release, $dbi );
-        $self->set_release( $pred_peptide_source_id, $release, $dbi );
+    # upload the xrefs
+    $self->upload_xref_object_graphs( $xrefs, $dbi );
+
+  }
+
+
+  # process the release file
```
wbazant: If the arguments specify a release file and this doesn't work - no release file, can't parse it, or something - it probably represents a critical failure due to something being very wrong, and I wonder if the parser shouldn't fail fast instead of at the end. Maybe the more robust sequence would be:
2) get release string
1) get and process xrefs
3) set release
instead of current 1,2,3.

misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
```diff
@@ -17,340 +17,341 @@ limitations under the License.
...
+    if ( !defined $refseq_fh ) {
+      warn "WARNING: Can't open RefSeqGPFF file $file\n";
+      return;
+    }
+
+    # this will hold the array of xrefs to bulk insert
+    my $xrefs;
+
+    do {
+      local $/ = "\/\/\n";
```
wbazant: I like that this is in a `do` block :)

misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
```diff
@@ -17,340 +17,341 @@ limitations under the License.
...
-    print "RefSeq_peptide_predicted source ID = $pred_peptide_source_id\n";
-    print "RefSeq_mRNA_predicted source ID = $pred_mrna_source_id\n" ;
-    print "RefSeq_ncRNA_predicted source ID = $pred_ncrna_source_id\n" ;
+
+  $self->{species_id} = $species_id;
+  $self->{dbi} = $dbi;
+  $self->{verbose} = $verbose;
+
+  # get RefSeq source ids
+  while (my ($source_prefix, $source_name) = each %{$REFSEQ_SOURCES}) {
```
wbazant: I like that these things are all packed together instead of being in 10 different arguments. Using the self hash to store it all is fine but you could also package them into an object, say `$reference_info`, and pass it to `xref_from_record` this way. You'll keep the benefits, it will be clear that `xref_from_record` needs all that reference info. If you also pass `$dbi` as argument then `xref_from_record` will become a pure function if you're into that.

modules/t/xref_parser.t
tgrego: reopened some of the tests that now pass, some continue off
take a look at them to see what can be done

misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
tgrego: type is now always based on the name of the source.
That is RefSeq_peptide and RefSeq_dna that translate into types peptide and dna.

misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
tgrego: a matter of style, I think @ and % are used far too often.
In this case @ would be equally good.

misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
tgrego: indeed, closed now
thanks for noticing

misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
tgrego: It's not a critical failure, it's not even a failure just a warning.
did set it to verbose only

misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
mkszuba: I personally would explicitly assign [] to this in order to make this visually stand out as an arrayref, that said this _is_ a matter of style.

On the other hand, given this is used number-of-records-in-$refseq_fh times as an array and only once as a reference, it might make sense to make this an array after all.

misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
```diff
@@ -17,340 +17,330 @@ limitations under the License.
...
+    # upload the xrefs
+    $self->upload_xref_object_graphs( $xrefs, $dbi ) if $xrefs;
+
+  }
+
+
+  # process the release file
+  if ( defined $release_file ) {
+    # get filehandle
+    my $release_fh = $self->get_filehandle($release_file);
```
mkszuba: In the refactored BaseParser _get_filehandle()_ will throw on errors but for now, you should check definedness of its return value - which you already do for the main data file - to be on the same side.

misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
```diff
@@ -17,340 +17,330 @@ limitations under the License.
...
+
+    # get file header
+    my $release = do { local $/ = "\n*"; <$release_fh> };
+    $release_fh->close();
+
+    $release =~ s/\s+/ /xg;
+
+    if ( $release =~ m/(NCBI.*Release\s\d+)\s(.*)\sDistribution/x ) {
```
mkszuba: Ouch. You've got /x enabled, use it! Moreover, I would consider adding + to all the \s to make the match less fragile - even there really are only single ones at present.

misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
```diff
@@ -17,340 +17,330 @@ limitations under the License.
...
+      my ($rel_number, $rel_date) = ($1, $2);
```
mkszuba: Not a major issue given this is only used once but from the point of view of performance, it would be better to call the match in list context and assign captures directly to variables than to use numbered capture variables.

misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
```diff
@@ -17,340 +17,330 @@ limitations under the License.
...
+  # get the species name, prepare species related data checks
+  my %species2name = $self->species_id2name($dbi);
+  $species_name //= shift @{$species2name{$species_id}};
+
+  $self->{name2species_id} = { map{ $_=>$species_id } @{$species2name{$species_id}} };
+  my %species2tax  = $self->species_id2taxonomy($dbi);
+  $self->{taxonomy2species_id} = { map{ $_=>$species_id } @{$species2tax{$species_id}} };
+
+  # process the source files
+  foreach my $file (@{$files}) {
```
mkszuba: Given there are nested loops in this parser, it really wouldn't hurt to label all of them.

misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
```diff
@@ -17,340 +17,330 @@ limitations under the License.
...
-        $self->set_release( $ncrna_source_id,        $release, $dbi );
-        $self->set_release( $pred_mrna_source_id,    $release, $dbi );
-        $self->set_release( $pred_ncrna_source_id,   $release, $dbi );
-        $self->set_release( $pred_peptide_source_id, $release, $dbi );
+    # this will hold the array of xrefs to bulk insert
+    my $xrefs;
+
+    do {
+      local $/ = "\/\/\n";
+      while ( my $genbank_rec = $refseq_fh->getline() ) {
```
mkszuba: Given there are nested loops in this parser, it really wouldn't hurt to label all of them.

misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
```diff
@@ -17,340 +17,330 @@ limitations under the License.
...
+
+    $release =~ s/\s+/ /xg;
+
+    if ( $release =~ m/(NCBI.*Release\s\d+)\s(.*)\sDistribution/x ) {
+      my ($rel_number, $rel_date) = ($1, $2);
+      my $release_string = "$rel_number, $rel_date";
+
+      # set release info
+      $self->set_release( $source_id, $release_string, $dbi );
+      for my $source_name (sort values %{$REFSEQ_SOURCES}) {
```
mkszuba: Does it really matter if we set releases on sources in a specific order? I would say no _i.e._ there is no need to sort the values.

misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
```diff
@@ -17,340 +17,330 @@ limitations under the License.
...
+
+      print "RefSeq release: '$release_string'\n" if $verbose;
+
+    } else {
+      warn "WARNING: Could not set release info from release file '$release_file'\n" if $verbose;
     }
 
-  return 0; # successful
+  } else {
+    warn "WARNING: No release_file available\n" if $verbose;
```
mkszuba: Just get rid of this, a warning that is only output in verbose mode is would be close to useless even if we could enable verbose mode in Hive jobs.

misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
```diff
@@ -17,340 +17,330 @@ limitations under the License.
...
+      # set release info
+      $self->set_release( $source_id, $release_string, $dbi );
+      for my $source_name (sort values %{$REFSEQ_SOURCES}) {
+        $self->set_release( $self->{source_ids}->{$source_name}, $release_string, $dbi );
+      }
+
+      print "RefSeq release: '$release_string'\n" if $verbose;
+
+    } else {
+      warn "WARNING: Could not set release info from release file '$release_file'\n" if $verbose;
```
mkszuba: This whole block will only get executed if a path to the release file is provided, which makes me think that if someone doesn't care about  this not being set they wouldn't provide the release file in the first place. In short, if this code does get executed yet it fails to do what it is supposed to do, abort.

misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
```diff
@@ -17,340 +17,330 @@ limitations under the License.
...
   }
 
-  my @xrefs;
+  my ($description) = $genbank_rec =~ /DEFINITION\s+  # Find the field identifier
+                                  (.*)                # get the description
+                                  \s+ACCESSION/xms;   # until the next field
 
-  local $/ = "\/\/\n";
+  # remove any newlines and spaces, and make sure is within mysql limits
+  $description =~ s/\n//xg;
```
mkszuba: I am fairly confident this line is redundant because \s should include the newline

misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
```diff
@@ -17,340 +17,330 @@ limitations under the License.
...
+  $description =~ s/\s+/ /xg;
```
mkszuba: 1. This IS supposed to simply remove all whitespace, right? Because if it is supposed to replace strings of multiple whitespace characters with a single space, using /x means it does not.

misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
```diff
@@ -17,340 +17,330 @@ limitations under the License.
...
 
-  my $type = $self->type_from_file($file);
-  return unless $type;
+  $xref->{DESCRIPTION} = $description;
 
-  while ( my $entry = $refseq_io->getline() ) {
+  # sequence is multiline, each line starts with base number and has spaces all over. ends with //
+  my ($seq) = $genbank_rec =~ /\s*ORIGIN\s+ # Find the field identifier
+                          (.+)         # get all sequence lines
+                          \/\//xms;    # until the end of the field (//)
```
mkszuba: I would suggest using a different regex terminator - which PDB recommends (specifically, {}) for multi-line patterns anyway - so that you do not have to escape the slashes. More readable that way.

misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
```diff
@@ -17,340 +17,330 @@ limitations under the License.
...
+  # get rid of the base number and the whitespace for a sequence string
+  $seq =~ s/[\d\s]+//xg;
 
-      push @xrefs, $xref if $xref;
+  $xref->{SEQUENCE} = $seq;
 
-  } # while <REFSEQ>
 
-  $refseq_io->close();
+  my @protein_ids = $genbank_rec =~ /\/protein_id=\"(.+?)\"/xg;
```
mkszuba: It isn't really necessary to escape ", is it.

misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
```diff
@@ -17,340 +17,330 @@ limitations under the License.
...
-      my @seq_lines = split /\n/, $seq;
-      my $parsed_seq = "";
-      foreach my $x (@seq_lines) {
-        my ($seq_only) = $x =~ /^\s*\d+\s+(.*)$/;
-        next if (!defined $seq_only);
-        $parsed_seq .= $seq_only;
-      }
-      $parsed_seq =~ s#//##g;    # remove trailing end-of-record character
-      $parsed_seq =~ s#\s##g;    # remove whitespace
+  my @coded_by_list = $genbank_rec =~ /\/coded_by=\"(.*?):/xg;
```
mkszuba: It isn't really necessary to escape ", is it.

misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
```diff
@@ -17,340 +17,330 @@ limitations under the License.
...
+  if (!exists $xref->{PAIR}) {
+    $xref->{PAIR} = $refseq_pair if defined $refseq_pair;
+  }
 
-      ( my $acc_no_ver, $ver ) = split( /\./, $ver );
 
-      $xref->{ACCESSION} = $acc;
-      if($acc eq $acc_no_ver){
-         $xref->{VERSION} = $ver;
+  my @gene_ids = $genbank_rec =~ /db_xref=\"GeneID:(.+?)\"/xg;
```
mkszuba: It isn't really necessary to escape ", is it.

misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
```diff
@@ -17,340 +17,330 @@ limitations under the License.
...
+  @gene_ids = uniq( @gene_ids );
+
+  # process existing entrez_gene_ids as dependent xrefs
+  GENEID:
+  foreach my $gene_id (@gene_ids) {
+
+    next GENEID unless (defined $self->{entrez}->{$gene_id});
+
+    push @{$xref->{DEPENDENT_XREFS}}, {
+        SOURCE_ID         => $self->source_id_from_name('EntrezGene'),
```
mkszuba: You might want to do this hash lookup outside the loop, just in case the Perl interpreter isn't particularly big on optimisation.

misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
```diff
@@ -17,340 +17,330 @@ limitations under the License.
...
+        LINKAGE_SOURCE_ID => $acc_source_id,
+        ACCESSION         => $gene_id,
+        LABEL             => $self->{entrez}->{$gene_id}
+    };
+
+    push @{$xref->{DEPENDENT_XREFS}}, {
+        SOURCE_ID         => $self->source_id_from_name('WikiGene'),
```
mkszuba: Same here.

misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
```diff
@@ -17,340 +17,330 @@ limitations under the License.
...
+    # split the version number
+    my ($pair_acc, $pair_version) = split(/\./x, $refseq_pair);
+
+    # Add xrefs for RefSeq mRNA as well where available
+    foreach my $refseq_acc (@{ $self->{refseq_accs}->{$pair_acc} }) {
+      foreach my $entrez_id (@{ $self->{entrez_ids}->{$gene_id} }) {
+        $self->add_dependent_xref({
+          master_xref_id => $refseq_acc,
+          acc            => $entrez_id,
+          source_id      => $self->source_id_from_name('EntrezGene'),
```
mkszuba: Same...

misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
```diff
@@ -17,340 +17,330 @@ limitations under the License.
...
+          dbi            => $self->{dbi}
+        });
       }
-      else{
-         print "$acc NE $acc_no_ver\n";
+      foreach my $wiki_id (@{ $self->{wiki_ids}->{$gene_id} }) {
+        $self->add_dependent_xref({
+          master_xref_id => $refseq_acc,
+          acc            => $wiki_id,
+          source_id      => $self->source_id_from_name('WikiGene'),
```
mkszuba: ...and here as well.

misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
```diff
@@ -17,340 +17,330 @@ limitations under the License.
...
+        ACCESSION         => $gene_id,
+        LABEL             => $self->{entrez}->{$gene_id}
+    };
+
+    next GENEID unless (defined $refseq_pair);
+
+    # split the version number
+    my ($pair_acc, $pair_version) = split(/\./x, $refseq_pair);
+
+    # Add xrefs for RefSeq mRNA as well where available
```
mkszuba: I trust you have confirmed there is no easy way of having _upload_xref_object_graphs()_ handle these for you?

misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
```diff
@@ -17,340 +17,330 @@ limitations under the License.
...
-                }
-              }
-            }
-          }
-        }
-      }
+  if ( exists $self->{source_ids}->{$name} ) {
+    $source_id = $self->{source_ids}->{$name};
+  } elsif ( $self->{verbose} ) {
+    warn "WARNING: can't get source ID for name '$name'\n";
```
mkszuba: Is this really a non-fatal thing? Feels like it should be an error.

misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
```diff
@@ -17,340 +17,330 @@ limitations under the License.
...
-      # Don't add SGD Xrefs, as they are mapped directly from SGD ftp site
+  my $source_id;
+  my $prefix = substr($acc, 0, 2);
 
-      # Refseq's do not tell whether the mim is for the gene of morbid so ignore for now.
-      return $xref;
+  if ( exists $REFSEQ_SOURCES->{$prefix} ) {
+    $source_id = $self->source_id_from_name( $REFSEQ_SOURCES->{$prefix} );
+  } elsif ( $self->{verbose} ) {
+    warn "WARNING: can't get source ID for accession '$acc'\n";
```
mkszuba: Same.

misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
```diff
@@ -17,340 +17,330 @@ limitations under the License.
...
-# --------------------------------------------------------------------------------
+# get type from filename path. this includes the source name and that's enough to extract it
+# type can be either peptide or dna
+sub type_from_file {
+  my ($self, $file) = @_;
+
+  my ($type) = $file =~ /RefSeq_(peptide|dna)/x;
+
+  if ( $self->{verbose} ) {
+    warn "WARNING: Could not work out sequence type for '$file'\n" unless $type;
```
mkszuba: Same.

misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
```diff
@@ -17,340 +17,330 @@ limitations under the License.
...
-      $parsed_seq =~ s#//##g;    # remove trailing end-of-record character
-      $parsed_seq =~ s#\s##g;    # remove whitespace
+  my @coded_by_list = $genbank_rec =~ /\/coded_by=\"(.*?):/xg;
+
+  my $coded_by = pop @coded_by_list;
+
+  $xref->{PAIR} = $coded_by if defined $coded_by;
+
+  my ($refseq_pair) = $genbank_rec =~ /DBSOURCE\s+REFSEQ: accession (\S+)/x;
+  if (!exists $xref->{PAIR}) {
```
premanand17: Why not combine the conditions with '&&'  (&& defined $refseq_pair)

misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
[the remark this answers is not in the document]
mkszuba: It's a local cache for the BaseParser method you have just mentioned.

misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
```diff
@@ -22,7 +22,7 @@ package XrefParser::RefSeqGPFFParser;
 use strict;
 use warnings;
 use Carp;
-use List::MoreUtils qw(uniq);
+use List::Util qw(uniq);
```
mkszuba: Alas, _uniq_ is NOT in List::Util in case of Perl 5.14.

tgrego:
having a release file as a mandatory argument makes existing wormbase tests fail as they do not provide one.
I've again commented it out and will check what we actually want to do (be mandatory or just nice to have). Inclined for the mandatory option.
@wbazant not sure if you're still following this thread but it affects your code

tgrego:
ok, I've now made my mind about the release file.
It will be optional in this code as it used to be.
However I'll make it mandatory in the ensembl-xref repo.

misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
tgrego: @gene_ids will contain only one gene_id (perhaps ever).
In the old code it actually just picks the first item, I decided to be more conservative and make sure any existing gene_ids get processed because the file format allows more than one to exist.

cpanfile
```diff
@@ -4,6 +4,7 @@ requires 'HTTP::Tiny';
 requires 'IO::Compress::Gzip';
 requires 'URI::Escape';
 requires 'Config::IniFiles';
+requires 'List::Util', '>= 1.45';
```
mkszuba: This should be in the feature 'xref_mapping' section, shouldn't it.

misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
```diff
@@ -17,340 +17,330 @@ limitations under the License.
...
-        $self->set_release( $mrna_source_id,         $release, $dbi );
-        $self->set_release( $ncrna_source_id,        $release, $dbi );
-        $self->set_release( $pred_mrna_source_id,    $release, $dbi );
-        $self->set_release( $pred_ncrna_source_id,   $release, $dbi );
-        $self->set_release( $pred_peptide_source_id, $release, $dbi );
+    # this will hold the array of xrefs to bulk insert
+    my $xrefs;
+
+    do {
+      local $/ = "\/\/\n";
```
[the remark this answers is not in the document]
mkszuba: What's your take on this, @tgrego ?

misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
```diff
@@ -17,340 +17,330 @@ limitations under the License.
...
-my $wiki_source_id;
-my %entrez;
+use List::MoreUtils qw(uniq);
+use Readonly;
 
-sub run {
+use parent qw( XrefParser::BaseParser );
+
+# Refseq sources to consider. Prefixes not in this list will be ignored
+Readonly my $REFSEQ_SOURCES => {
```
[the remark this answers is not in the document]
mkszuba: @tgrego ping

misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
mkszuba: @tgrego ping. Moreover, if it really **is** non-fatal, shouldn't we use carp() instead of warn()?

misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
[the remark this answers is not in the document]
tgrego: will take out the unnecessary escaping

misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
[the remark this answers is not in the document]
tgrego: no particular reason, I just prefer hashrefs

misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
tgrego: same

misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
tgrego: ok, I feel it's safer to actually have it a critical error, so now confesses

misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
tgrego: same

misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
```diff
@@ -17,340 +17,330 @@ limitations under the License.
 
 =cut
 
-# Parse RefSeq GPFF files to create xrefs.
-
 package XrefParser::RefSeqGPFFParser;
```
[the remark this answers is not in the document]
tgrego: will do that for the new repo

misc-scripts/xref_mapping/XrefParser/RefSeqGPFFParser.pm
tgrego: done

cpanfile
tgrego: not sure... it's part of the xref parsing and not actually the mapping...

mkszuba: True but the directory which holds all the xref code is called _xref_mapping_ even though it does more than just mapping, moreover that way it is clear that this is not a dependency of the core API and thus will be safe to remove once the repository split has been completed.

tgrego: done

PR #340, opened 2018-11-27
Moved SQL scripts to https://github.com/Ensembl/ols-ensembl-loader 

marcoooo:
## Requirements

None AFAIK

## Description

Deleted Ontology related scripts, moved to ols-ensembl-loader repository

## Use case

None 

## Benefits

Schema data are in related repo

## Possible Drawbacks

None, AFAIK

## Testing

None

N/A


marcoooo:
Related PR on the other side https://github.com/Ensembl/ols-ensembl-loader/pull/15

magaliruffier:
Looking at the travis failure, this is due to a change in the compara schema and has nothing to do with the update you are proposing
We do however have a test ontology database which is derived from the current schema in misc-scripts/ontology/sql/tables.sql
I am not sure how well we will be able to keep that up-to-date if the schema exists in a separate repository

marcoooo:
> Looking at the travis failure, this is due to a change in the compara schema and has nothing to do with the update you are proposing
> We do however have a test ontology database which is derived from the current schema in misc-scripts/ontology/sql/tables.sql
> I am not sure how well we will be able to keep that up-to-date if the schema exists in a separate repository

Well, good news I didn't break anything with my PR. 
Two solutions here:
- I setup a copy of your test in a perl module inside ols-ensembl-loader, and keep with the schema moving.
- we keep files on your side, and I assume we'll be able to generate scripts for new release patches from our side if needed, but actually, this Db schema won't change a lot, or change all at once :-)

I must say I'd prefer the latter.

marcoooo:
Won't do. keep current Sql layout.

PR #353, opened 2019-01-10
Improve symmetry of store / remove in CoordSystemAdaptor

kiwiroy:
## Description

The internal caches are not accurately maintained for the CoordSystemAdaptor when calling `$csa->remove($cs)`. This is especially a problem for `default` and `sequence level` coordinate systems.

## Use case

```
$cs = $csa->fetch_sequence_level();
$csa->remove($cs);
$cs->name('fix typo');
$csa->store($cs);
```

## Benefits

Completeness, symmetry.

## Possible Drawbacks

Unsure.

## Testing

New test for `$csa->remove($cs)` and side effects included which all pass. No regressions detected.

PR #354, opened 2019-01-10
s/localtime/gmtime/ for successful tests down under.

kiwiroy:
## Description

Fixed tests, I hope, for those in distant, to GMT, timezones.

## Use case

NZ testing on localhost. Specifically, it was difficult to understand consequences when preparing #353.

## Benefits

* Easier to spot regressions.
* More contributions from far off lands?

## Possible Drawbacks

Both `gmtime` and `localtime` have entries in `perlport`.

## Testing

They are the tests.

### DIFF
--- a/modules/t/exon.t
+++ b/modules/t/exon.t
@@ -224,10 +224,10 @@ my $supfeat_count = count_rows($db, 'supporting_feature');
 $exon = $exonad->fetch_by_stable_id('ENSE00000859937');
 
 # check the created and modified times
-my @date_time = localtime( $exon->created_date());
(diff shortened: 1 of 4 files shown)


mkszuba:
@avullo Those failures haven't got anything to do with this change though, have they? This PR only changes a bunch of test suites, which at least in theory should be independent of dbConnection.t.

Moreover, I have already had a look and all three are caused by the same problem: _disconnect_when_inactive()_ apparently not doing what it is supposed to do. Oh, and Travis doesn't run dbConnection.t under Perl 5.14 so it isn't clear if that problem is 5.26-specific or not.,

PR #358, opened 2019-01-10
Added prepare_cached as a wrapper to the DBConnect.pm module

markmcdowall:
## Description

Added a new wrapper function to the DBConnection.pm module for the prepare_cached so that the statements do not need to get prepared multiple times.

Also no trailing whitespace.

## Use case

This should help for loading of data and querying of the database in pipelines where there are multiple requests using the same statement that changes the variables.

## Benefits

Should improve the responsiveness of queries that run multiple times

## Possible Drawbacks

Theoretically none. Would only affect those that decide to use it. It should have no effect on those that do not use it.

## Testing

A new test has been added. Tests should not be run with the 4.050 version of the DBD MySQL driver as there appears to be a wider bug in this code.

### DIFF
--- a/modules/Bio/EnsEMBL/DBSQL/DBConnection.pm
+++ b/modules/Bio/EnsEMBL/DBSQL/DBConnection.pm
@@ -744,7 +744,6 @@ sub prepare_cached {
      throw("Attempting to prepare_cached an empty SQL query.");
    }
 
-   #warn "SQL(".$self->dbname."):" . join(' ', @args) . "\n";
    if ( ( $self->reconnect_when_lost() ) and ( !$self->db_handle()->{Active} ) ) {
       $self->reconnect();
    }
--- a/modules/Bio/EnsEMBL/DBSQL/DBConnection.pm
+++ b/modules/Bio/EnsEMBL/DBSQL/DBConnection.pm
@@ -722,6 +722,13 @@ sub prepare {
   Description: Prepares a SQL statement using the internal DBI database handle
                and returns the DBI statement handle. The prepared statement is
                cached so that it does not have to be prepared again.
+
+               When using this function it is important to ensure either
+               completel retrieval of all information from the request when
+               performing execute() or call finish(). An Active error will be
+               raised if there are currently active handles
+
+               For further information please consult https://metacpan.org/pod/DBI#prepare_cached
   Returntype : DBI statement handle
   Exceptions : thrown if the SQL statement is empty, or if the internal
                database handle is not present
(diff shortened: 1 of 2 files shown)


modules/Bio/EnsEMBL/DBSQL/DBConnection.pm
```diff
@@ -713,6 +713,47 @@ sub prepare {
...
+  Returntype : DBI statement handle
+  Exceptions : thrown if the SQL statement is empty, or if the internal
+               database handle is not present
+  Caller     : Adaptor modules
+  Status     : Experimental
+
+=cut
+
+sub prepare_cached {
+   my ($self,@args) = @_;
```
mkszuba: Not a major readability issue in this case but please add a space after the comma anyway.

modules/Bio/EnsEMBL/DBSQL/DBConnection.pm
```diff
@@ -713,6 +713,47 @@ sub prepare {
...
+
+   if( ! $args[0] ) {
+     throw("Attempting to prepare_cached an empty SQL query.");
+   }
+
+   #warn "SQL(".$self->dbname."):" . join(' ', @args) . "\n";
```
mkszuba: No commented-out code, please. If you want to keep this for future reference, either make this executed conditionally or leave it active here and remove it altogether in another commit.

modules/Bio/EnsEMBL/DBSQL/DBConnection.pm
```diff
@@ -713,6 +713,47 @@ sub prepare {
...
+   if ( ($self->reconnect_when_lost()) and (!$self->db_handle()->ping()) ) {
```
mkszuba: 1. Parts of this might be redundant - as per our conversation a moment ago, calling DBConnection::db_handle() already involves a connection attempt.
2. Have you checked if this actually does anything when _disconnect_when_inactive()_ is enabled?
3. Please check the  attribute _Active_ of the database handle rather than the return value of _ping()_ to check if we are connected to the database, the latter is not guaranteed to give the correct answer - especially for MySQL.

modules/Bio/EnsEMBL/DBSQL/DBConnection.pm
```diff
@@ -713,6 +713,47 @@ sub prepare {
...
+   }
+   my $sth = $self->db_handle->prepare_cached(@args);
+
+   # return an overridden statement handle that provides us with
+   # the means to disconnect inactive statement handles automatically
+   bless $sth, "Bio::EnsEMBL::DBSQL::StatementHandle";
+   $sth->dbc($self);
+   $sth->sql($args[0]);
+
+   $self->query_count($self->query_count()+1);
```
mkszuba: Spaces around the plus sign, please.

modules/Bio/EnsEMBL/DBSQL/DBConnection.pm
```diff
@@ -722,6 +722,13 @@ sub prepare {
   Description: Prepares a SQL statement using the internal DBI database handle
                and returns the DBI statement handle. The prepared statement is
                cached so that it does not have to be prepared again.
+
+               When using this function it is important to ensure either
+               completel retrieval of all information from the request when
```
mkszuba: "That's a misprint, innit?"</round_the_horne>

markmcdowall: This could be changed to something like:

"If not retrieving all rows then `finish()` should be called on the object to free up the statement handle."

nerdstrike: So close... come on Mark, you can do it

PR #360, opened 2019-01-11
dbConnection.t: do not use $dbh->ping() to check if disconnected from DB [release/95]

mkszuba:
## Description

Make _dbConnection.t_ compatible with newly released DBD::mysql-4.050 by using _$dbh->{Active}_ rather than _$dbh->ping()_ to confirm _disconnect_when_inactive()_ works.

## Use case

DBD::mysql::ping() cannot be relied to return _false_ if the client has been disconnected from the database server because it actively attempts to reconnect to the database and only failed to do so in the past because of a bug. With the first version with that bug fixed, 4.050, having been released a couple of days ago, everyone always using the latest release of DBD::mysql (_e.g._ our Travis builds) now see the three _dbConnection.t_ tests aiming at confirming correct operation of disconnect-when-inactive mode fail.

## Benefits

DBD::mysql-4.050 can be used without test failures, which among other things means Travis becoming happy again.

## Possible Drawbacks

None I can think of, DBI documentation says that while the exact meaning of the "Active" attribute is hazy it *is* to be set to false on the database by disconnect() and all drivers I have seen so far do observe this requirement.

## Testing

The change is _in_ unit tests.

Pass.

Yes, no regression detected.

### DIFF
--- a/modules/t/dbConnection.t
+++ b/modules/t/dbConnection.t
@@ -120,9 +120,11 @@ is_deeply($dbc->to_hash(), \%dbc_args, 'Checking to_hash() can roundtrip a DBCon
 #
 my $dbh = $dbc->db_handle();
 
+ok( $dbh->{Active}, 'DB handle is active before enabling disconnect_when_inactive()' );
+
 $dbc->disconnect_when_inactive(1);
 
-ok(!$dbh->ping());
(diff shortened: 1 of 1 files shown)


nerdstrike:
We want this fix into 95, as it will impact anyone in the wild running tests on a fresh install of ensembl.

The tests are failing on copyright only, as it forked before 2019 arrived.

PR #364, opened 2019-02-12
Biotype so_term

tgrego:
## Requirements

## Description

This adds so_term column to the biotype column and to the Biotype objects.
Production master_biotype table needs updating (https://github.com/Ensembl/ensembl-production/pull/218)
Core Features have also been updated to include feature_so_term method, in the likelihood of feature_so_acc. Features of compara, regulation and variation need updating through separate pull requests to follow. 

## Use case

Biotype objects will now report the so_term in addition to the so_acc, thus an ontology database dependency is avoided. Features will also report the so_term.

## Benefits

Improved Biotype objects and no need for ontology database to look up so terms.

## Possible Drawbacks

A bit more data in the biotype table.

## Testing

yes

pass

yes

### DIFF
--- a/modules/t/biotype.t
+++ b/modules/t/biotype.t
@@ -52,19 +52,21 @@ is($biotype1->object_type, 'gene', 'Biotype is from Gene object');
 is($biotype1->name, 'protein_coding', 'Biotype name is protein_coding');
 is($biotype1->biotype_group, 'coding', 'Biotype group is coding');
 is($biotype1->so_acc, 'SO:0001217', 'Biotype protein_coding refers to SO:0001217');
+is($biotype1->so_term, 'protein_coding_gene', 'Biotype protein_coding refers to SO term protein_coding_gene');
 throws_ok { $biotype1->so_acc('test') } qr/so_acc must be a Sequence Ontology accession/, 'so_acc() requires a SO acc like string';
 throws_ok { $biotype1->object_type('test') } qr/object_type must be gene or transcript/, 'object_type() must be gene or transcript';
 
 # test transcript biotype object
 my $transcript = $gene->canonical_transcript;
 debug("transcript biotype");
-is($transcript->biotype, 'protein_coding', "Trancript biotype is protein_coding");
+is($transcript->biotype, 'protein_coding', "Transcript biotype is protein_coding");
 my $biotype2 = $transcript->get_Biotype;
 ok($biotype2->isa("Bio::EnsEMBL::Biotype"), "Biotype object retrieved successfully");
 is($biotype2->object_type, 'transcript', 'Biotype is from Transcript object');
 is($biotype2->name, 'protein_coding', 'Biotype name is protein_coding');
 is($biotype2->biotype_group, 'coding', 'Biotype group is coding');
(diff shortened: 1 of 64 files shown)


tgrego:
it was on purpose, this branch requires the changes from https://github.com/Ensembl/ensembl/pull/363.
Will magically disappear as that branch is merged to master.

mkszuba:
In that case please base this PR against _schema_update_97_ for now, as it is this really is not reviewer-friendly.

modules/t/biotype.t
```diff
@@ -52,19 +52,21 @@ is($biotype1->object_type, 'gene', 'Biotype is from Gene object');
...
 is($biotype1->so_acc, 'SO:0001217', 'Biotype protein_coding refers to SO:0001217');
+is($biotype1->so_term, 'protein_coding_gene', 'Biotype protein_coding refers to SO term protein_coding_gene');
 throws_ok { $biotype1->so_acc('test') } qr/so_acc must be a Sequence Ontology accession/, 'so_acc() requires a SO acc like string';
 throws_ok { $biotype1->object_type('test') } qr/object_type must be gene or transcript/, 'object_type() must be gene or transcript';
 
 # test transcript biotype object
 my $transcript = $gene->canonical_transcript;
 debug("transcript biotype");
-is($transcript->biotype, 'protein_coding', "Trancript biotype is protein_coding");
+is($transcript->biotype, 'protein_coding', "Transcript biotype is protein_coding");
```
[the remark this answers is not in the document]
tgrego: don't want to edit commits just for this tiny thing, will keep attention to in the future commit separately when I spot those small issues.

sql/table.sql
```diff
@@ -2197,6 +2200,7 @@ CREATE TABLE biotype (
   description     TEXT,
   biotype_group   ENUM('coding','pseudogene','snoncoding','lnoncoding','mnoncoding','LRG','undefined','no_group') DEFAULT NULL,
   so_acc          VARCHAR(64),
+  so_term         VARCHAR(1023),
```
markmcdowall: Why are you using such a large number for the varchar? Even though the used space is variable MySQL will have allotted memory units reserved for the storage of the string, which will be much larger in this case. This will reflect badly when sorting or temp tables are required.

Might be worth analysing the required size that is need and restricting what you are storing. Is worth adding a test in the code to ensure that there is not truncation of the inserted text at a later date.

tgrego: That data is present and extracted from the ensembl_ontology_<version> database, term table, name column, which is a VARCHAR(1023).

It could be a TEXT instead, but it's usually better to be a VARCHAR if within it's limits and as you say it will only take as much space as it needs...
The usage is expected to be through the Biotype object, which is a select of a specific biotype name and object_type combination.

markmcdowall: `TEXT` would be the wrong way, if the object does not need to be up to 1023 then limiting down to a smaller number will help MySQL allot the correct memory size.

tgrego: As I said, the source data is 1023 so it can theoretically be that large.
I do think it probably won't be, but looking at the sequenceontology website I can't find any real numbers for how large a term can be.
The only way to play safe is to make it as large as the source data can be.
The table is however quite small, and it's usage limited (and mostly as the select I mentioned).

PR #365, opened 2019-02-13
extract correct dbtype and version across naming conventions

magaliruffier:
## Requirements

## Description

The database identification pattern has been updated to work consistently across databases used in vertebrate and invertebrates.

## Use case

This changes allows the script to be run across vertebrate and invertebrate database and extract the same version, for example extract 96 from homo_sapiens_core_96_38 and vitis_vinifera_core_43_96_3
Prior to that change, it would create two separate lookup databases, one of version 96 for vertebrate databases and one of version 43 for invertebrates.

## Benefits

The script can be run across all divisions and store the data in a single lookup database rather than creating two separate ones.

## Possible Drawbacks

The resulting database can be very big.

## Testing

NA
The modified script was run for release 96 to create a single lookup database to be used by the joint REST server.

NA

NA

### DIFF
--- a/misc-scripts/stable_id_lookup/populate_stable_id_lookup.pl
+++ b/misc-scripts/stable_id_lookup/populate_stable_id_lookup.pl
@@ -224,7 +224,7 @@ sub process_dbs {
 
   foreach my $db (@dbs) {
 
-    if ( $db =~ /([\w\_]+)_(core|otherfeatures)_([\d\_\w]+)/ )
+    if ( $db =~ /([a-z]+_[a-z0-9]+(?:_[a-z0-9]+)?_([a-z]+)(?:_\d+)?)_(\d+)_\w+/ )
     {
       my ( $species, $dbtype, $dbversion ) = ( $1, $2, $3 );
       print "SPECIES: $species\tDBTYPE $dbtype\tDBVERSION: $dbversion\n";


tgrego:
I just has a closer look at the description and can see that the new (correct) db_version is indeed expected.
The species will also change to include the db_type: in the `homo_sapiens_core_96_38` example where it was `homo_sapiens` will now be `homo_sapiens_core`. This one is not described.

magaliruffier:
The regex was run on all the core databases available across all divisions for release 96, so I am fairly confident it picks up the databases we need

misc-scripts/stable_id_lookup/populate_stable_id_lookup.pl
```diff
@@ -224,7 +224,7 @@ sub process_dbs {
 
   foreach my $db (@dbs) {
 
-    if ( $db =~ /([\w\_]+)_(core|otherfeatures)_([\d\_\w]+)/ )
+    if ( $db =~ /([a-z]+_[a-z0-9]+(?:_[a-z0-9]+)?_([a-z]+)(?:_\d+)?)_(\d+)_\w+/ )
```
ens-bwalts: As far as I can tell, this depends on the `like "%$version%"` clause in the show query to work correctly. That said, it parses out the correct things on the database names in e! and e!g databases.

PR #367, opened 2019-02-15
Restore analysis adaptor's ability to store/update web_data

james-monkeyshines:
## Description
When the 'evil eval' for making hashes out of strings was removed from the main adaptor code, it was added to the AnalysisAdaptor, to cater for the last remaining usage in the API, analysis.web_data. At this time the inverse operation was also removed, for stringifying a hash to be inserted or updated in the database. I believe this to be incorrect; there is a test for correct storage of web_data, but due to a bug in the test, it was not working properly; storage in the db was never tested, because data would always be retrieved from a cache instead.

Note that this relates to https://github.com/Ensembl/ensembl/pull/351, and effectively reverts the code changes there; it is not intended (or useful) for an update to delete web_data from the object.

## Use case
As a general principle, 'store' functions should store all of an object's data, rather than silently omit bits. In particular, EG and Production pipelines extract analysis data from the production db and insert it into core dbs, via the API - without this fix, the web_data is always set to NULL, which causes problems down the road. This is mitigated for some teams by the practice of synchronising production and core dbs, but others (VectorBase, WormBase) do not do this.
I think it is better practice for pipelines to add all of the analysis data when the relevant pipeline is run, rather than relying on a script to be run at a later date to synchronise from the production db; this change supports that practice.

## Benefits
'store' and 'update' methods (and consequently EG and Production pipelines) work as expected.

## Possible Drawbacks
None I can think of.

## Testing
Yes

Pass

Yes

### DIFF
--- a/modules/Bio/EnsEMBL/DBSQL/AnalysisAdaptor.pm
+++ b/modules/Bio/EnsEMBL/DBSQL/AnalysisAdaptor.pm
@@ -62,6 +62,8 @@ use Bio::EnsEMBL::Analysis;
 use Bio::EnsEMBL::DBSQL::BaseAdaptor;
 use Bio::EnsEMBL::Utils::Exception;
 
+use Data::Dumper;
+$Data::Dumper::Terse = 1;
 
 use vars qw(@ISA);
 use strict;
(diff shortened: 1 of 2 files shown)


james-monkeyshines:
Ah, looks like the perl critic test wasn't run locally when I did the full test suite, because an environment variable wasn't set. Will have a look at this next week...

markmcdowall:
@mkszuba would you mind having a look as someone from the infrastructure side?

Cheers

PR #368, opened 2019-02-21
Utils IO into ensembl-io

tgrego:
## Requirements

## Description

Related to the the pull request https://github.com/Ensembl/ensembl-io/pull/79.
IO code, and specially the reading/writing of bioinformatics file formats, should be contained in the same repo.
Thus the code `Bio::EnsEMBL::Utils::IO::*` has been moved to the ensembl-io repo.

ensembl-io has been added to the travis builds, but it is not a requirement of the core ensembl repo.
A single test requires it (fastaSequenceAdaptor.t) and is skipped if it's not present.

## Use case

Most users of core will not notice this, those who use `Bio::EnsEMBL::Utils::IO::*` will need to add ensembl-io as a dependency.

## Benefits

Keep IO code in the same place, make core ensembl repo cleaner.

## Possible Drawbacks

Extra dependency for some users.

## Testing

yes

yes

yes


markmcdowall:
@avullo @mkszuba 

Changes seem reasonable, if you could have a look that would be great.

Are there any other places in ensembl core work where we might need to be aware of these changes (other than ensembl-io)?

Cheers

tgrego:
I can only think of ensembl-rest, but ensembl-io is already a dependency so the switch will go unnoticed.

tgrego:
utils/io code predates ensembl-io. It is not core API related code, but io related code.
ideally it would be deprecated and superseded by the newer ensembl-io code. However it has been found that ensembl-io code is not a replacement for utils/io in all circumstances, due to being considerably slower.
This is compromise: cleanup the core API repo and get IO code in a consistent repo (even being 2 different codebases) making sure outputs are consistent where possible.

avullo:
Sure, I see that.

muffato:
We use `FASTASerializer` and `GFFParser`. I can see that Production, Regulation, ensembl-analysis, ensembl-tools use several serializers too. If you merge this pull-request, that's additional dev time we urgently need to schedule before e97. If the interface of those objects is exactly the same (i.e. we only need to add a checkout of ensembl-io to the environment and change the package names) that's quite straightforward, but still because it impacts so many teams, I think this change should rather be declared as an intention, discussed with the other teams and then scheduled for a certain release.

tgrego:
All that will be required is to have ensembl-io as an added dependency.
Interfaces and namespaces will not change.

muffato:
OK, fair enough, but please announce your change to internal@. Not all the teams may be using ensembl-io (on the right branch)

modules/t/fastaSequenceAdaptor.t
```diff
@@ -25,11 +25,19 @@ use Bio::EnsEMBL::Test::MultiTestDB;
...
 use Bio::EnsEMBL::DBSQL::FastaSequenceAdaptor;
-use Bio::EnsEMBL::Utils::IO::FileFaidx;
 use Bio::EnsEMBL::Utils::IO qw/work_with_file/;
 
+# Check if ensembl-io repo is available
+eval {
+  require Bio::EnsEMBL::Utils::IO::FASTASerializer;
+  require Bio::EnsEMBL::Utils::IO::FileFaidx;
+};
+if ($@) {
```
mkszuba: This might work as expected in case of _require_ but as a rule, the reliable way of telling whether an eval has failed is to look at its return value and only use _$@_ to propagate the error message, if any.

travisci/harness.sh
```diff
@@ -2,7 +2,7 @@
 
 ENSDIR="${ENSDIR:-$PWD}"
 
-export PERL5LIB=$ENSDIR/bioperl-live:$ENSDIR/ensembl-test/modules:$PWD/modules:$ENSDIR/ensembl-variation/modules:$ENSDIR/ensembl-compara/modules:$PWD/misc-scripts/xref_mapping
+export PERL5LIB=$ENSDIR/bioperl-live:$ENSDIR/ensembl-test/modules:$PWD/modules:$ENSDIR/ensembl-io/modules:$PWD/modules:$ENSDIR/ensembl-variation/modules:$ENSDIR/ensembl-compara/modules:$PWD/misc-scripts/xref_mapping
```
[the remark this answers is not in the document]
tgrego: fixed, thanks for noticing

modules/t/fastaSequenceAdaptor.t
tgrego: I used the 'standard' ensembl way (slice.t, schemaPatches.t, housecleaning_perlCritic.t and circularSlice.t all use it)

PR #372, opened 2019-04-01
Feature/stop codon readthrough edit

s-mm:
## Description

ENSCORESW-3110- Feature to handle stop codon readthrough in amino acid sequence.

## Use case

At the moment, for a set of 12 transcripts, the stop codon readthrough is identified by an asterisk. This feature will modify the amino acid sequence to display the stop codon readthrough with an 'X'.

## Benefits

This will bring stop codon readthrough representation in line with the stop codon readthrough representation in RefSeq.

## Possible Drawbacks

None.

## Testing
One dummy test added in translation.t which executes successfully. An actual test case cannot be added at the moment as sequences containing asterisk representation for stop codon readthrough is unavailable in core database. This is available in loutre database.


mkszuba:
@s-mm Let us wait until #374 has been merged so that Travis (hopefully) stops failing on unrelated problems. Or you could merge _bugfix/e97_testdb_patches_master_ into your branch once (if?) the Travis build for #374 has succeeded.

mkszuba:
I have just changed the PR base to _bugfix/e97_testdb_patches_master_ to make it easier to see what it actually changes. Note to whoever ends up merging this: **please change the base back to master beforehand**.

modules/t/translation.t
```diff
@@ -343,4 +343,28 @@ ok(!scalar(@alt_tls));
...
+  -END     => 2,
+  -ALT_SEQ => 'X',
+  -CODE    => '_stop_codon_readthrough',
+  -NAME    => 'Stop Codon Readthrough'
+);
+
+$tr_scrt->translation->add_Attributes($scrt->get_Attribute());
+
+my $tlseq_scrt = $tr_scrt->translate->seq();
+debug("X inserted: " . $tlseq_scrt);
```
mkszuba: Please use _diag()_ or _note()_ (see the "Diagnostics" section of Test::More documentation) instead. Yes, I know we have got calls to _debug()_ all over this test file.

modules/t/translation.t
```diff
@@ -343,4 +343,28 @@ ok(!scalar(@alt_tls));
...
+is($tlseq_scrt =~ /X/, 1, 'X inserted');
```
mkszuba: One more thing: it wouldn't hurt to extend the pattern used in this regex to cover some neighbouring polypeptides. It won't be that much more expensive computationally and that way the test will be a bit more strict than just "is there an X in the sequence now".

modules/Bio/EnsEMBL/StopCodonReadthroughEdit.pm
```diff
@@ -0,0 +1,83 @@
...
+
+  A representation of a stop codon readthrough edit in a sequence. 
+
+=head1 METHODS
+
+=cut
+
+package Bio::EnsEMBL::StopCodonReadthroughEdit;
+
+use strict;
```
mkszuba: I would do 'use warnings' here too, to be on the safe side.

modules/Bio/EnsEMBL/StopCodonReadthroughEdit.pm
```diff
@@ -0,0 +1,83 @@
...
+
+use parent qw(Bio::EnsEMBL::SeqEdit);
+
+=head2 new
+
+  Arg [-START/END]  : 
+       int - start and end postion of the stop codon readthrough edit in the sequence
```
mkszuba: This isn't currently used, is it. That said, I *would* recommend using a named argument for passing the coordinate to the constructor - it's the way it is generally done in Ensembl and it makes the code easier to read.

As for what this argument should be named, something like POSITION or COORDINATE makes sense from the logical point of view but I would ask _e.g._ HAVANA people what term they nomally use.

modules/Bio/EnsEMBL/StopCodonReadthroughEdit.pm
```diff
@@ -0,0 +1,83 @@
...
+  Returntype : Bio::EnsEMBL::StopCodonReadthroughEdit
+  Exceptions : none
+  Caller     : general
+  Status     : Stable
+
+=cut
+
+sub new {
+  my ($self, $coord) = @_;
+  my $caller = shift;
```
mkszuba: This unfortunately is wrong, $caller should be assigned the first argument passed to new() but the way the code is right now it is passed the third, which will be _undef_. Finally, you have got _$self_ declared both here and in line 72.

The simplest way of fixing this would be to change _$self_ to _$caller_ in line 68 and get rid of line 69.

modules/Bio/EnsEMBL/StopCodonReadthroughEdit.pm
```diff
@@ -0,0 +1,83 @@
...
+   
+  my $class = ref($caller) || $caller;
+  my $self = $class->SUPER::new(        
+        -START   => $coord,
+        -END     => $coord,
+        -ALT_SEQ => 'X',
+        -CODE    => '_stop_codon_readthrough',
+        -NAME    => 'Stop Codon Readthrough');
```
mkszuba: Something I missed the previous time: is it really necessary to declare both the code and the name? If I remember correctly how AttributeAdaptor works, one only needs the code and the rest gets filled by the corresponding database entry.

And speaking of database entry, declaring a new attribute type requires adding a new row to the table _attrib_type_. For the release this will have to be done by Production in their metadata database (which by the way means almost for sure that you should declare an intention for this, if you haven't yet) but I would very much recommend already adding it to the test database, just so that it is harder to forget. By the way, this is likely one of the things a non-dummy test would have caught.

modules/Bio/EnsEMBL/StopCodonReadthroughEdit.pm
```diff
@@ -0,0 +1,82 @@
...
+
+  Bio::EnsEMBL::StopCodonReadthroughEdit - Object representing a stop codon readthrough edit in a sequence
+
+=head1 SYNOPSIS
+
+  my $stopcodonrtedit = Bio::EnsEMBL::StopCodonReadthroughEdit->new($coord);
+
+=head1 DESCRIPTION
+
+  A representation of a stop codon readthrough edit in a sequence.
```
nerdstrike: This description is useless to the reader. I'd suggest making the Synopsis better, i.e. instantiate $coord, then call some accessors on the object or store it with an adaptor or _something_.

Perhaps you could explain what a readthrough edit is? Who edited it?

modules/Bio/EnsEMBL/StopCodonReadthroughEdit.pm
```diff
@@ -0,0 +1,82 @@
...
+  Caller     : general
+  Status     : Stable
+
+=cut
+
+sub new {
+  my ($self, $position) = @_;
+
+  my $class = ref($self) || $self;
+  my $stopcodonrtedit = $class->SUPER::new(
```
nerdstrike: Historically we have made compound variable names using _ to make them legible, e.g. stop_codon_rt_edit

modules/t/translation.t
```diff
@@ -343,4 +344,23 @@ ok(!scalar(@alt_tls));
 # Test generic_count(), inherited method from BaseAdaptor
 is($ta->generic_count(), @{$ta->list_dbIDs()}, "Number of features from generic_count is equal to the number of dbIDs from list_dbIDs");
 
+#56-57
+#
+# Handling stop codon readthrough edit
+#
+
+my $tra_scrt = $db->get_TranscriptAdaptor();
```
nerdstrike: You'll make more friends if you name variables less cryptically. This is a transcript adaptor. $adaptor, $transcript_adaptor, and $tra would all be less cryptic.

modules/Bio/EnsEMBL/StopCodonReadthroughEdit.pm
mkszuba: Not just historically, this is explicitly mentioned in our internal style guide.

mkszuba:
...and I have just realised the attribute code introduced in this PR is not valid because the type of _attrib_type.code_ is VARCHAR(20). Have let @s-mm know this will have to be fixed. I wonder why the test harness doesn't fail on that new line in _attrib_type.txt_; another fun property of MyISAM?

nerdstrike:
I think MySQL silently truncates unless it is configured to barf.

PR #375, opened 2019-04-04
[ENSCORESW-3077] Travis: Unify configuration of Slack notifications

mkszuba:
Use the same Slack access token as well as the same notification configuration for all Infrastructure repositories.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ language: perl
 os:
   - linux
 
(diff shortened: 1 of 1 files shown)


ens-bwalts:
Tests on pushed branch failing due to missing DB patches. Tests against PR branch, which includes DB patches are passing. Will go ahead and merge.

PR #379, opened 2019-04-29
Registry: add support for versioned ensembl_metadata DBs [master]

mkszuba:
## Description

Extend the Registry so that it can detect Ensembl metadata databases whose names contain the release number, with backward compatibility with unversioned databases but preference for versioned ones if both forms are present.

## Use case

Since release 96 Ensembl metadata databases have been named _ensembl_metadata_XY_, where _XY_ is the release number. Unfortunately the Registry has not been updated accordingly so it keeps looking for _ensembl_metadata_. See ENSCORESW-3119 for additional information.

## Benefits

Registry will be able to correctly instantiate MetadataDBAdaptors again.

## Possible Drawbacks

A slight increase in computational overhead during Registry initialisation, primarily due to the fact we retain backward compatibility with the old naming convention.

## Testing

No.

N/A

Yes, no regression detected.


modules/Bio/EnsEMBL/Registry.pm
```diff
@@ -1818,6 +1820,12 @@ sub load_registry_from_db {
         $taxonomy_db      = $db;
     } elsif ( $db =~ /^ensembl_metadata$/ ) {
         $ensembl_metadata_db      = $db;
+    }
+    elsif ( $db =~ m{ \A ensembl_metadata_(\d+) \z }msx ) {
+      if ( $1 eq $software_version ) {
+        $ensembl_metadata_db_versioned = $db;
+        $ensembl_metadata_version      = $1;
```
ens-bwalts: `$ensembl_metadata_version` is only used elsewhere to indicate that a versioned metadata db has been found. Were you intending to use it elsewhere, or is this line to help clarify the regex? Otherwise, line 2360 and line 2371 could be updated to check for 
`( defined $ensembl_metadata_db_versioned ) ` with the same functionality.

PR #384, opened 2019-05-14
ENSCORESW-3147: correctly capture all required fields from file

magaliruffier:
## Requirements

## Description

The RFAM parser does not capture all required fields when run as part of the eHive xref pipeline. This is because the file is read line by line from disk, while the information for a single entry is split across multiple lines.
The proposed change groups the lines in logical blocks representing a single RFAM entry.

## Use case

Since the move to the eHive-based xref pipeline, all RFAM xrefs are missing descriptions and their label is identical to the accession when it should not be.
The proposed fix ensures both label and description are stored in the database and can be displayed on the browser.
This was reported by a user who compared the same gene across two zebrafish assemblies, see
http://apr2018.archive.ensembl.org/Danio_rerio/Transcript/Summary?db=core;g=ENSDARG00000082665;r=25:8247670-8247787;t=ENSDART00000116947
versus
http://dec2017.archive.ensembl.org/Danio_rerio/Transcript/Summary?db=core;g=ENSDARG00000082665;r=25:8247670-8247787;t=ENSDART00000116947

## Benefits

Useful information from RFAM is correctly stored and displayed to our users.

## Possible Drawbacks

The proposed fix is not the prettiest and results in storing multiple lines in memory. However, the RFAM file is small so this should not have a major impact on memory requirements.

## Testing

NA, the current xref pipeline does not have unit tests.
The pipeline was run with and without the fix though to ensure the data is captured correctly.

NA

NA

### DIFF
--- a/misc-scripts/xref_mapping/XrefParser/RFAMParser.pm
+++ b/misc-scripts/xref_mapping/XrefParser/RFAMParser.pm
@@ -141,11 +141,17 @@ sub run_script {
   } else {
     my $file_io = $self->get_filehandle($file);
     if ( !defined $file_io ) {
-      print "ERROR: Can't open HGNC file $file\n";
+      print "ERROR: Can't open RFAM file $file\n";
       return 1;
     }
+    my $temp_line;
     while (my $line = $file_io->getline()) {
-      push(@lines, $line); 
+      if ($line =~ /\/\//) {
+        push @lines, $temp_line;
+        $temp_line = '';
+      } else {
+        $temp_line .= $line . "\n";
+      }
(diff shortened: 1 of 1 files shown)


magaliruffier:
Travis failure is on sliceVariation.t and strainSlice.t, both seem to be related to changes in the variation schema
This will require patching the test databases but should not be related to changes in this PR

magaliruffier:
You're right, sorry about that. I branched off master but was not on the latest version.
It looks like there might be some mismatching commits (did we remove some commits from master), so I will create a new branch

PR #385, opened 2019-05-14
Make the Coveralls badge in the README work

mkszuba:
## Description

Update Coveralls URIs in README.md so to get the badge working.

## Use case

Our Coveralls badge as it is keeps saying "unknown" because it doesn't specify the branch in the URI, which apparently is necessary in case of Coveralls (and recommended for Travis as well). Moreover, the URI itself might be outdated.

## Benefits

Actually get the badge to work.

## Possible Drawbacks

Keeping it working for release branches will require updating the README as part of the release procedure.

## Testing

No.

N/A

No, this is a change to documentation - and a quantitative one at that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Ensembl Core API
 
-[![Build Status](https://travis-ci.org/Ensembl/ensembl.png?branch=master)][travis]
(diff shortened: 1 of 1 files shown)


tgrego:
Actually the `?branch=master` at line 7 is not needed and is perhaps best removed.
That way there is only one place to update the branch at branching and the link always stays the same.

PR #391, opened 2019-06-07
Remove deprecated GOTermAdaptor and SOTermAdaptor modules

mkszuba:
## Description

[ENSCORESW-2871] Remove deprecated GOTermAdaptor and SOTermAdaptor modules.

## Use case

These adaptors have been deprecated since 2011 and non-functional due to having been removed from the list in OntologyDBAdaptor since March 2018.

## Benefits

Less dead code in the repository, less confusing to users looking at the DBSQL/ directory to see what adaptors are available.

## Possible Drawbacks

None.

## Testing

No.

N/A

Yes, no regression observed.


modules/Bio/EnsEMBL/DBSQL/SOTermAdaptor.pm
```diff
@@ -1,77 +0,0 @@
...
-A specialization of Bio::EnsEMBL::DBSQL::OntologyTermAdaptor,
-specifically for Sequence ontology (SO) terms.  See the
-documentation of Bio::EnsEMBL::DBSQL::OntologyTermAdaptor for
-further information.
-
-=head1 METHODS
-
-=cut
-
-package Bio::EnsEMBL::DBSQL::SOTermAdaptor;
```
ens-bwalts: There's a reference to SOTermAdaptor in ensembl-variation. However, I'm not seeing that method ever gets called anywhere. For that matter, I'm not sure that method still works (for example, where does $self->{adaptor} ever get set in that class as currently implemented?). Have we gotten an OK or don't care from Variation on this?

https://github.com/Ensembl/ensembl-variation/blob/396640b5e0f5e03f2ddfa863948f291a1e59eb07/modules/Bio/EnsEMBL/Variation/OverlapConsequence.pm#L493

mkszuba: > Have we gotten an OK or don't care from Variation on this?

I would say we don't care, seeing as technically speaking that call should have been non-functional since March 2018. Still, I've just pinged them about it on Slack - just in case.

mkszuba: Variation have just confirmed that the function in question is no longer in use (and they'll remove it now).

PR #393, opened 2019-06-17
ENSCORESW-3132: Limited implementation of Intervals with start > end …

ens-bwalts:
## Requirements

## Description

Changes to Bio::EnsEMBL::Utils::Interval allowing intervals with a start > end (e.g. spanning the origin of a circular chromosome). This is a quick, simple implementation that allows origin-spanning intervals to be created without throwing an exception. These intervals will identify themselves as origin-spanning, and will react sensibly if queried for overlap or relative position.

Note that the current implementation of IntevalTrees cannot handle an origin-spanning Interval. The current short-term solution is to throw an exception when an origin-spanning interval is put into one of these trees. This is sub-optimal, and the IntervalTree implementations should be updated to handle origin-spanning Intervals with urgency.

## Use case

The new Mapper implementation (see PR #332 ) uses Intervals and Interval Trees that do not handle origin-spanning intervals. It is not possible to even create such an interval, so several scripts that load origin-spanning features have been breaking as a result. Although this fix is quite basic, at least features with start > end can be loaded and worked with in a limited way.

## Benefits

Features with a start > end can be handled in a limited way again

## Possible Drawbacks

This is a very quick and simple implementation, and focuses mostly on Intervals, not how they are handled by IntervalTrees. As a result, although origin-spanning intervals can now be created and subjected to basic queries, the IntervalTrees that handle more complex interval operations are still not able to handle these intervals. The short-term solution is to throw exceptions when an origin-spanning interval is an operand in an operation that the current IntervalTree implementations cannot handle. The IntervalTree implementations will need to be improved to handle more operations with origin-spanning Intervals soon.

Note that the C implementation of IntervalTreeMutable (called through XS) will need to be updated in a separate PR.

## Testing

Yes

Pass

Yes

### DIFF
--- a/modules/Bio/EnsEMBL/Utils/Interval.pm
+++ b/modules/Bio/EnsEMBL/Utils/Interval.pm
@@ -35,7 +35,7 @@ Bio::EnsEMBL::Utils::Interval
 =head1 SYNOPSIS
 
   # let's get an interval spanning 9e5 bp and associated it with some data
-  my $i2 = Bio::EnsEMBL::Utils::Interval->new(1e5, 1e6, { 'key1' => 'value1', 'key2' => 'value2' });
+  my $i1 = Bio::EnsEMBL::Utils::Interval->new(1e5, 1e6, { 'key1' => 'value1', 'key2' => 'value2' });
 
   # and another one which overlaps with the previous,
   # but with scalar associated data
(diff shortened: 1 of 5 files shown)


modules/Bio/EnsEMBL/Utils/Interval.pm
```diff
@@ -159,7 +193,11 @@ sub data {
 sub is_empty {
   my $self = shift;
 
-  return $self->start >= $self->end;
+  if ($self->spans_origin) {
```
muffato: By definition of `spans_origin`, we have `$start > $end` in the_then_ block, so it will always return 0. And in the _else_ block `$start <= $end`, so it will only return 1 if `$start -- $end`, i.e. if the interval is a "point"

PR #398, opened 2019-06-21
E99 version bump

mkszuba:
## Description

Bump the API and the schema version to 99. Patch the test databases accordingly (includes the master-branch equivalent of #397. Tell Travis to have the Registry shut up about version mismatches between the API and the databases so that we can still get meaningful test results in spite of Compara, Variation and ontology database schemas remaining at 98.

## Use case

According to the revised Core release SOP we want to do N+1 version bumps right after branching _release/N_. For the current release cycle, the relevant Jira ticket is ENSCORESW-3163.

## Benefits

Early version bumping will make it clear that what we have on the master branch is not necessarily the same API/schema as on even the latest release branch. Moreover, it will give us more time to catch and address possible issues with release 100.

## Possible Drawbacks

The code-branching step if the release schedule will be slightly more complicated because it will be necessary to reenable version-mismatch warnings in .travis.yml before pushing a new release branch.

## Testing

No.

N/A

Have run the entire test suite with version-mismatch warnings both enabled and disabled. In the former case _ontologyTerm.t_ and _sequenceOntologyMapper.t_ fail due to having produced unexpected warnings, in the latter case everything works.


.travis.yml
```diff
@@ -40,6 +40,10 @@ install:
   - cp travisci/MultiTestDB.conf.travisci.SQLite modules/t/MultiTestDB.conf.SQLite
   - mysql -u root -h localhost -e 'GRANT ALL PRIVILEGES ON *.* TO "travis"@"%"'
 
+before_script:
+  # This should probably be removed on release branches
```
ens-bwalts: I'll be laying awake for an extra 30 seconds tonight worrying about how this opens us up to a risk of false negatives in testing. Having removal of this line in the release checklist mitigates this risk. It would be nice if there were some way to have harness.sh check for a release mismatch in the test databases from variation, etc. and conclude that we are on an n+1 branch, but at the moment I'm not sure about how to do that sensibly and not be at least as much of a pain as remembering to delete this line.

mkszuba: I have already noted the need to disable creation of this file on release branches in Core Release Procedure, just in case. As for automation, one could always consider the value of $TRAVIS_BRANCH to decide whether to disable version checking or not - but it should be done carefully because we've got development/PR branches to account for in addition to _master_ and release branches.

ens-bwalts: Sorry, I wasn't clear -- should have said that I saw you added that item to the release checklist, which takes care of the issue. Thinking about the issue a bit more, I'm not sure, given how many other test DBs this needs to interact with, that a systematic approach would ever be possible.

PR #402, opened 2019-07-25
Prevent ChecksumParser from losing one xref per input file

mkszuba:
## Description

Shift the numbering of checksum xrefs to be inserted by ChecksumParser by one. Also, check if the temporary output file has successfully been opened before attempting to use it.

## Use case

The way ChecksumParser calculates values of _checksum_xref_id_ now causes it to lose one such xref per input file; see the commit message for details. May or may not be related to ENSCORESW-3196.

## Benefits

Will stop losing RNAcentral and UniParc xrefs.

## Possible Drawbacks

None.

## Testing

No.

N/A

No, ChecksumParser is not covered by the test suite.

### DIFF
--- a/misc-scripts/xref_mapping/XrefParser/ChecksumParser.pm
+++ b/misc-scripts/xref_mapping/XrefParser/ChecksumParser.pm
@@ -47,6 +47,7 @@ sub run {
   }
   $verbose ||=0;
 
+  # FIXME: this will fail if the input file is in a read-only directory (ENSCORESW-3197)
   my $target_file = $files->[0].'.mysqlinput';
   my $input_fh = $self->get_filehandle($files->[0]);
   if(-f $target_file) {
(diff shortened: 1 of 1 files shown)


misc-scripts/xref_mapping/XrefParser/ChecksumParser.pm
```diff
@@ -71,7 +71,7 @@ sub _transfer_contents {
   while(my $line = <$input_fh>) {
     chomp $line;
     my ($upi, $checksum) = split(/\s+/, $line);
-    my @output = ($counter++, $source_id, $upi, $checksum);
+    my @output = (++$counter, $source_id, $upi, $checksum);
```
ens-bwalts: I'm not a fan of using ++ except when it's the only thing happening on a line or in cases like for loops where it's idiomatic. It's hard to read and can lead to subtle bugs like the one that's fixed here. Would prefer it added standalone above. However, as this is three commits down in the history will leave it up to you to change or not.

mkszuba: At first I wanted to avoid changing the code more than necessary - but heck, I don't like in-line increments either and we might as well do it properly. Done.

PR #405, opened 2019-08-08
Backport changes to the xref-parser test framework from ensembl-xref

mkszuba:
## Description

Backport changes to the xref unit-test framework from _ensembl-xref_ to the main repository.

## Use case

See ENSCORESW-3234. Although the first iteration of the test framework developed by @nerdstrike for the purpose of actually letting us unit-test xref parsers was committed to _ensembl_, it has since been modified in _ensembl-xref_.

## Benefits

We will continue to be able to run xref-parser unit tests.

## Possible Drawbacks

We will use even more Travis CPU time.

## Testing

Yes, the two unit-test files already present in _misc-scripts/xref_mapping_ have been updated accordingly.

They pass.

Yes, everything passes. For the record, at least on my system running the main test suite automatically picks up tests from the xref directory.


mkszuba:
> I see that linting won over being able to regenerate the schema from a modified DB. That's fine I, but it might increase future "fun".

*Could* we actually regenerate the schema? IMHO it would be good to be able to do so but I vaguely remember that you had some difficulties making that work due to, ahem, certain peculiarities of the xref schema.

> Are you certain you want this in the core repo? I had expected the xref pipeline to live apart from the core ORM, since it does not use core much at all.

We do want to ultimately move it out, yes. That said, I think it will be safer to do this in stages: achieve working state in the core repo, archive _ensembl-xref_, then create a new repository for this code and update the pipeline in _ensembl-production_ accordingly.

 The only point of contact is in the upload at the end, and I don't think it uses the API for that. It also adds additional testing dependencies to the most basic component of Ensembl. In the past that would have been a no-no.

PR #407, opened 2019-08-09
Fix fetch_all_by_outward_search

AsierGonzalez:
## Description

Fixed _fetch_all_by_outward_search()_ to expand the search window in increments of the RANGE value. IN the current implementation the search range is increased exponentially instead of linearly. This has now been fixed by storing the initial RANGE value.

## Use case
Currently, calling this function with RANGE=10,000 and MAX_RANGE=500,000 will do the search in the following ranges: 10000, 20000, 60000, 240000 and 1200000. This pull request changes this behaviour to search in ranges from 10000 to 50000 in increments of 10000.

## Testing

Have you added/modified unit tests to test the changes?

Yes.

If so, do the tests pass/fail?

They pass.

### DIFF
--- a/modules/Bio/EnsEMBL/DBSQL/BaseFeatureAdaptor.pm
+++ b/modules/Bio/EnsEMBL/DBSQL/BaseFeatureAdaptor.pm
@@ -1440,18 +1440,19 @@ sub fetch_nearest_by_Feature {
 
 sub fetch_all_by_outward_search {
   my $self = shift;
-   my ($ref_feature, $respect_strand, $opposite_strand, $downstream, $upstream, $search_range,
+  my ($ref_feature, $respect_strand, $opposite_strand, $downstream, $upstream, $start_search_range,
        $limit,$not_overlapping,$five_prime,$three_prime, $max_range) =
         rearrange([qw(FEATURE SAME_STRAND OPPOSITE_STRAND DOWNSTREAM UPSTREAM RANGE LIMIT NOT_OVERLAPPING FIVE_PRIME THREE_PRIME MAX_RANGE)], @_);
   my $factor = 1;
   $limit ||= 1;
-  $search_range ||= 1000;
+  $start_search_range ||= 1000;
+  my $current_search_range = $start_search_range;
   $max_range ||= 10000;
   my @results;
-  while (scalar @results < $limit && $search_range <= $max_range) {
-    $search_range = $search_range * $factor;
+  while (scalar @results < $limit && $current_search_range <= $max_range) {
+    $current_search_range = $start_search_range * $factor;
     @results = @{ 
-      $self->fetch_all_nearest_by_Feature(-RANGE => $search_range, 
+      $self->fetch_all_nearest_by_Feature(-RANGE => $current_search_range,
                                           -FEATURE => $ref_feature, 
                                           -SAME_STRAND => $respect_strand, 
                                           -OPPOSITE_STRAND => $opposite_strand,
--- a/modules/Bio/EnsEMBL/DBSQL/BaseFeatureAdaptor.pm
+++ b/modules/Bio/EnsEMBL/DBSQL/BaseFeatureAdaptor.pm
@@ -1449,9 +1449,9 @@ sub fetch_all_by_outward_search {
   my $current_search_range = $start_search_range;
   $max_range ||= 10000;
   my @results;
-  while (scalar @results < $limit && $current_search_range <= $max_range) {
+  while (scalar @results < $limit && $current_search_range < $max_range) {
     $current_search_range = $start_search_range * $factor;
-    @results = @{ 
+    @results = @{
       $self->fetch_all_nearest_by_Feature(-RANGE => $current_search_range,
                                           -FEATURE => $ref_feature, 
                                           -SAME_STRAND => $respect_strand, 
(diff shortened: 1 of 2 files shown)


AsierGonzalez:
Hi @mkszuba,
Thank you for your feedback. I did see that the template mentioned the documentation of the testing but I didn't really know what I should put there. I only did the very simple test of running the old and the new version of the function with the same variants and print the size of the search window and the output. Would that be enough?

mkszuba:
It's a start but we would very much prefer if you wrapped this into a unit test which could be run as part of the _ensembl_ test suite. Moreover, running the whole test suite to confirm the change doesn't cause any side effects elsewhere in the code is pretty much a must.

In this particular case, I see there is a test involving _fetch_all_by_outward_search()_ present in _getNearestFeature.t_ which I am fairly sure should now fail.

AsierGonzalez:
What you say sounds sensible but I am afraid that I am not familiar with Perl testing so I would appreciate any links to Ensembl docs (or other sources) where it is explained how to run the test suite and how to add new tests. If I have that information I can add a test to check that the function does not return features outside the search range, which would fail in the current implementation. The [current test](https://github.com/Ensembl/ensembl/blob/release/97/modules/t/getNearestFeature.t#L181-L182) passes for both the old and new implementation because there is a feature in the limit of the search range.
I am also confused about one of the two Travis builds failing. It seems that the failure is due to issues installing Perl v5.14 so I believe this is not related to the changes I made.

mkszuba:
> What you say sounds sensible but I am afraid that I am not familiar with Perl testing so I would appreciate any links to Ensembl docs (or other sources) where it is explained how to run the test suite and how to add new tests. If I have that information I can add a test to check that the function does not return features outside the search range, which would fail in the current implementation. 

Excellent. Documentation for setting up the test environment can be found [here](https://github.com/Ensembl/ensembl-test/blob/master/scripts/README), one thing that might be worth adding is that for running a single test file you can use the command _prove_ which should have come with Perl (something like _prove -v ensembl/modules/t/getNearestFeature.t_). Plus of course there is Travis, whenever it gets to running another build - we tend to keep it rather busy.

> The [current test](https://github.com/Ensembl/ensembl/blob/release/97/modules/t/getNearestFeature.t#L181-L182) passes for both the old and new implementation because there is a feature in the limit of the search range.

That's not good, this test is supposed to check for a feature _outside_ the initial search range... Care to try and modify it so that gives a real rather than a false positive?

> I am also confused about one of the two Travis builds failing. It seems that the failure is due to issues installing Perl v5.14 so I believe this is not related to the changes I made.

We are aware if that problem, it is indeed unrelated to your changes.

nerdstrike:
This is my fault for not properly testing the outward iterator. @AsierGonzalez I propose the following:

If the fourth iteration is 240 bp, then adding a test at L183 should help, like so:

`
# Test multiplication error in search window of fetch_all_by_outward_search
@results = @{ $sfa->fetch_all_by_outward_search(-FEATURE => $a, -LIMIT => 5, -RANGE => 10, -MAX_RANGE => 140)};
cmp_ok($results[2]->[0]->display_id, 'eq', 'b', 'Check for closest feature in initial search range');
ok(! defined $results[3], 'Feature $c is not found because it is out of the MAX_RANGE';
`

The distance between $a and $c is 160, hence the original test still finds $c once you've corrected the multiplier. If it limits correctly, it should not find $c. Will that help get this fix in?

AsierGonzalez:
Hi all,
Sorry for not replying to the previous messages, I'm swamped with work at the moment and this is in my to-do list. Thank you for the useful feedback and links to the three of you.
@nerdstrike, the idea is to do something like that, if I we reduce the limit to less than 160 bp the current method will stil find $c but the correct implementation won't.
I'll make the changes as soon as I can and I'll let you know.

mkszuba:
Thank you! Unfortunately even after this change the test in question continues to pass on _master_.

AsierGonzalez:
I know, I mean to update you on this but I didn't have time.
Last week I tried to set up the test environment as per your suggestions but I didn't succeed. It was really frustating, because I tried to install it both on a MAC and a Debian machine and failed miserably in both. I believe that the problem on the Debian machine is because the test are not suited to use MySQL 8 but there is no information about it in the documentation.
Anywa, th summary is that I gave up and decided to use Travis instead. I was puzzled to see that this test still passes. I need to look into it again but do you have any ideas?

AsierGonzalez:
I have finally managed to make the tests run so the fix should be ready to be merged. However, I guess that it should now be merged with the release/99 branch and there are some conflincts. Should I just close this PR and start a new one with the current version? @mkszuba, @ens-bwalts, @nerdstrike

nerdstrike:
Well done @AsierGonzalez. I leave the integration issues to my colleagues. It should be relatively straightforward to base your changes off master and lose all the testing baggage your PR has acquired.

mkszuba:
Thanks a lot for your work on this issue, @AsierGonzalez ! I am afraid it is already too late to have this included into release 99, my recommendation is for you to start with the current state of _master_, cherry-pick your changes from the current _bugfix/fix_fetch_outward_search_, then force-push it to GitHub as the new _bugfix/fix_fetch_outward_search_ in order to keep this PR open.
Or since I have just done the cherry-picking myself in order to see if it caused any problems and pushed the result to GitHub (see _Ensembl:bugfix/fix_fetch_outward_search_), you could just confirm I didn't miss anything and subsequently skip the first two steps :)

AsierGonzalez:
Hi @mkszuba, I have checked _Ensembl:bugfix/fix_fetch_outward_search_ and all my changes are there. What should we do now?

mkszuba:
Just force-push it to _AsierGonzalez:bugfix/fix_fetch_outward_search_ and we shall take it from there.

PR #416, opened 2019-09-03
Backport MIMParser from ensembl-xref

mkszuba:
## Description

Port changes made to MIMParser in _ensembl-xref_, including unit tests, back to _ensembl:feature/xref_sprint_.

## Use case

See ENSCORESW-3217.

## Benefits

## Possible Drawbacks

## Testing

Yes.

They will pass once #413 and #415 have been merged.


mkszuba:
I have just compared the output of MIMParser on this branch vs _release/98_, all observed differences have been accounted for. These are:
 - the old version set xref _info_type_ to DEPENDENT under the assumption OMIM xrefs are always dependent on NCBI ones, the new one sets it to UNMAPPED and leaves it to Mim2GeneParser to change it to either DIRECT or DEPENDENT if an appropriate mapping exists;
 - the old version frequently erroneously concatenated words in _label_ and _description_ because it didn't add any spaces in between while concatenating subsequent relevant lines of the input file, the new version does this the right way (I hope).

PR #419, opened 2019-09-03
Backport HPAParser from ensembl-xref

mkszuba:
## Description

Port changes made to HPAParser in _ensembl-xref_, including unit tests, back to _ensembl:feature/xref_sprint_. Originally worked on by Mark McDowall.

## Use case

See ENSCORESW-3213.

## Benefits

## Possible Drawbacks

## Testing

Yes.

They pass.


mkszuba:
Output comparison with _release/98_ - all accounted for:
 - the old version sets _xref.description_ to _translation_direct_xref.linkage_xref_ empty strings, the new one leaves them at NULL

misc-scripts/xref_mapping/XrefParser/HPAParser.pm
```diff
@@ -45,71 +79,60 @@ sub run {
...
-  $file_io->getline()
-    ;    # skip header # Antibody,antibody_id,ensembl_peptide_id,link
+  my @expected_columns = qw( Antibody antibody_id ensembl_peptide_id link );
+  my $header           = $input_file->getline($file_io);
 
+  if ( scalar @{$header} != scalar @expected_columns ) {
+    confess "input file $file has an incorrect number of columns";
+  }
+
+  $input_file->column_names(@expected_columns);
```
ens-bwalts: This seems dangerous, shouldn't the column names be set from `@{$header}`? That way, if the format changes there is at least a chance for the parser to get the right data if the change is a simple rearrangement.

mkszuba: Guess it was Mark's Python background (in Python attempting to read a dict value corresponding to a nonexistent key results in an exception) showing :-)

Anyway, I have opted for a more strict header validation procedure as implemented in the other parsers I have worked on. Moreover, I have got rid of _getline_hr()_ because it is known to be slow.

PR #423, opened 2019-09-12
ENSCORESW-3222: Backport VGNCParser

tgrego:
## Description

Port changes made to VGNCParser in ensembl-xref, including unit tests, back to ensembl:feature/xref_sprint.

## Use case

See ENSCORESW-3222

## Benefits

Salvage more work from the xref sprint.

## Possible Drawbacks

Increased test-suite run time.

## Testing

yes

pass

yes


mkszuba:
Have you compared the output of this version of VGNCParser to one from either _master_ or the current release branch? If so, are all differences accounted for?

misc-scripts/xref_mapping/t/vgnc.t
```diff
@@ -0,0 +1,100 @@
...
+  <http://www.ensembl.org/Help/Contact>.
+
+=cut
+
+use strict;
+use warnings;
+use Test::More;
+use Test::Exception;
+use FindBin '$Bin';
+use lib "$Bin/";
```
mkszuba: Do we really need this? I would rather avoid messing with @INC unless there is a good reason for it.

misc-scripts/xref_mapping/XrefParser/VGNCParser.pm
```diff
@@ -15,6 +15,63 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+
+=head1 NAME
+
+XrefParser::VGNCParser
+
+=head1 DESCRIPTION
+
+A parser class to parse the VGNC source.
+VGNC is the official naming source for some vertebrates species
+(currently chimp, cow, dog and horse).
```
mkszuba: I would remove this bit and save ourselves the trouble of having to update it every time we add the parsing of VGNC data for another species - of which we will, according to an e-mail Bethan Yates sent to ensembl-core on 2019-06-10, get quite a few of soon. In fact, the above list is already out of date because we have had cat, macaque and mouse-lemur VGNC xrefs in Ensembl since release 98.

misc-scripts/xref_mapping/XrefParser/VGNCParser.pm
```diff
@@ -26,6 +83,13 @@ use Text::CSV;
 
 use parent qw( XrefParser::HGNCParser );
 
+
+=head2 run
+  Description: Runs the VGNCParser
+  Return type: N/A
```
mkszuba: Technically, we return an integer here... In practice it's always 0 because refactored parsers are all supposed to abort with an exception unless successful - but we do return something.

misc-scripts/xref_mapping/XrefParser/VGNCParser.pm
```diff
@@ -15,6 +15,63 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
```
ens-bwalts: This should be replaced with the note to look at the notice file. I missed this when merging #421 (ZFINDesc) - apologies

misc-scripts/xref_mapping/t/vgnc.t
```diff
@@ -0,0 +1,100 @@
+=head1 LICENSE
+
+See the NOTICE file distributed with this work for additional information
+regarding copyright ownership.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
```
ens-bwalts: Likewise, notice file.

tgrego:
For pan_troglodytes new parser loads a total of 16563 VGNC xrefs and 376 synonyms, the exact same number as the master branch.

PR #430, opened 2019-10-16
Add next-value buffering to Bio::EnsEMBL::Utils::Iterator

muffato:
## Use case

I'm instantiating a Bio::EnsEMBL::Utils::Iterator with a simple sub that does this:
```perl
my $fetch_sub = sub {
    return $sth->fetchrow_arrayref;
}
```

Then, when I iterate I get this error:
```DBD::mysql::st fetchrow_arrayref failed: fetch() without execute()```

This is because `$iterator->has_next()` keeps on calling the coderef even though it has already returned _undef_ and DBD::mysql::st thinks enough is enough.

## Description

Here, I use _exists_ instead of _defined_ and I leave _undef_ in `$self->{next}` once it's been reached.

## Benefits

Can use DBI fetch methods in the iterator

## Possible Drawbacks

N/A

## Testing

Yes

Yes

Yes

### DIFF
--- a/modules/Bio/EnsEMBL/Utils/Iterator.pm
+++ b/modules/Bio/EnsEMBL/Utils/Iterator.pm
@@ -141,6 +141,26 @@ sub new {
 }
 
 
+=head2 _fetch_next_value_if_needed
+
+  Example    : $iterator->_fetch_next_value_if_needed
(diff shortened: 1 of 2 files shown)


muffato:
Thank you !

PR #418, opened 2019-09-03
Backport EntrezGeneParser from ensembl-xref

mkszuba:
## Description

Port changes made to EntrezGeneParser in _ensembl-xref_, including unit tests, back to _ensembl:feature/xref_sprint_. Originally worked on by Alessandro.

## Use case

See ENSCORESW-3211.

## Benefits

Salvage more work from the 2018 xref sprint.

## Possible Drawbacks

Increased test-suite run time (note: Travis does not automatically execute tests from _misc-scripts/xref_mapping_ yet).

## Testing

_Have you added/modified unit tests to test the changes?_

Yes.

_If so, do the tests pass/fail?_

They will pass once #415 has been merged.

_Have you run the entire test suite and no regression was detected?_

Yes (including tests from _misc-scripts/xref_mapping_), no regression detected.

PR #431, opened 2019-10-16
Backport MGI_Desc_Parser from ensembl-xref

mkszuba:
## Description

Port changes made to MGI_Desc_Parser in _ensembl-xref_, including unit tests, back to _ensembl:feature/xref_sprint_.

## Use case

See ENSCORESW-3216. Originally worked on by Prem.

## Benefits

## Possible Drawbacks

## Testing

Yes.

They pass.

### DIFF
--- a/misc-scripts/xref_mapping/XrefParser/MGI_Desc_Parser.pm
+++ b/misc-scripts/xref_mapping/XrefParser/MGI_Desc_Parser.pm
@@ -87,9 +87,9 @@ sub run {
 
   my $input_file = Text::CSV->new({
     sep_char           => "\t",
-    empty_is_undef     => 1,
+    quote_char         => undef,
+    escape_char        => undef,
     strict             => 1,
-    allow_loose_quotes => 1,
   }) or confess "Cannot use file $file: " . Text::CSV->error_diag();
 
   my $xref_count = 0;
--- a/misc-scripts/xref_mapping/XrefParser/MGI_Desc_Parser.pm
+++ b/misc-scripts/xref_mapping/XrefParser/MGI_Desc_Parser.pm
@@ -1,9 +1,7 @@
-
 =head1 LICENSE
 
-Copyright [1999-2015] Wellcome Trust Sanger Institute and the EMBL-European Bioinformatics Institute
-Copyright [2016-2019] EMBL-European Bioinformatics Institute
-
+See the NOTICE file distributed with this work for additional information
+regarding copyright ownership.
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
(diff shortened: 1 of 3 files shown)


mkszuba:
Output comparison with _release/98_: only one difference but an interesting one, turns out that when the old parser processes MGI:4457207 it produces the synonym name _&quot;VC&quot;_ whereas it is in fact simply _"VC"_. The refactored parser returns the synonym name as it should be i.e. does not insert HTML entities where they are not needed.

mkszuba:
Output comparison with _release/99_ after the change to the handling of quotation marks introduced by 5bb6bf9: no differences!

misc-scripts/xref_mapping/XrefParser/MGI_Desc_Parser.pm
```diff
@@ -18,135 +16,126 @@ limitations under the License.
...
-  my @expected_columns =
-    qw(accession chromosome position start end strand label status marker marker_type feature_type synonym_field);
-
   # read header
-  my $header = $input_file->getline($mgi_io);
-  if ( scalar @{$header} != scalar @expected_columns ) {
-    croak "input file $file has an incorrect number of columns";
+  my @header_fields = $input_file->header($mgi_io);
+  if ( scalar @header_fields != $EXPECTED_NUMBER_OF_COLUMNS ) {
+    confess "Header of input file $file has an incorrect number of columns";
```
ens-bwalts: As with #418, I think a check of the header fields can prevent future headaches if they change the format.

mkszuba: Done.

misc-scripts/xref_mapping/XrefParser/MGI_Desc_Parser.pm
```diff
@@ -18,135 +16,125 @@ limitations under the License.
...
-
-  if((!defined $source_id) or (!defined $species_id) or (!defined $files) ){
-    croak "Need to pass source_id, species_id and files as pairs";
+  my $source_id  = $ref_arg->{source_id};
+  my $species_id = $ref_arg->{species_id};
+  my $files      = $ref_arg->{files};
+  my $verbose    = $ref_arg->{verbose} // 0;
+  my $dbi        = $ref_arg->{dbi} // $self->dbi;
+
+  if ( ( !defined $source_id ) or
```
nerdstrike: Personally I would take the NOT( AND AND AND) version of this sort of logic, as it easier to process, but whatever

misc-scripts/xref_mapping/XrefParser/MGI_Desc_Parser.pm
```diff
@@ -18,135 +16,125 @@ limitations under the License.
...
+    });
     if ( $verbose && !$marker ) {
       print "$accession has no description\n";
     }
-    $xref_count++;
-    my @synonyms;
-    if ( defined( $acc_to_xref{$accession} ) ) {
-      @synonyms = split( /\|/, $data->{'synonym_field'} )
-        if ( $data->{'synonym_field'} );
+    $xref_count += 1;
```
nerdstrike: Boy, you don't like ++

PR #441, opened 2019-11-11
Remove deprecated variation-related methods

tgrego:
## Description

A few deprecated methods have been deprecated and are due to be removed at release 95.
Those should be removed effective of release 100.
The extra 5 release should be plenty of time for any users to have updated their code.

## Use case

See ENSCORESW-3294

## Benefits

Less deprecation notices and runtime of our test harness.

## Possible Drawbacks

Users still relying on those methods have been getting deprecation notices for a long time, if they did not update their code they will have to do so now.

## Testing

yes

pass

yes


ens-bwalts:
Should have been more clear that I only checked the core repo - ambiguity between Ensembl (the project) and ensembl (the repo). Apologies.

at7:
I should have removed the deprecation message when I first found out about the compara/web dependency. I don't know why I didn't. Probably because I was hoping to come up with a solution which until today I haven't.

tgrego:
good thing we asked variation!
I'll close this pull request and we'll think of the next move

PR #443, opened 2019-11-13
Gitlab CI: enable triggering of dependent builds on Travis

mkszuba:
## Description

Extend .gitlab-ci.yml so that _ensembl_ pipelines on EBI GitLab can trigger dependent builds in projects still using Travis.

## Use case

The switch from to GitLab CI is to take place on a per-team basis so there will be dependent projects on Travis.

## Benefits

Ensure CI compatibility of Core projects on EBI GitLab with dependent projects still on Travis.

## Possible Drawbacks

In order for all dependent builds to be triggered successfully the user whose API token is used to authenticate with Travis REST API must have write access to all dependent repositories, which will have to be set up manually (in case of all-Travis runs there is implicit organisation-wide trust).

## Testing

No.

N/A

N/A, changes only affect the running of tests on EBI GitLab.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -41,7 +41,7 @@ before_script:
 
 test:perl5.14-sqlite:
   stage: test
-  image: perl:5.14-threaded
+  image: perl:5.14
   variables:
     COVERALLS: "false"
     DB: "sqlite"
@@ -50,7 +50,7 @@ test:perl5.14-sqlite:
 
 test:perl5.30-mysql:
   stage: test
-  image: perl:5.30-threaded
+  image: perl:5.30
   variables:
     # Note: relies on the secret variable COVERALLS_REPO_TOKEN for report uploads to work
     COVERALLS: "true"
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -62,24 +62,47 @@ test:perl5.30-mysql:
...
-# Example trigger job using the above, disabled for obvious reasons
-.post:trigger_foo:
+# Dependent builds on Travis
+# Relies on the secret variable TRAVIS_AUTH_TOKEN to actually work,
+# moreover the account associated with the token must have write
+# access to *all* dependent repositories; to be exact what it needs
+# is the Travis 'create_request' permission but Travis permissions are
+# generated from GitHub ones and it seems that in order to have
+# 'create_request' on the latter one requires 'write' on the former.
+post:trigger_travis:
```
ens-bwalts: does this need to be `.post`?

mkszuba: Good point, with the "when: always" condition in place all delegating it to the second stage did was delay the downstream build.

PR #446, opened 2019-11-20
[ENSCORESW-3183][ENSCORESW-3176] Mapper performance improvements

ens-bwalts:
## Requirements

## Description

These changes improve the speed of the recent interval tree based Mapper implementation. There are two main sets of fixes in this PR:

- Improvements to the immutable interval tree code to speed up building a new tree (by not sorting already-sorted lists, and using less stringent typechecking)
- Caching interval trees used by the mapper for reuse, rather than re-creating them each time `map_coordinates()` or `map_insert()` is called

I also fixed up `_dump()` so that it works again and is more useful, as I needed to use that method to aid in implementing these changes.

## Use case

Other teams have raised several JIRA tickets complaining about the slowness of the release/97 interval tree based Mapper. The Mapper is used directly or indirectly in many, many places throughout Ensembl, so its performance issues have wide impact for production as well as other Ensembl users. These are gathered in JIRA epic ENSCORESW-3183, and this in particular addresses the difficulty reported by production in ENSCORESW-3176

## Benefits

A 2-3 order of magnitude speedup in performance. For example, the Mapper calls involved in dumping the sequence of human chromosome 20 speeds up from  4954 seconds (82 minutes, 465 ms per call) to 19.7 seconds (1.8 ms per call). Most of this improvement is due to caching of trees in Mapper. The improvements in the IntervalTree itself cut the runtime by about 1/3 to 1/2

## Possible Drawbacks

We need to make sure that Mapper adaptors (e.g. AssemblyMapperAdaptor) continue to cache the Mappers themselves, so that they can be re-used. Otherwise, we end up having to go through the expense of rebuilding all these data structures all over again.

Because of these changes, Mapper is now fully subject to the two hard things in computer science: naming things, cache invalidation, and off by one errors. Any further changes to the Mapper need to be carefully considered to be sure stored trees are correctly instantiated and removed.

## Testing

Yes, tests for tree caching and invalidation in mapper.t

pass

yes

In addition, I have run sequence dumper scripts with these new changes and the output is identical to the pre-change output.

### DIFF
--- a/modules/Bio/EnsEMBL/Mapper.pm
+++ b/modules/Bio/EnsEMBL/Mapper.pm
@@ -1023,12 +1023,36 @@ sub _dump{
   if( !defined $fh ) {
     $fh = \*STDERR;
   }
-  foreach my $id ( keys %{$self->{'_pair_hash_from'}} ) {
-    print $fh "From Hash $id\n";
-    foreach my $pair ( @{$self->{'_pair_hash_from'}->{uc($id)}} ) {
+
+  my $from = $self->{'from'};
+  my $to = $self->{'to'};
+
+  print $fh "--------------------begin mapper dump--------------------\n";
+  print $fh "dumping from-hash _pair_$from\n";
+  foreach my $id ( keys %{$self->{"_pair_$from"}} ) {
+    print $fh "{_pair_$from}->{" . uc($id) . "}:\n";
+    foreach my $pair ( @{$self->{"_pair_$from"}->{uc($id)}} ) {
       print $fh "    ",$pair->from->start," ",$pair->from->end,":",$pair->to->start," ",$pair->to->end," ",$pair->to->id,"\n";
     }
+    if (defined($self->{"_tree_$from"}->{uc($id)})) {
+      print $fh "{_tree_$from}->{" . uc($id) . "} instantiated\n";
+    } else {
+      print $fh "{_tree_$from}->{" . uc($id) . "} empty\n";
+    }
   }
+  print $fh "dumping to-hash _pair_$to\n";
+  foreach my $id ( keys %{$self->{"_pair_$to"}} ) {
+    print $fh "{_pair_$to}->{" . uc($id) . "}:\n";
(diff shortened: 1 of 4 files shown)


modules/Bio/EnsEMBL/Mapper.pm
```diff
@@ -1163,5 +1179,44 @@ sub _is_sorted {
...
+  my $from_intervals;
+
+  foreach my $i (@{$pair_list}) {
+    my $start = $i->{$pair_side}{start};
+    my $end = $i->{$pair_side}{end};
+
+    if ($end < $start) {
+      my $tmp = $start;
+      $start = $end;
+      $end = $tmp;
```
tgrego: this might be slightly more efficient as
```perl
($start, $end) = ($end, $start) 
```

modules/Bio/EnsEMBL/Mapper.pm
```diff
@@ -1023,12 +1012,36 @@ sub _dump{
...
+    foreach my $pair ( @{$self->{"_pair_$to"}->{uc($id)}} ) {
+      print $fh "    ",$pair->to->start," ",$pair->to->end,":",$pair->from->start," ",$pair->from->end," ",$pair->from->id,"\n";
+    }
+    if (defined($self->{"_tree_$to"}->{uc($id)})) {
+      print $fh "{_tree_$to}->{" . uc($id) . "} instantiated\n";
+    } else {
+      print $fh "{_tree_$to}->{" . uc($id) . "} empty\n";
+    }
   }
+  print $fh "---------------------end mapper dump---------------------\n";
```
tgrego: do we want this mapper dump in production?

ameya1981:
Some more feedback from Variation testing:

modules/Bio/EnsEMBL/Mapper.pm
ens-bwalts: Cleaned this up to look a bit better. This has been in since the early days of Mapper, and is useful, so I'd argue for keeping some sort of _dump in.

PR #447, opened 2019-11-20
Task variation/remove deprecation status

at7:
We cannot remove those methods from the core API. The webcode API depends on them. They are used for example here: https://github.com/Ensembl/ensembl-webcode/blob/release/98/modules/EnsEMBL/Web/Query/GlyphSet/Variation.pm#L205-L209 And here https://github.com/Ensembl/ensembl-webcode/commits/release/98/modules/EnsEMBL/Web/Component/TextSequence.pm
The slice used in the webcode is a compara AlignSlice. If I rewrite the method to be used in the VariationFeatureAdaptor and pass the AlignSlice from the webcode the method doesn't work.

Dan S commented a while ago saying:
_I'm having to revert that because it doesn't seem to interact well with AlignSlices (rather than real Slices). This breaks variant display on the Phylogenetic Context page (focus variant and other variants do not display). It looks like this is because the new method (in the VariationFeatureAdaptor) eventually calls get_and_filter_Slice_projections in Bio::EnsEMBL::DBSQL::BaseFeatureAdaptor inside the core code, which objects that the slice has no adaptor available via ->adaptor._

### DIFF
--- a/DEPRECATED.md
+++ b/DEPRECATED.md
@@ -93,10 +93,6 @@ A year after deprecation (4 Ensembl releases), the method is removed from the AP
  - Bio::EnsEMBL::**Slice**::*calculate_theta()*
  - Bio::EnsEMBL::**Slice**::*get_all_IndividualSlice()*
  - Bio::EnsEMBL::**Slice**::*get_by_Individual()*
- - Bio::EnsEMBL::**Slice**::*get_all_somatic_VariationFeatures()*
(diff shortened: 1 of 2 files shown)


at7:
The methods were deprecated in the first place because it looked straight forward to move the functionality into the variation API. The code has then been updated in the web code where it eventually caused problems and had to be reverted back to using the code from the core API. It will not be used by the new website but will still be needed for the current website for quite a while

ens-bwalts:
Thanks for the update, @at7

PR #455, opened 2019-12-16
Patch to support longer assembly names in the mapping session table.  [release/100]

tgrego:
## Description

see https://github.com/Ensembl/ensembl/pull/453

## Use case

## Benefits

## Possible Drawbacks

## Testing


tgrego:
#453 needs reverting and fixing...
closing

PR #460, opened 2019-12-17
Core db schema patch to support longer assembly names in the mapping_session table.

tgrego:
## Description

This is a copy of https://github.com/Ensembl/ensembl/pull/458 against master.

## Use case

## Benefits

## Possible Drawbacks

## Testing


ens-carlos:
It was arbitrary to match the new_db_name. In the assembly registry I think we use 255 VARCHAR arbitrarily as well.

PR #467, opened 2020-01-10
JSON-formatted web_data

james-monkeyshines:
## Description
Storing web_data as a flattened Perl hash is cumbersome, and, as it turns out, unnecessary - it's required by the webcode, but it is converted into JSON for processing. So we can make everone's lives easier by storing it as JSON in the first place. (cf. https://www.ebi.ac.uk/panda/jira/browse/ENSINT-396)

## Use case
The main use case is for the webcode to retrieve the web_data in order to determine how to display a track. Setting the web_data field via the API is uncommon, but might occasionally be necessary.

## Benefits
The web_data field no longer needs special treatment, it's just a JSON string that is set/got like any other text field.

## Possible Drawbacks
Changing this requires tight integration with webcode changes. Those have been made (https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-5249), and the plan is to get a sandbox running ASAP for testing the web display.

## Testing
Yes, analysis.t updated.

Pass

Yes, all fine.


james-monkeyshines:
Assuming this is OK, please let me know if you want a separate PR against the master branch.

ameya1981:
Pls raise a PR to master too.

PR #474, opened 2020-03-13
Update of gene and transcript includes version update

mira13:
Description

On gene and transcript update, update version as well;

Use case

When other teams runs gene/transcript updates, they have a problem to separately update version with direct sql request, so we just include version update in common update.

Benefits

Less code in pipeline, remove from pipelines authors responsibility to update a version separately from common update

Possible Drawbacks

Version will be updated in all cases

Testing

existing tests covers proposed code, data health checks will cover the rest of the testcases


ameya1981:
Though the changes look fine for the version update, this is not yet approved for e101 as an intention. Will approve this once the intention goes through.

ameya1981:
Same PR to be raised against release/101

PR #482, opened 2020-03-31
Bugfix/map insert fix

mira13:
## Description

Fix map_insert, include original coordinates on demand

## Use case
Full description is here: 
https://www.ebi.ac.uk/panda/jira/browse/ENSCORESW-3413

## Testing

Should not have impact on tests

### DIFF
--- a/modules/Bio/EnsEMBL/Mapper.pm
+++ b/modules/Bio/EnsEMBL/Mapper.pm
@@ -490,48 +490,87 @@ sub map_insert {
   my @coords = $self->map_coordinates($id, $start, $end, $strand, $type, $include_original_region);
 
   if(@coords == 1) {
-    my $c = $include_original_region ? @coords[0]->{'mapped'} : $coords[0];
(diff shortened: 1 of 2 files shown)


modules/Bio/EnsEMBL/Mapper.pm
```diff
@@ -499,9 +499,9 @@ sub map_insert {
     # adjust coordinates, remove gaps
     my ($c1, $c2);
     if($strand == -1) {
-      ($c2,$c1) = @coords;
+      ($c2,$c1) = $include_original_region ? @coords->{'mapped'} : @coords;
```
ens-bwalts, line 502: I don't think that syntax works in all versions of Perl; I can't get it to work in 5.14 or 5.26. Would be safer to use

`map {$_->{'mapped'}} @coords` to get the array slice here.

ens-bwalts: Reworked.

PR #488, opened 2020-06-05
[ENSWEB-5706] Allow multiple "ancestral" databases in the Registry

muffato:
## Use case

See ENSWEB-5706 and related tickets for more information. The REST server sees all MySQL servers (Vertebrates and non-Vertebrates), which contain in total two ancestral databases, but the current Registry model only allows one. The Compara API then fails because it is not able to access the other one

## Description

Ancestral databases are now registered with their division name, preventing a clash between the Plants and Vertebrates databases.

Note that for backwards compatibility and for consistency with the rest of the code, the name "Vertebrates" is currently omitted.

:warning: _Do not merge until *all* these related pull-requests have been approved !_: Ensembl/ensembl-compara#197, Ensembl/ensembl-production#430

Equivalent pull-requests: #491 for `release/101` and #492 for `master`

## Benefits

Now able to load all ancestral databases in the same Registry instance

## Possible Drawbacks

Applications that load the NV server in the Registry won't be able to access the Plants ancestral database with "Ancestral sequences", but since this database has been introduced in e100, I think there is still time to change its Registry name and state that this is the nomenclature going forward.

## Testing

No. The databases created by `registry.t` are prefixed by the user name and `test_db`. They wouldn't pass the regex used by `load_registry_from_db`

Failures are in:
* `compara.t` (because `.travis.yml` clones the master branch of Compara instead of the `release/100` branch
* `housekeeping_apache2.t`, because it is now 2020 and the files haven't had their copyright statement updated. I have only updated the one of `Registry.pm`

### DIFF
--- a/modules/Bio/EnsEMBL/Registry.pm
+++ b/modules/Bio/EnsEMBL/Registry.pm
@@ -1,7 +1,7 @@
 =head1 LICENSE
 
 Copyright [1999-2015] Wellcome Trust Sanger Institute and the EMBL-European Bioinformatics Institute
-Copyright [2016-2019] EMBL-European Bioinformatics Institute
+Copyright [2016-2020] EMBL-European Bioinformatics Institute
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


muffato:
Question: can you confirm that you don't have a mechanism to merge release branches so I will have to make another PR for `release/101` and one for `master` ?

ens-bwalts:
@muffato yes, we prefer separate PRs for separate branches

PR #490, opened 2020-06-10
ENSCORESW-3476: import direct mappings to UniProt isoforms

magaliruffier:
## Requirements

## Description

Add direct mappings to Uniprot isoforms, in addition to the existing mappings to the canonical accession

## Use case

A single Uniprot accession can contain multiple sequences, mapping to different Ensembl transcripts
Adding the mappings to the specific isoform allows us to identify which of these sequences is an exact match to the equivalent Ensembl peptide

## Benefits

More specific information available for users

## Possible Drawbacks

## Testing

The change was tested on human and it has successfully added additional xref mappings to this new sources without changing existing xrefs

### DIFF
--- a/misc-scripts/xref_mapping/XrefParser/UniProtParser.pm
+++ b/misc-scripts/xref_mapping/XrefParser/UniProtParser.pm
@@ -57,7 +57,7 @@ sub run {
 
   my $file = @{$files}[0];
 
-  my ( $sp_source_id, $sptr_source_id, $sp_release, $sptr_release, $sptr_non_display_source_id, $sp_direct_source_id, $sptr_direct_source_id );
+  my ( $sp_source_id, $sptr_source_id, $sp_release, $sptr_release, $sptr_non_display_source_id, $sp_direct_source_id, $sptr_direct_source_id, $isoform_source_id );
 
   $sp_source_id =
     $self->get_source_id_for_source_name('Uniprot/SWISSPROT','sequence_mapped', $dbi);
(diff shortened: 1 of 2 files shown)


misc-scripts/xref_mapping/XrefParser/UniProtParser.pm
```diff
@@ -459,6 +464,21 @@ sub create_xrefs {
             $num_direct_sptr++;
           }
           push @{$xref->{DIRECT_XREFS}}, \%direct;
+
+          my $uniprot_acc = $accessions[0];
+          if ($extra[1] =~ /($accessions[0]-[0-9]+)/) {
+            $isoform = $1;
```
ens-bwalts, line 470: It's quite likely I'm missing something here, but I'm wondering if `$isoform` will be `[P31946-1]` with the literal curly braces included at this point?

magaliruffier: $extra[1] looks like `ENSG00000159840. [Q15942-1]`
we are extracting the part that starts with `$accessions[0]` and ends with some digits, curly braces are not included in the match
once parsed, the xrefs look like this in the database:
```
115171	P31946-1	0	P31946-1	NULL	141	9606	DIRECT		NULL
115189	P62258-1	0	P62258-1	NULL	141	9606	DIRECT		NULL
115193	P62258-2	0	P62258-2	NULL	141	9606	DIRECT		NULL
```

ens-bwalts: Of course, it's just matching the accession. Brain stall on my part.

PR #432, opened 2019-10-18
Backport XenopusJamboreeParser from ensembl-xref

mkszuba:
## Description

Port changes made to XenopusJamboreeParser in _ensembl-xref_, including unit tests, back to _ensembl:feature/xref_sprint_.

## Use case

See ENSCORESW-3223. Originally worked on by Prem.

## Benefits

Salvage more work from the 2018 xref sprint.

## Possible Drawbacks

Increased test-suite run time (note: Travis does not automatically execute tests from _misc-scripts/xref_mapping_ yet).

## Testing

_Have you added/modified unit tests to test the changes?_

Yes.

_If so, do the tests pass/fail?_

They pass.

_Have you run the entire test suite and no regression was detected?_

Yes (including tests from _misc-scripts/xref_mapping_), no regression detected.

PR #491, opened 2020-06-10
[ENSWEB-5706] Allow multiple "ancestral" databases in the Registry

muffato:
Same as #488 but on `release/101`


ens-bwalts:
Production changes in ensembl-production #432 and #433

PR #495, opened 2020-06-29
ENSCORESW-3503: keep analysis id when storing coordinate xrefs

magaliruffier:
## Requirements

## Description

Entries in the object_xref table for coordinate xrefs are being stored with an analysis_id 0
The corresponding analysis is created but not used
The proposed change passes the analysis_id for the correct analysis as an argument, to be stored in the object_xref table

## Use case

analysis_id 0 is not really good
No analysis_id would already be better, but as the analysis object already exist, might as well use it here

## Benefits

No entry in object_xref with analysis_id 0

## Possible Drawbacks

None that I can see

## Testing

NA

NA

This part of the code has no test cases associated to it.
The whole xref pipeline was run before and after the change to verify the change works as expected
In particular, the datachecks run at the end of the pipeline were reporting those entries with analysis_id 0 and do not with the change


magaliruffier:
no

PR #499, opened 2020-07-22
Feature/stable id update

mira13:
## Description

Make stable_id updatable with update function of gene and transcript

## Use case

This is an old existing bug, claimed to be resolved by other Ensembl teams

## Benefits

no bug any more

## Possible Drawbacks

no drawbacks

## Testing

Unit tests are updated 

All tests are passed with travis as well


magaliruffier:
Looks sensible to me
the update behaviour is only changed if a new stable ID is assigned, existing use of the update method should not be affected

PR #500, opened 2020-07-23
Bugfix/null translation

mira13:
#478  Description

Bugfix, that makes canonical_translation_id NULL on translation delete

## Use case
When translation is removed, links to the translation still thers in transcript table

## Benefits

Data consistency for translation table

## Possible Drawbacks

Not found

## Testing

Unit tests are added for this case

All tests pass, with travis as well


magaliruffier:
Looks good to me.
The update/add/remove methods are notoriously incomplete, it is good to add fixes as and when we have use cases for them

PR #501, opened 2020-08-03
Feature/xref standardisation

magaliruffier:
## Requirements

## Description

The list of sources used for gene naming and gene description has been consolidated in a single place and used for most vertebrate species. This removes the need to maintain multiple lists for different species with the risk of them being out of sync.
Additionally, when multiple choices are available, a re-ordering has been applied to increase the chances of making the same choice when the pipeline is re-run

## Use case

We have genes where the name and the description are different, because different sources were used for each. This is confusing users. We also have cases where the same gene has a different name from one release to another, because there are two choices and one is arbitrarily chosen each time

## Benefits

More consistency in the xref assignment, less queries from users, easier explanation of the name assignment strategy

## Possible Drawbacks

None that I can see

## Testing

no test cases, but the xref pipeline has been run on a large number of species to check the results

the pipeline runs fine and produces consistent results on the species it was tested on

NA


magaliruffier:
the gene_display_xref_sources method returns a reference array and a hash
the gene_description_sources method used to return a reference array and now returns a reference array and a hash
the hash is the second argument and is therefore ignored as not expected
this also means that if we want to add an ignore_list (badly named list as it is in fact a hash) for the gene_display_xref_sources, the argument is already provided

ens-bwalts:
It looks to me like gene_description_sources used to return an actual array, not a reference array, which is what `@precedence` is expecting - there's no dereferencing.

magaliruffier:
That's a fair point, let me have a look

magaliruffier:
That was a good spot, I have changed the return type and was able to test it on our favourite xenopus example, ENSXETG00000034365
it now gets both display_xref and description from the same NCBIGene xref

ameya1981 (approved):
Approved after the change in return type for sub gene_description_sources

ens-bwalts (approved):

PR #503, opened 2020-08-13
ENSCORESW-3349 changes made loops in parser for better performance.

ameya1981:
## Requirements

## Description

Optimize loops in RefSeqCoordinateParser which runs for human xrefs. No change in output is expected.

## Use case

Potential memory issues caused by holding objects in memory

## Benefits

Lesser memory required for parsing

## Possible Drawbacks

NA

## Testing

There is no unit test for this parser.
Running the parser as a singlejob did not result into any memory failures(over multiple times).
The xref count from these changes is same as the xref count from e100 parser

NA

### DIFF
--- a/misc-scripts/xref_mapping/XrefParser/RefSeqCoordinateParser.pm
+++ b/misc-scripts/xref_mapping/XrefParser/RefSeqCoordinateParser.pm
@@ -26,6 +26,7 @@ use DBI;
 
 use base qw( XrefParser::BaseParser );
 use Bio::EnsEMBL::Registry;
+use Digest::MD5 qw(md5);
 
 sub run_script {
 
@@ -131,6 +132,7 @@ sub run_script {
 
   if (defined $project && $project eq 'ensembl') {
 # Can use user-defined database
+#
       if (defined $host) {
           $core_dba = Bio::EnsEMBL::DBSQL::DBAdaptor->new(
               '-host'     => $host,
@@ -154,6 +156,7 @@ sub run_script {
       }
       if (defined $ofhost) {
 # Can use user-defined database
+
           $otherf_dba = Bio::EnsEMBL::DBSQL::DBAdaptor->new(
               '-host'     => $ofhost,
               '-user'     => $ofuser,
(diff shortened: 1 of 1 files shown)


misc-scripts/xref_mapping/XrefParser/RefSeqCoordinateParser.pm
```diff
@@ -397,14 +415,14 @@ sub run_script {
...
-            foreach my $dependent_xref_id (@{$entrez_ids{$entrez_id}}) {
+            while (my $dependent_xref_id = shift @{$entrez_ids{$entrez_id}}) {
               $add_dependent_xref_sth->execute($xref_id, $dependent_xref_id);
             }
           }
 
 # Also store refseq protein as direct xref for ensembl translation, if translation exists
           if (defined $tl && defined $tl_of) {
-            if ($tl_of->seq eq $tl->seq) {
+            if (md5($tl_of->seq) eq md5($tl->seq)) {
```
magaliruffier, line 425: what advantage does using the md5 bring?
we still call the seq method, so I am assuming the same query is still being run underneath

misc-scripts/xref_mapping/XrefParser/RefSeqCoordinateParser.pm
```diff
@@ -369,7 +387,7 @@ sub run_script {
 # If a best match was defined for the refseq transcript, store it as direct xref for ensembl transcript
         if ($best_id) {
           my ($acc, $version) = split(/\./, $id);
-	  $version =~ s/\D//g if $version;
+	        $version =~ s/\D//g if $version;
```
magaliruffier, line 390: cosmetic changes should be avoided, and if absolutely necessary, submitted as a separate commit to avoid masking the actual code changes

magaliruffier (changes requested):
It is not very clear to me what these changes actually bring.
If the memory problem could not be reproduced with the original code, how do we know the changes improve things?

magaliruffier:
can't confirm if changes are needed

PR #507, opened 2020-08-17
Fix travis build

mira13:
## Description

Fixing tests and travis build script. Befor in travia as a result was used variable that is always 0. So we had successful tests whatever is failed. After travis was fixed, it became obvious that some tests are not pasing, they are fixed as well - transcript, translation and gene tests.
Few variables in sql requests were quoted, it doesn't make any difference in current sql, but makes code work in new sql server

detail of the fix: 
before, we took as a result last command output - $? , Last tests command is `perl test run` inside `if` statement, so when after `if` we took $? - it was result of `if`, but not the test run inside. Eventually of `if` obviously is always 0.

## Use case

Travis test result  was sucsessful, even if tests fails

## Benefits

We have correct test system. 

## Possible Drawbacks

Still few more fails of tests are not covered in this branch, so it will be futher work

## Testing

Yes, all tests now correct and pass
I have run all them localy and on server

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,7 @@ before_install:
 install:
   - cpanm -v --installdeps --notest . --with-all-features
   - cpanm -n Devel::Cover::Report::Coveralls
-  - cpanm -n DBD::SQLite
+  - cpanm -n DBD::SQLite JSON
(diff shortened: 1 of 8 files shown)


ens-bwalts (approved):
Updating the tests to avoid using hardcoded analysis and canonical_transcript IDs looks like a nice improvement. Otherwise this looks OK

PR #509, opened 2020-08-26
Feature/xref optimisation

magaliruffier:
## Requirements

## Description

The proposed changes simplify the stage where data is copied from the xref database into the core database.

## Use case

Some of the queries run are quite heavy on the MySQL server. When multiple species are being run in parallel, these queries can get stuck on a heavily loaded server.
As they deal with data that is not used in the core database, removing these steps from the pipeline makes the code simpler and less onerous to run.

## Benefits

Less contention on the server for multiple running jobs.

## Possible Drawbacks

Historical unmapped entries are not conserved.

## Testing

No test cases available but the pipeline was run on about 20 species with and without the proposed changes.
The results between runs are comparable.
The mapping stage runs faster with the changes

### DIFF
--- a/misc-scripts/xref_mapping/XrefMapper/XrefLoader.pm
+++ b/misc-scripts/xref_mapping/XrefMapper/XrefLoader.pm
@@ -562,69 +562,17 @@ GCNTSQL
 
   #######################################
   # Remember to do unmapped entries
-  # 1) make sure the reason exist/create them and get the ids for these.
-  # 2) Process where dumped is null and type = DIRECT, DEPENDENT, SEQUENCE_MATCH, MISC seperately
   ########################################
-  my %summary_failed;
-  my %desc_failed;
-  my %reason_id;
 
-  # Get the cutoff values
-  $sth = $xref_dbi->prepare("select distinct s.name, m.percent_query_cutoff, m.percent_target_cutoff from source s, source_mapping_method sm, mapping m where sm.source_id = s.source_id and sm.method = m.method");
-  $sth->execute();
-  my ($source_name, $q_cut, $t_cut);
-  $sth->bind_columns(\$source_name, \$q_cut, \$t_cut);
-
-  while ($sth->fetch) {
-    $summary_failed{$source_name} = "Failed to match at thresholds";
-    $desc_failed{$source_name}    = "Unable to match at the thresholds of $q_cut\% for the query or $t_cut\% for the target";
-  }
-  $sth->finish;
-
-  $summary_failed{"NO_STABLE_ID"} = "Failed to find Stable ID";
-  $desc_failed{"NO_STABLE_ID"}    = "Stable ID that this xref was linked to no longer exists";
-
-  $summary_failed{"FAILED_MAP"} = "Failed to match";
-  $desc_failed{"FAILED_MAP"}    = "Unable to match to any ensembl entity at all";
-
-  $summary_failed{"NO_MAPPING"} = "No mapping done";
-  $desc_failed{"NO_MAPPING"}    = "No mapping done for this type of xref";
-
-  $summary_failed{"MASTER_FAILED"} = "Master failed";
-  $desc_failed{"MASTER_FAILED"}    = "The dependent xref was not matched due to the master xref not being mapped";
-
-  $summary_failed{"NO_MASTER"} = "No Master";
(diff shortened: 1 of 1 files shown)


misc-scripts/xref_mapping/XrefMapper/XrefLoader.pm
```diff
@@ -65,21 +65,18 @@ sub update{
...
-DELETE ontology_xref, object_xref, xref, dependent_xref
-FROM ontology_xref, object_xref, xref 
-LEFT JOIN dependent_xref on xref_id = dependent_xref_id
-WHERE ontology_xref.object_xref_id = object_xref.object_xref_id AND object_xref.xref_id = xref.xref_id
-SQL
+  $sql = "DELETE ox FROM object_xref ox, xref x, external_db e WHERE x.xref_id = ox.xref_id AND x.external_db_id =e.external_db_id and e.db_name = 'GO'";
   $sth = $core_dbi->prepare($sql);
-  $affected_rows = $sth->execute();
-  print "\tDeleted $affected_rows PROJECTED ontology_xref row(s)\n" if $verbose;
+  $sth->execute();
```
ens-bwalts, line 79: Might be nice to confirm this deletion in a print statement, either by printing here or summing up the rows with the delete at lines 81-83

ens-bwalts (approved):
Otherwise this looks OK

misc-scripts/xref_mapping/XrefMapper/XrefLoader.pm
```diff
@@ -65,21 +65,18 @@ sub update{
...
   print "\tDeleted $affected_rows PROJECTED external_synonym row(s)\n" if $verbose;
-
-  # Delete all ontologies, as they are done by a separate pipeline
-  $sql = <<SQL;
-DELETE ontology_xref, object_xref, xref, dependent_xref
-FROM ontology_xref, object_xref, xref 
-LEFT JOIN dependent_xref on xref_id = dependent_xref_id
-WHERE ontology_xref.object_xref_id = object_xref.object_xref_id AND object_xref.xref_id = xref.xref_id
-SQL
+  $sql = "DELETE ox FROM object_xref ox, xref x, external_db e WHERE x.xref_id = ox.xref_id AND x.external_db_id =e.external_db_id and e.db_name = 'GO'";
```
ameya1981, line 77: Is the intention to delete only from object_xref or from other tables too?
In the earlier version, alongwith ontology_xref, object_xref, xref and dependent_xref records were deleted too(?) for all ontologies.
Not sure if the other deletions(if still required) are handled in some other fashion.

misc-scripts/xref_mapping/XrefMapper/XrefLoader.pm
```diff
@@ -65,21 +65,18 @@ sub update{
   #####################################
   print "Deleting all PROJECTIONs from this database\n" if $verbose;
 
-  my $sql = "DELETE es FROM xref x, external_synonym es WHERE x.xref_id = es.xref_id and x.info_type = 'PROJECTION'";
+  my $sql = "TRUNCATE table ontology_xref";
```
ameya1981, line 68: Truncate would reset the auto_increment counter - hope that value is not being used anywhere else

misc-scripts/xref_mapping/XrefMapper/XrefLoader.pm
magaliruffier: as all the jobs are now run in the eHive pipeline, we don't actually see any of the delete statements, apart for the rare cases where a job is run in standalone mode, which is why it seemed unnecessary here
I believe a truncate job also does not actually say how many rows are affected, unlike a 'delete from table' query, which is why it is so much faster

misc-scripts/xref_mapping/XrefMapper/XrefLoader.pm
magaliruffier: the GPAD loading pipeline, which now deals with updating the ontology_xref entries, has a clean up step at the start and end of the pipeline
the step here is redundant with the GPAD loading pipeline and quite heavy on the database

misc-scripts/xref_mapping/XrefMapper/XrefLoader.pm
magaliruffier: not to my knowledge

ameya1981 (approved):

PR #510, opened 2020-08-26
ENSCORESW-3534: xref Database is also a DBConnection object

magaliruffier:
## Requirements

## Description

Convert Database object to inherit from DBConnection

## Use case

DBConnection has a large number of methods implemented to deal with database connection management.
the Database object used in the xref pipeline is created from the DBConnection object
by adding the ISA, the Database object also inherits all the methods from the DBConnection object
 
## Benefits

The Database object can use methods like disconnect_if_idle 

## Possible Drawbacks

NA
## Testing

Tried running the ParseSource runnable by adding disconnect_if_idle onto the newly created Database object
This fails with the master code (disconnect_if_idle not available) but works with this code change

### DIFF
--- a/misc-scripts/xref_mapping/XrefParser/Database.pm
+++ b/misc-scripts/xref_mapping/XrefParser/Database.pm
@@ -28,6 +28,10 @@ use Bio::EnsEMBL::DBSQL::DBConnection;
 use File::Spec::Functions;
 use IO::File;
 use English;
+use vars qw(@ISA);
+use strict;
+
+@ISA = qw( Bio::EnsEMBL::DBSQL::DBConnection);
 
 sub new {
   my ( $proto, $arg_ref ) = @_;


ameya1981 (commented):
Adding the parent is ok.
Can the flag of  disconnect_if_idle be set when creating the DbConnection object at line 44?
Will that avoid explicit calls later?

ens-bwalts (approved):
Since this seems to work, it looks like a good solution.

In eHive, we created a CoreDBConnection wrapper to sub in the methods from `Bio::EnsEMBL::DBSQO::DBConnection` - https://github.com/Ensembl/ensembl-hive/blob/version/2.5/modules/Bio/EnsEMBL/Hive/DBSQL/CoreDBConnection.pm

However, this was due to the need to decouple eHive from the rest of the Ensembl codebase, which I don't think is necessary here. So a simple solution is probably sufficient.

magaliruffier:
> Adding the parent is ok.
> Can the flag of disconnect_if_idle be set when creating the DbConnection object at line 44?
> Will that avoid explicit calls later?

disconnect_if_idle is useful for jobs that are IO or memory limited and do not need a constant connection to the mysql server
it is however discouraged if the job regularly connects back to the database, as it will cause more connections
having the option to enable it for the IO/memory intensive jobs and leave the default behaviour for query intensive jobs enables us that flexibility

ameya1981 (approved):
Approved post clarification

PR #512, opened 2020-09-14
Feature/ubuntu 20

mira13:
## Description

The changes performed to make ensembl tests compatable with Ubuntu 20, Mysql 8
## Use case

SOme test are failing for Ubuntu_20 and mysql8, so it'a fix for them

## Benefits

It's a step towards new system

## Possible Drawbacks

Not observed
## Testing

Yes, they pass with new ensembl-test repository changes

yes


mira13:
@ens-bwalts This pr is essential now to keep master branch building fine

mira13:
All ensembl tests will fail until this branch is merged. 

The matter is in compressed databases, i have changed them from binary to hex in ensembl, and the same - loading for that databases in ensembl-test changed to hex. So ensembl-test is merged, and ensembl is not.

magaliruffier (changes requested):
Same comment as on https://github.com/Ensembl/ensembl/pull/516 regarding quoting either only 'rank' or all columns
Otherwise, the changes look fine to me, having an explicit order by in the methods makes a lot of sense

magaliruffier (approved):
changes look sensible

PR #515, opened 2020-09-24
Quotes around rank

mira13:
## Description

Quotes around rank in sql requests

## Use case

Rank without quotes doesn't work for Mysql8
## Benefits

Able to run Mysql8 tests 

## Possible Drawbacks

Not found

## Testing

No

Yes, all local tests are fine


magaliruffier (approved):
OK, subset of https://github.com/Ensembl/ensembl/pull/512

PR #516, opened 2020-09-24
Quotes around rank

mira13:
## Description

Quotes around rank in sql requests

## Use case

Rank without quotes doesn't work for Mysql8
## Benefits

Able to run Mysql8 tests 

## Possible Drawbacks

Not found

## Testing

No

Yes, all local tests are fine

### DIFF
--- a/sql/patch_50_51_d.sql
+++ b/sql/patch_50_51_d.sql
@@ -42,11 +42,11 @@ ALTER TABLE meta
 ALTER TABLE coord_system ADD COLUMN species_id INT(10) UNSIGNED NOT NULL DEFAULT 1 AFTER coord_system_id;
 
 -- Drop the indexes from coord_system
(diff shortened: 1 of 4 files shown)


magaliruffier (changes requested):
From what I understand, rank is a protected vocabulary in Mysql 8, hence the need to quote any 'rank' columns in our schema
I can see that in some cases, only rank has been quoted, but in other places, all columns for that table have been quoted.
For consistency, it would be better to either quote only 'rank' (probably the quickest solution), or all the columns once and for all (will take a bit more time but might also future proof us against any other protected names in MySQL)

mira13:
@magaliruffier  fixed, has put quotes for all fields at all places

mira13:
Same commit applied, all fields quoted

magaliruffier (approved):
OK, subset of https://github.com/Ensembl/ensembl/pull/512

PR #517, opened 2020-09-24
Draft: Feature/travis ubuntu20

mira13:
## Description

Travis.yml modifed to cut old unused closures and add ubuntu 20 - perl 30 jobs

## Use case

All builds will include ubuntu 20 now

## Benefits

We can move forward to new system

## Testing

No

All tests are running successfully

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,7 @@
-dist: trusty
-
 language: perl
 
-os:
-  - linux
+dist:
+  - trusty
 
 services:
   - mysql
@@ -11,6 +9,7 @@ services:
 perl:
   - '5.14'
   - '5.26'
+  - '5.30'
 
 env:
   matrix:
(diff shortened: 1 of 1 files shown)


magaliruffier:
Looking at the test matrix, we are still running:
- 5.14 on sqlite, no coveralls
- 5.26 on mysql, with coveralls
and have now added:
- 5.30 on mysql, no coveralls
- 5.30 on sqlite, no coveralls

Are we expecting different behaviours on sqlite based on the Perl version? Or could we skip the 5.30 on sqlite?

Each new PR will now trigger 4 separate jobs, so it's worth thinking about the right balance to test as many combinations as possible for regressions, while keeping the number of runs low

mira13:
Removed exceed job. Now only one job is added - for mysql 8 ubuntu 20

magaliruffier (approved):
Looks reasonable to me
One new compatibility test suite has been added to check trusty distribution and Perl 5.30

PR #520, opened 2020-09-25
ENSCORESW-3553: store RGD separately based on type of link

magaliruffier:
## Requirements

## Description

For RGD which have a direct link to an Ensembl stable ID, use a different source than RGD which have an inferred link via RefSeq

## Use case

RGD accessions can be linked to a RefSeq accession or directly to an Ensembl stable ID.
Storing these links as two separate xref entries means we can treat them separately and prioritise the better link when available.
Without the distinction, we can have the same accession mapped to two different stable IDs and one link is arbitrarily chosen over the other. In some cases, the chosen link is invalid and no link to RGD is kept.

## Benefits

More links to RGD, more reproducible, more reliable

## Possible Drawbacks

NA

## Testing

NA
NA
NA

The xref pipeline currently fails for rat because some links become invalid after loading into the core database.
With the proposed fix, the pipeline runs successfully and RGD xrefs are correctly mapped to Ensembl genes

### DIFF
--- a/misc-scripts/xref_mapping/XrefParser/RGDParser.pm
+++ b/misc-scripts/xref_mapping/XrefParser/RGDParser.pm
@@ -60,6 +60,14 @@ sub run {
     confess 'Need to pass source_id, species_id and files as pairs';
   }
 
+  my $source_sql = "select source_id from source where name = 'RGD' and priority_description = 'direct_xref'";
+  my $sth = $dbi->prepare($source_sql);
+  $sth->execute();
+  my ($direct_source_id);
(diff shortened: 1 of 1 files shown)


magaliruffier:
Same as https://github.com/Ensembl/ensembl/pull/518 but rebased against master for updated test suite

ens-bwalts (approved):
Looks like it does what it says in the change description.

ameya1981:
Code changes capture the "direct" source id for the direct linking between RGD and Ensembl.

ameya1981 (approved):
Correct source id captured for direct RGD linking

PR #522, opened 2020-09-25
Correct compressed dbs

mira13:
## Description

Correct compressed dbs, that fits to hex loading in ensembl-test

## Use case

Now test suit fails due to uncompateble compressed dbs, here is fix

## Benefits

Tests run correctly

## Possible Drawbacks

not found
## Testing

No

yes, all runs fine


magaliruffier (approved):
This change has already been merged into master, added to branches for backwards compatibility

PR #524, opened 2020-10-07
Refactor code in SliceAdaptor::fetch_by_region

nwillhoft:
## Description

There is code in SliceAdaptor::fetch_by_region() that has been marked for refactoring, specifically the code dealing with synonym/wildcard/fuzzy-matching. During work on creating a chromosome alias feature, I worked on refactoring this code into two new subroutines. This has led to splitting the original task into two (1. chromosome alias feature, and 2. refactoring code). See JIRA tickets ENSCORESW-3021 and ENSCORESW-3562.

## Use case

This refactored code will be used by any `SliceAdaptor::fetch_by_region` request where the `seq_region_name` user argument is **not** an exact match to a `seq_region_name` attribute in the relevant core database.

I understand that this code is used extensively by Ensembl users and therefore a note here to say that these changes will of course be extensively tested. (Use cases to be added here during this work.)

## Benefits

In brief, the removal of superfluous code and making the code more readable. 

During work on the chromosome alias feature, it became clear that the synonym/wildcard/fuzzy-match code needed to be refactored into subroutines to avoid large chunks of duplicate code being used in SliceAdaptor::fetch_by_region.

## Testing

Not yet added any.

Not yet run.

### DIFF
--- a/modules/Bio/EnsEMBL/DBSQL/SliceAdaptor.pm
+++ b/modules/Bio/EnsEMBL/DBSQL/SliceAdaptor.pm
@@ -2727,5 +2727,186 @@ sub _build_circular_slice_cache {
   $sth->finish();
 } ## end _build_circular_slice_cache
 
+=head2 fetch_by_seq_region_synonym
+
+  Args       : $cs, $coord_system_name, $seq_region_name, $start, $end, $strand, $version, $no_fuzz
+  Example    : $self->fetch_by_seq_region_synonym( $cs, $cs->name(), $seq_region_name, $start, $end, $strand, $version, $no_fuzz );
+  Description: fetches all the seq region synonyms (or uses wildcard) when requested
+  Returntype : Bio::EnsEMBL::SliceAdaptor, or none 
+  Exceptions : None
+  Caller     : general
+  Status     : (refactored from fetch_by_region)
+
+=cut
+
+sub fetch_by_seq_region_synonym {
+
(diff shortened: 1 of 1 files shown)


modules/Bio/EnsEMBL/DBSQL/SliceAdaptor.pm
```diff
@@ -2727,5 +2666,186 @@ sub _build_circular_slice_cache {
...
+  Returntype : Bio::EnsEMBL::SliceAdaptor, or none 
+  Exceptions : None
+  Caller     : general
+  Status     : (refactored from fetch_by_region)
+
+=cut
+
+sub fetch_by_seq_region_synonym {
+
+  my ( $self, $cs, $coord_system_name, $seq_region_name, $start, $end, $strand, $version, $no_fuzz ) = @_;
```
ens-bwalts, line 2683: I think this can be simplified to remove `$coord_system_name`. At this point, if you have a CoordSystem object, then you can get the name directly from it (with `$cs->name()`). The user-supplied `$coord_system_name` is not used anymore in this method, and we're past the point where we are checking for the special name "toplevel"

nwillhoft: Thanks @ens-bwalts , I've removed `$coord_system_name` from the input parameters to `fetch_by_seq_region_synonym` and modified the subroutine accordingly.

nwillhoft:
Closing this PR as it has since been broken down into smaller PRs.

PR #525, opened 2020-10-07
Update to annotation provider in EMBL/GenBank files

james-monkeyshines:
## Description
We used to have a meta_key, `provider.name` (and an associated `provider.url`), but this did not allow us to capture the fact that sometimes the assembly needs to be credited to one provider, and the gene annotation to another. Vertebrate databases used the field to indicate the annotation provider, and non-vertebrates used it for the assembly provider. To better represent these use cases, `provider.name` was replaced by two new meta_keys, `assembly.provider_name` and `annotation.provider_name` (https://www.ebi.ac.uk/panda/jira/browse/ENSINT-361).

In the comments of EMBL and GenBank files, the `provider.name` was formerly used to indicate the source of the annotation; this was replaced by `assembly.provider_name` in a previous PR (#506), but it would be more accurate to use `annotation.provider_name`, and only if that is undefined, fall back to `assembly.provider_name`.

Further, since an annotation can have multiple providers, it is good to list them all, rather than select one to include in the comments.

## Use case
For vertebrates, this change does not make much difference, because `assembly.provider_name` is typically not defined, so the code falls back to the generic 'Ensembl' in any case. But for non-vertebrates, which are typically annotated by non-Ensembl groups, this allows for more accurate attribution.

## Benefits
Better attribution for annotation in ftp files.

## Possible Drawbacks
None I can think of.

## Testing
Yes

Pass

Yes

### DIFF
--- a/modules/Bio/EnsEMBL/Utils/SeqDumper.pm
+++ b/modules/Bio/EnsEMBL/Utils/SeqDumper.pm
@@ -476,34 +476,8 @@ sub dump_embl {
   #References (we are not dumping refereneces)
   #Database References (we are not dumping these)
 
-  #Get annotation source 
-  my @providers = ();
-  my @provider_urls = ();
-
-  foreach ( @{$meta_container->list_value_by_key('annotation.provider_name')} ) {
-    push @providers, $_ if $_ ne '';
-  }
-  foreach ( @{$meta_container->list_value_by_key('annotation.provider_url')} ) {
-    push @provider_urls, $_ if $_ ne '';
-  }
-  if ( ! scalar(@providers) ) {
-    foreach ( @{$meta_container->list_value_by_key('assembly.provider_name')} ) {
-      push @providers, $_ if $_ ne '';
-    }
-    foreach ( @{$meta_container->list_value_by_key('assembly.provider_url')} ) {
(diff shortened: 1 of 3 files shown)


nwillhoft (changes requested):
Hi @james-monkeyshines , 

These changes look really useful in providing the accurate annotation and assembly metadata. I only have a couple of suggestions from looking through the code.

Could the new code that gets the annotation source information in `dump_embl()` and `dump_genbank()` be refactored into a new subroutine? For example, the lines 480-504/688-711 appear to be duplicated and could be refactored into one subroutine that returns the `@providers` and `@providers_urls` variables.

Another suggestion would be to add tests for 'assembly.provider_name' and 'assembly.provider_url' as this looks as though it would boost the coveralls coverage stats.

Thanks,
Natalie

ameya1981:
Hi @james-monkeyshines , the cpanfile may need an addition for Module::Refresh for the tests.

ameya1981 (approved):
Changes made are passing the tests. annotation.* keys get added to the meta table.

nwillhoft (approved):
New code refactored into new subroutine, and new tests added to increase testing coverage. Tests are passing the build.

PR #527, opened 2020-10-16
Create chromosome alias in CoordSystem object

nwillhoft:
## Description

As described in [JIRA ticket ENSCORESW-3021](https://www.ebi.ac.uk/panda/jira/browse/ENSCORESW-3021), sequence assembly data has changed in format in recent years, resulting in recent databases having only a single 'primary_assembly' coordinate system. This is compared to older assembly data with multiple coordinate systems including: 'chromosome', 'contig', 'scaffold'.

## Use case

Currently, a coordinate system named 'chromosome' may not exist in an Ensembl database, due to a change in sequence assembly formatting in databases created in the last few years. Examples include the Parus major and Clupea harengus databases, which contain a single coordinate system named 'primary_assembly'. If a user requests a chromosome slice object and there is no explicitly-named chromosome coordinate system in the query database, this new feature uses karyotype-based attributes in the database to identify a coordinate system to add an appropriate ('chromosome') alias to.

## Benefits

Where an Ensembl Core species database does not have coordinate system named chromosome, this code will create an alias to the appropriate coordinate system when a user requests a chromosome slice object, as long as the requested name has associated karyotype attributes.

## Possible Drawbacks

Once the tests have been updated (as mentioned below), there are no drawbacks that I can think of.

## Testing

Yes, tests have been added to `coordSystem.t`.

The tests in `coordSystem.t` pass.

The test suite is passing in the Travis-CI build.

### DIFF
--- a/modules/Bio/EnsEMBL/CoordSystem.pm
+++ b/modules/Bio/EnsEMBL/CoordSystem.pm
@@ -184,6 +184,10 @@ sub new {
     throw('The RANK argument must be a positive integer');
   }
 
+  if ( defined($alias_to) ) {
+    throw("The ALIAS_TO argument can only be defined as 'chromosome'") unless $alias_to eq "chromosome";
+  }
+
   $self->{'version'}        = $version;
   $self->{'name'}           = $name;
   $self->{'top_level'}      = $top_level;
@@ -393,6 +397,9 @@ sub rank {
 
 sub alias_to {
   my ($self, $alias_to) = @_;
+  if (defined($alias_to)) {
+    throw("The alias can only be set to 'chromosome'") unless $alias_to eq "chromosome";
+  }
   $self->{'alias_to'} = $alias_to if defined $alias_to;
   return $self->{'alias_to'};
 }
--- a/modules/Bio/EnsEMBL/CoordSystem.pm
+++ b/modules/Bio/EnsEMBL/CoordSystem.pm
@@ -110,6 +110,8 @@ use vars qw(@ISA);
                              coordinate system
                -ADAPTOR   - (optional) The adaptor which provides database
                             interaction for this object
+               -ALIAS_TO  - (optional) Sets an alias for a coordsystem. If set 
+                            it should only be set to 'chromosome'
   Example    : $cs = Bio::EnsEMBL::CoordSystem->new(-NAME    => 'chromosome',
                                                     -VERSION => 'NCBI33',
                                                     -RANK    => 1,
--- a/modules/Bio/EnsEMBL/CoordSystem.pm
+++ b/modules/Bio/EnsEMBL/CoordSystem.pm
@@ -132,10 +132,11 @@ sub new {
 
   my $self = $class->SUPER::new(@_);
 
-  my ( $name, $version, $top_level, $sequence_level, $default, $rank ) =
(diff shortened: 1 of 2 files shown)


nwillhoft:
The test suite is passing when run by travis-ci. As various tests failed when I ran the test suite locally, I need to check how my local environment differs to that used by travis-ci.

magaliruffier:
Two areas to investigate regarding the tests:
- patches and other alternative regions have chromosome as a coordinate system but are flagged as alternative with an entry in the assembly exception table
It might be worth checking whether the alias fetching SQL needs to be updated to deal with these
- some tests can run fine on their own, but fail when the whole test suite is run.
When the whole test suite is run, test databases are re-used between tests rather than created from scratch
If any of the tests adds/updates data, this can affect downstream tests

nwillhoft:
> Two areas to investigate regarding the tests:
> 
> * patches and other alternative regions have chromosome as a coordinate system but are flagged as alternative with an entry in the assembly exception table
>   It might be worth checking whether the alias fetching SQL needs to be updated to deal with these
> * some tests can run fine on their own, but fail when the whole test suite is run.
>   When the whole test suite is run, test databases are re-used between tests rather than created from scratch
>   If any of the tests adds/updates data, this can affect downstream tests

Thanks mag. I believe the failed tests I saw locally were due to your second point of test dbs being re-used because if I only run the tests in `assemblyException.t`, they all pass as expected.

modules/Bio/EnsEMBL/CoordSystem.pm
```diff
@@ -132,10 +132,11 @@ sub new {
 
   my $self = $class->SUPER::new(@_);
 
-  my ( $name, $version, $top_level, $sequence_level, $default, $rank ) =
+  my ( $name, $version, $top_level, $sequence_level, $default, $rank, $alias_to ) =
```
ens-bwalts, line 135: Add the -ALIAS_TO option to the documentation

ens-bwalts (changes requested):

mira13:
Please check and if needed add appropriate changes to coordinate system apadtor that creates, removes and updates coordinatesystem object, to also support alias_to

mira13 (changes requested):
Please update CoordinateSystemAdaptor store/update/remove methods if needed to support new field.

nwillhoft:
> Please update CoordinateSystemAdaptor store/update/remove methods if needed to support new field.

Thanks @mira13 . I've discussed this with @ens-bwalts and alias-related changes to the database are currently planned for a future date. For this current feature we've been working on updating the CoordSystem and SliceAdaptor objects, and will include your suggestions in a plan to update database schema changes etc. at a future date.

mira13:
@nwillhoft it can happen that current changes will be merged in current release, and future planned - not. It is a bug in release.

nwillhoft:
Have discussed database schema updates with @magaliruffier and @ens-bwalts. They have decided that the database schema will need updating at a future date and have approved the scope of the current changes (i.e. code changes to modify CoordSystem objects). @mira13 's point needs keeping in mind - that users might file bug reports if they find a new alias field in the code and expect to find it in the database (but it's not there).

nwillhoft:
Will add in more checks to ensure users can only create an alias called 'chromosome'.

mira13 (approved):

ens-bwalts (approved):
Changes as requested

PR #531, opened 2020-10-20
Fix error msg, to avoid DBI error on connection error

mira13:
## Description

Change error msg: don't propose install DBI if it is just connection problem

## Use case

WHen user doesn't have mysql connection correct, now error msg proposees to install DBI module, we must avoid it and propose to check connection

## Benefits

Clear error msg

## Possible Drawbacks

Not found

## Testing

No, this script is not a part of API

All tests works fine

### DIFF
--- a/misc-scripts/ping_ensembl.pl
+++ b/misc-scripts/ping_ensembl.pl
@@ -168,7 +168,7 @@ eval {
 if($error) {
   print "ERROR: Error detected when connecting to Ensembl!\n";
   
-  if($error =~ /Can't connect to MySQL server on/) {
+  if($error =~ /Can't connect to MySQL server on /) {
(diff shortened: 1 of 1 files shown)


magaliruffier (commented):
The added checks seem reasonable, we are able to provide more targeted error messages.
Can I just confirm why the 'Cannot connect to' error catching has been removed? Is that never triggered or are there older versions that could spawn that type of error message?

mira13:
@magaliruffier  the msg is still there, just moved to "elseif" . So now if we have "connect to Mysql" error - we distinglish it from "connect to" and from DBI.

magaliruffier (approved):
That makes sense, sorry for missing it first time around

PR #533, opened 2020-10-29
ENSCORESW-3586: single list of filters for all vertebrate species

magaliruffier:
## Requirements

## Description

Consolidate all description filters into a single place

## Use case

Descriptions can contain text that is not useful as a description. Various regexes are used to filter these out when assigning gene descriptions.
There are several different lists applied depending on the species.
There is no good reason that a filter human uses is not applicable to another vertebrate species.
Having a single filter makes it easier to ensure it is applied consistently across species, and updates happen only in one place.

## Benefits

More consistent filter application across species.
Single point to update if additional filters are required

## Possible Drawbacks

Some filters might not be desired in some species, although I can't think of a good example why

## Testing

There are no test cases for the xref pipeline
The whole pipeline has been run to check for potential errors
Individual modules have been checked for compilation errors

The pipeline ran successfully

NA


ameya1981 (approved):
Changes are quite straight forward.

ens-bwalts (approved):
Looks like they all got incorporated and they all look valid (and they're alphabetised, too!). Having consolidated the list, it looks like there are some patterns that may be redundant (like `'^SIMILAR TO PUTATIVE[ \.]'` and `'SIMILAR TO PUTATIVE[ \.]'` but that could be a future cleanup task.

PR #537, opened 2020-11-04
ENSCORESW-3587: apply ignore filter for gene description

magaliruffier:
## Requirements

## Description

The ignore filter that is used for gene display xref assignment is also used for gene description

## Use case

When assigning gene display xref, a series of filters are applied to ignore certain xrefs, particularly if they come from predicted sources or have unhelpful names (LOC* for example)
The proposed code change re-uses this same filter when assigning the gene description.

## Benefits

The symbol and description used for the same gene are consistent.

## Possible Drawbacks

Some genes that used to have a description will not have one any more
Typically, these are genes that had a clone name or no symbol assigned

## Testing

There are no test cases for the mapping part of the xref pipeline.
The whole pipeline has been tested on zebrafish, with and without the proposed code changes.

The pipeline runs through completion.
Genes that have descriptions in the updated run match the descriptions from before the code change.
Some genes that had a description before the code change do not have a description in the updated run. These are all genes with a clone name

NA

### DIFF
--- a/misc-scripts/xref_mapping/XrefMapper/DisplayXrefs.pm
+++ b/misc-scripts/xref_mapping/XrefMapper/DisplayXrefs.pm
@@ -46,7 +46,7 @@ sub gene_description_sources {
   my $self     = shift;
 
   my ($precedence, $ignore) = @{$self->gene_display_xref_sources()};
-  return @$precedence;
+  return [$precedence,$ignore];
 
 }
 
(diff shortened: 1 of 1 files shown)


ameya1981 (approved):
Mainly refactor and call to gene_description_sources updated as per new return values.

nwillhoft (approved):
This PR refactors code into a new subroutine, `DisplayXrefs::_apply_ignore`, so that it can be reused to ignore certain xrefs. This provides functionality to filter gene description assignments, in addition to filtering gene display assignments. Changes have been verified by the pipeline running to completion.

PR #539, opened 2020-11-04
ENSCORESW-3548: don't add clone names if nothing else available

magaliruffier:
## Requirements

## Description

Remove clone name assignment step in OfficialNaming module.

## Use case

For merged species (human, mouse, rat, zebrafish), the OfficialNaming module adds clone names when no xref is suitable to be used as display xref.
These clone names are not useful any more and it is better to leave the genes without a display xref, defaulting to the Ensembl stable ID instead.
The proposed code change removes the step assigning those clone names

## Benefits

No more clone names.
OfficialNaming code is simplified.
Display_xref assignment is quicker, as the last step is removed.

## Possible Drawbacks

Some genes have no display_xref and default to stable ID instead

## Testing

There are no test cases for the mapping part of the xref pipline.
The whole pipeline has been run on all four merged species twice, with the code on master and with the proposed code changes.

The pipeline runs through to the end with the proposed changes.
The assigned display xref are comparable between the two runs, apart from the clone names which are not assigned when running the updated code.

NA

### DIFF
--- a/misc-scripts/xref_mapping/XrefMapper/OfficialNaming.pm
+++ b/misc-scripts/xref_mapping/XrefMapper/OfficialNaming.pm
@@ -211,7 +211,6 @@ SQ0
   my %xref_added; # store those added  $xref_added{$accession:$source_id} = $xref_id;
   my %seen_gene;
 
-  my %ens_clone_genes;
   my %official_name_used;
 
   my $ignore_sql =<<IEG;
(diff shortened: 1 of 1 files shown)


ameya1981 (commented):
The changes look fine. Though the description at the start of the file may need an update with clone names not being used.

nwillhoft:
Changes look fine. This PR has removed the clone name assignment step in the OfficalNaming module because clone names have been replaced by the more-useful Ensembl stable id. Two subroutines have been removed (`set_transcript_and_gene_display_xref_via_clone_name` and `get_clone_name`) as well as the calls to these subroutines.

nwillhoft (approved):
Changes look fine. This PR has removed the clone name assignment step in the OfficalNaming module because clone names have been replaced by the more-useful Ensembl stable id. Two subroutines have been removed (`set_transcript_and_gene_display_xref_via_clone_name` and `get_clone_name`) as well as the calls to these subroutines.

ameya1981 (approved):
Approved for clone name removal.

PR #546, opened 2020-11-19
Fix/metadata db versionned

marcoooo:
## Description

Fix a unwanted warning message about specific metadata database naming error: (from ENSPROD-6279 for context):

> No database version for database ensembl_metadata_qrp . You must be using a post version 34 database with version 34 or later code.
> You need to update your database or use the appropriate Ensembl software release to ensure your script does not crash
> For ensembl_metadata_qrp@mysql-ens-meta-prod-1 there is a difference in the software release (101) and the database release (0). You should update one of these to ensure that your script does not crash.

## Use case

When using metadata report_genome.pl script but this may apply to any script using registry against Rapid Release or GRCH37 dedicated metadata databases.

## Benefits

No more warning! :-)

## Possible Drawbacks

None

## Testing

_Have you added/modified unit tests to test the changes?_ No

_Have you run the entire test suite and no regression was detected?_ No wait for travis to do it.

### DIFF
--- a/modules/Bio/EnsEMBL/Registry.pm
+++ b/modules/Bio/EnsEMBL/Registry.pm
@@ -2919,7 +2919,8 @@ sub version_check {
     }
 
     # ensembl_metadata was unversioned prior to release 96
-    if ( $dba->dbc()->dbname() eq 'ensembl_metadata' ) {
+      # we now have multiple pattern for metadata db name (qrp and grch37) - valid until we merge those.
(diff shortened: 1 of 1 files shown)


magaliruffier (approved):
Changes look reasonable to me.
I like that you've allowed the qrp and grch37 patterns to make it work for you, without making it too permissive for other naming
Thanks also for the added comment, future us will be grateful

ens-bwalts (approved):
As we discussed today, this fits the SOP for ensembl_metadata better.

PR #548, opened 2020-11-23
ENSCORESW-3609: only use best transcript if no best translation

magaliruffier:
## Requirements

## Description

Select best transcript only if no best translation found.

## Use case

When selecting the best RefSeq match, we consider first the best translation match, then the best transcript match.
In a few cases, we end up selecting a transcript that is not the best translation match.
This happens if the best translation is selected first, then overwritten by the next transcript, which has the same transcript score or better.

## Benefits

We choose the best translation match where available.

## Possible Drawbacks

NA

## Testing

There are no test cases for this parser. The code change was run on human and the reported mismatches were fixed.
The results were compared to a complete run on human prior to the code change, the same number of transcripts were matched and there were a few more peptides matches.

NA
NA

### DIFF
--- a/misc-scripts/xref_mapping/XrefParser/RefSeqCoordinateParser.pm
+++ b/misc-scripts/xref_mapping/XrefParser/RefSeqCoordinateParser.pm
@@ -359,9 +359,12 @@ sub run_script {
                   $best_score = $score;
                 }
               }
-            } elsif ($score >= $best_score) {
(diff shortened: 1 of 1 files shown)


ens-bwalts (approved):
Changes work as described in the description - will prevent overwriting a best translation score.

ameya1981 (approved):
Changes look ok.

PR #551, opened 2021-01-11
Bugfix/access karyotype cache fix

nwillhoft:
## Description

Adding fix to retrieving data from karyotype cache.

## Use case

Printing a slice from Parus major (for example), or any species with a 'primary_assembly' coordinate system aliased with 'chromosome' will currently return an empty 'end' string in the Slice object. The karyotype cache has a data structure to allow for the storage of multiple coordinate systems with karyotype attributes, from which one of these would be chosen to be aliased. This PR fixes a bug to correctly retrieve data from the karyotype cache.

## Benefits

Bug fix. No more undef in the Slice object that should not be undef.

## Possible Drawbacks

None that I'm aware of.

## Testing

Yes.

The tests pass.

Test build passes.

### DIFF
--- a/modules/Bio/EnsEMBL/DBSQL/SliceAdaptor.pm
+++ b/modules/Bio/EnsEMBL/DBSQL/SliceAdaptor.pm
@@ -294,13 +294,20 @@ sub fetch_by_region {
   # use $key to access karyotype cache and define $arr
   if ( defined($key) && ($key eq "karyotype_cache") ) {
     
+    my $coord_system_id = $cs->dbID();
+    
     # retrieve values from karyotype cache
-    my $coord_system_id = $cs->{$key}{$seq_region_name}{'coord_system_id'};
-    my $seq_region_id   = $cs->{$key}{$seq_region_name}{'seq_region_id'};
-    $length             = $cs->{$key}{$seq_region_name}{'length'};
+    my $seq_region_id   = $cs->{$key}{$seq_region_name}{$coord_system_id}{'seq_region_id'};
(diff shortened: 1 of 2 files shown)


magaliruffier (commented):
End is now correctly set by using the value of length if undef
What does the change in coord_system_id assignment do?

ens-bwalts (commented):
I have the same question as mag regarding how the coord_system_id is getting fetched. I'm guessing this is to handle the case where we have multiple coord_systems that can be aliased to chromosomes (if we have different assemblies in the same DB for example) - is that correct?

nwillhoft:
Thanks a lot for the reviews @magaliruffier and @ens-bwalts. 

@ens-bwalts this is indeed to handle cases where we have multiple coordinate systems that could be aliased to chromosomes.

Just to try and summarise the functionality previously added in (which has caused this bug) for multiple coordinate systems with karyotype attribs - If there are multiple coordinate systems with karyotype attributes, _create_chromosome_alias() chooses the coordinate system (CS) with the highest rank (lowest number) and sets the ‘chromosome’ alias in that chosen CS. It also stores a karyotype cache in the CS to retrieve chromosome data later on. This coordinate system object is then returned by the subroutine.

If the CS has a karyotype cache and this key is being used to define $arr (and the slice object), I had understood that I can access the $coord_system_id of the CS that has been aliased by using the $cs->dbID method. I then use this CS id to extract the relevant data from the karyotype cache.

magaliruffier (approved):
Thanks for the explanation, sounds reasonable to me

ens-bwalts (approved):

PR #556, opened 2021-01-26
Add test for nowarnings to catch Undefined value error

mira13:
## Requirements

Small adition, to check no unexpected warnings. Was proposed in August, but was waiting until compara resolve their issue and make this test pass.

## Description
The issue is -
Sometimes we had undetected warnings in compara tests, and we cought them manually by looking through build. That should be done automatically.

A code changes performed - is only about the code block that makes expected warning is moved under expected warning cover without changes.

## Benefits

Now all failed test will be catched

## Possible Drawbacks

Not found

## Testing

Yes, it's test modification

YEs

### DIFF
--- a/modules/t/compara.t
+++ b/modules/t/compara.t
@@ -1,12 +1,12 @@
 # Copyright [1999-2015] Wellcome Trust Sanger Institute and the EMBL-European Bioinformatics Institute
 # Copyright [2016-2021] EMBL-European Bioinformatics Institute
-# 
+#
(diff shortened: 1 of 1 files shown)


nwillhoft (approved):
This PR adds in a warnings_exist clause to `compara.t`to catch any expected warnings being thrown. Excess white space characters at the top of the file have also been removed. The build checks pass.

TamaraNaboulsi (approved):
This passes the build checks and looks good to me. It only adds a few lines of code to catch expected warnings, surrounding the test code.

PR #561, opened 2021-02-03
add allele_string for variation_features

at7:
## Description
The sliceVariation.t was throwing a warning "Use of uninitialized value $allele_string in substitution (s///) at ...VariationFeatureAdaptor.pm line 1413." because of missing allele_string data in the variation feature table.

## Use case

We need to test fetching variation data from slice objects.

## Benefits

NA

## Possible Drawbacks

NA
## Testing

sliceVariation.t runs without warning messages.


mira13 (approved):
We requested this changes to avoid Uninitialized value error. Looks fine

nwillhoft (approved):
This PR updates a test data file used by `sliceVariation.t`. Values of `allele_string` in the test data file have been updated so that there’s no longer a warning about an uninitialised value. Build check passes.

PR #564, opened 2021-02-08
Add no warnings to slice variation test

mira13:
## Description

Added no warnings check to Variation tests, for early catch of errors.

## Use case
Sometimes we have warnins at variation tests, that are errors, but still tests pass. So we need to add no warnings check

## Benefits

Early and automatic bug catch

## Possible Drawbacks

This repo will fail until versions are fixed

## Testing

it is test changes

yes

### DIFF
--- a/modules/t/sliceVariation.t
+++ b/modules/t/sliceVariation.t
@@ -1,12 +1,12 @@
 # Copyright [1999-2015] Wellcome Trust Sanger Institute and the EMBL-European Bioinformatics Institute
 # Copyright [2016-2021] EMBL-European Bioinformatics Institute
-# 
+#
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
-# 
+#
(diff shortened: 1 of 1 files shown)


nwillhoft (changes requested):
The module `Test::Warnings` is being added in to check that there are no warnings generated by the tests. I can see that this module was commented out 5 years ago as it was generating warnings on Variation-deprecated methods (https://github.com/Ensembl/ensembl/commit/4b933094c4f82c320b0e490749461c4bd66de9f1). Looking at the build logs, it doesn't look like adding this module back in is causing such errors, so this change seems fine. The only little thing I would suggest is to remove the old commented-out `Test::Warnings` line at line 20.

mira13:
@nwillhoft  thanks a lot, i didnt notice the commented line. Just fixed.

TamaraNaboulsi (approved):
Un-commented Test::Warnings so warnings are displayed in variation tests. Change seems good to me.

nwillhoft (approved):
Small requested change has been made, changes look good.

PR #568, opened 2021-03-15
Final clause check

mira13:
## Description

This check prevents adding ORDER BY twice

## Use case

The problem occurs when using otter with ensebmbl 104+
Otter has its own sorting for gene, exon and transcript adaptors, it makes conflict withh new added ORDER by in ensembl (due to mysql version upgrade, new myswl requiers explicit ordering)

## Benefits

We can use otter with new ensembl

## Possible Drawbacks

not found

## Testing

no, its a bug fix

all tests run fine


TamaraNaboulsi (approved):
Added a check on whether sql statements contain an ORDER BY clause, and only concatenating if neither or only one of them has the clause to avoid adding ORDER BY twice. Looks good to me.

sgiorgetti (approved):
Looks good to me!!

PR #571, opened 2021-06-03
add database type gene2phenotype

at7:
## Description

We introduced a new ensembl database type [gene2phenotype](https://github.com/Ensembl/ensembl/blob/release/104/modules/Bio/EnsEMBL/Registry.pm#L158) a few years ago. We have now setup a test database which we want to patch like our other ensembl databases before branch.

## Use case
Patch gene2phenotype test database before branch.

## Benefits
Use established system for updating the gene2phenotype test database.

## Possible Drawbacks
None
## Testing
I ran the schema_patcher.pl to patch ensembl-gene2phenotype database to 105 which worked successfully after adding the updates described in this PR.

### DIFF
--- a/misc-scripts/schema_patcher.pl
+++ b/misc-scripts/schema_patcher.pl
@@ -55,7 +55,7 @@ Usage:
   --pass / -p\tdatabase user password (optional, no default)
 
   --type / -t   restrict to database schema type
-                (i.e. core, compara, funcgen, variation, production or ontology)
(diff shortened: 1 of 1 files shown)


ens-bwalts (approved):
gene2phenotype is added as an option everywhere core|funcgen|variation show up. Documentation updated too. I did a quick look at every time $database is matched against a regex to see if the introduction of a digit in the schema type might confuse them, and they all look OK.

Plus, the script works against a test case.

mira13 (approved):
Looks fine for me, one db added

PR #573, opened 2021-07-21
Bug fix for generating new exon stable ids

nwillhoft:
## Requirements

## Description

I've recently used the `generate_stable_id.pl` script to create new stable ids for various genes/transcripts/exons/translations. I was not able to provide information for the `start` parameter and so required the program to find the maximum current stable id for each of these types.

I found 2 issues when trying to run this script on my data, which I've tried to fix in this PR.

1. When the start parameter is not defined and the program looks for the maximum gene stable id, I found that it would return 'LRG_999' and not an 'ENSG'-based stable id. I've added in a WHERE clause to the relevant SQL query to refine the search to 'ENS' stable ids, as I (perhaps wrongly?) assumed we are not interested in incrementing 'LRG'-ids.
2. The SQL UPDATE string generated for exons was incorrect as it started with 'ENSEE', rather than 'ENSE'. This was due to the current highest stable id being returned before the 'E' in the prefix could be removed.

## Use case

I've recently used the `generate_stable_id.pl` script to create new stable ids for various genes/transcripts/exons/translations. I was not able to provide information for the `start` parameter and so required the program to find the maximum current stable id for each of these types.

To give more context, I've used this script to create new stable ids for genes/transcripts/exons/translations transferred from X chromosome PAR regions onto Y chromosome PAR regions. These objects have been transferred onto the correct coordinates and stored in a test database, and now require new stable ids creating for each. This script is being used to generate the SQL I need to update the stable ids in my test database.

## Benefits

In this context, the script will now generate the correctly formatted stable ids for new exons.

## Possible Drawbacks

The assumption that gene stable ids start with 'ENS' may not be correct and needs clarifying. For example, at the moment the SQL query returns 'LRG_999' (when no start parameter is defined) and it may be that some users need to update LRG-based stable ids.

## Testing

No.

### DIFF
--- a/misc-scripts/generate_stable_ids.pl
+++ b/misc-scripts/generate_stable_ids.pl
@@ -145,7 +145,7 @@ sub get_max_stable_id_from_gene_archive {
   my ( $dbi, $type ) = @_;
 
   # Try to get from relevant archive.
-  my $sth = $dbi->prepare("SELECT MAX($type) FROM gene_archive");
+  my $sth = $dbi->prepare("SELECT MAX($type) FROM gene_archive WHERE stable_id LIKE 'ENS%'");
   $sth->execute();
 
   my $rs;
@@ -241,9 +241,7 @@ sub get_highest_stable_id {
   # G/T/P for gene etc. (Exon dealt with above.)
 
   my ( $prefix, $suffix ) = $max =~ /([a-zA-Z]+)([0-9]+)/;
-  if ( $type eq 'exon' ) {
-    $prefix =~ s/E$//;
-  } elsif ( $type eq 'gene' ) {
+  if ( $type eq 'gene' ) {
     $prefix =~ s/G$//;
   } elsif ( $type eq 'transcript' ) {
     $prefix =~ s/T$//;
--- a/misc-scripts/generate_stable_ids.pl
+++ b/misc-scripts/generate_stable_ids.pl
@@ -118,7 +118,7 @@ sub increment_stable_id {
   } elsif ( $stable_id =~ m/([a-zA-Z]+)/ ) {
     $prefix = $stable_id;
   } else {
-    die(   "unrecongnized stable_id format "
+    die(   "unrecognized stable_id format: $stable_id "
          . "- should match ([a-zA-Z]+)([0-9]+) or ([a-zA-Z]+) !!\n" );
   }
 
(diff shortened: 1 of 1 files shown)


misc-scripts/generate_stable_ids.pl
```diff
@@ -171,7 +171,7 @@ sub get_highest_stable_id {
 
   # Get highest stable ID from the relevant table.
 
-  my $sth = $dbi->prepare("SELECT MAX(stable_id) FROM $type");
+  my $sth = $dbi->prepare("SELECT MAX(stable_id) FROM $type WHERE stable_id LIKE 'ENS%'");
```
ens-bwalts, line 174: This is probably the best compromise for now. This script isn't set up to handle weird cases like LRGs or species that don't follow the typical schema (e.g. saccharomyces_cerevisiae). Anyone working with those stable ids needs a different script.

Although at the moment there aren't any LRGs in the human gene_archive, for consistency, you should make this change in `get_max_stable_id_from_gene_archive` (at line 148) as well.

misc-scripts/generate_stable_ids.pl
```diff
@@ -234,7 +238,7 @@ sub get_highest_stable_id {
   }
 
   # Assuming that this is a correctly formatted stable id -> remove the
-  # G/T/P/E for exon etc.
+  # G/T/P for gene etc. (Exon dealt with above.)
 
   my ( $prefix, $suffix ) = $max =~ /([a-zA-Z]+)([0-9]+)/;
   if ( $type eq 'exon' ) {
```
ens-bwalts, line 244: Could delete this to make the code more clear - execution will never reach this point if `$type eq 'exon'`

ens-bwalts (changes requested):
In addition to the inline comments: it would be kind to add a note under sub usage that this script really only works for "ENS..." type stable IDs. It doesn't work with LRGs or the weird external stable IDs that are seen in some species.

misc-scripts/generate_stable_ids.pl
nwillhoft: Thanks Brandon, I've updated the SQL for the gene archive query.

misc-scripts/generate_stable_ids.pl
nwillhoft: Thanks, I've deleted that exon-based clause.

ens-bwalts (approved):

TamaraNaboulsi (approved):
When finding the start to create new stable ids, added where clause to only consider ENS ids. Also, removing extra 'E' from generated id. Changes look good, tests are passing, and note was added to say that this only works for 'ENS..' ids.

PR #574, opened 2021-08-12
Bugfix/dump mysql

magaliruffier:
## Requirements

## Description

The dump_mysql script assumed direct access to the MySQL server
As this is not often the case in production environment, the script has been updated to not use server-restricted SQL commands

## Use case

As part of the move to a new production infrastructure, access to MySQL servers has been restricted.
Some old scripts assumed direct access, which is less secure and now seldom allowed.
This change allows the script to be used from any environment against a remote SQL server

## Benefits

Script can be run in any environment

## Possible Drawbacks

The code is more complex

## Testing

No unit tests per say, the script was tested by Infrastructure and Variation though
Script runs successfully in the new infrastructure

NA

### DIFF
--- a/misc-scripts/db/dump_mysql.pl
+++ b/misc-scripts/db/dump_mysql.pl
@@ -359,8 +359,8 @@ sub data {
     "-e 'SELECT * FROM ${db}.${table}'",
     '|',
     'sed -r ',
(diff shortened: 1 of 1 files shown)


ens-bwalts (approved):
Thank you - that SELECT INTO OUTFILE has been a huge annoyance and source of confusion for a couple of years.

PR #576, opened 2021-08-17
Update hardcoded queue name and exonerate path so that it works on the new cluster.

ens-carlos:
## Requirements

## Description

Update hardcoded queue name so that it works on the new cluster.
## Use case

The hardcoded queue name is no longer used in the new cluster. It needs to be updated so stable ID mapping can be run.
## Benefits

Stable ID mapping will be able to be run again without modifying the queue name in the local copy of the source code.
## Possible Drawbacks

If stable ID mapping needs to be run in the old cluster, the queue name will have to be modified manually again in a local copy.
## Testing

N/A.

N/A.

### DIFF
--- a/misc-scripts/id_mapping/default.conf
+++ b/misc-scripts/id_mapping/default.conf
@@ -49,7 +49,7 @@ gene_score_threshold        = 0.125
 
 ;; Exonerate
 min_exon_length             = 15
-exonerate_path              = /nfs/software/ensembl/RHEL7/linuxbrew/bin/exonerate
+exonerate_path              = /hps/software/users/ensembl/ensw/C8-MAR21-sandybridge/linuxbrew/bin/exonerate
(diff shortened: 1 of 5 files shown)


ens-bwalts (approved):
Thank you for this PR. These changes look codon compatible for the moment, and I think it is OK to accept this as is given the urgency of the situation. It does bring a couple of issues to our attention -- there are a couple places where we should be fetching LSF options from the conf file instead of hardcoding them -- but that is our issue to clean up, not yours.

At the moment, one does not need to put in the `rusage` bit in bsub commands on codon. However, they do no harm and might prevent issues should systems decide to reconfigure LSF to require them in the future.

I did a quick check to see if there are any places where the change in tmp directory policy might cause issues, and didn't find any.  Lots of paths do depend on `basedir`, which is woefully out of date, but I get the impression that this gets overridden on the command line anytime this code is run.

PR #577, opened 2021-08-24
Allow multiple alt allele groups to be returned

nwillhoft:
## Requirements

## Description

A given gene id can be associated with one or more alt allele group ids. This relationship was previously defined as 1:1 in the core db, but is now a 1:m relationship as a result of this project work. The code now needs updating so that one or more alt allele group ids can be retrieved via the Perl API, which will allow the web pages to show all the relevant Gene Allele information (not just one alt allele group).

## Use case

Following my recent work on storing genes (transcripts/exons/translations) from Y chromosome PAR regions, the gene PPP2R3B now has alleles associated with two alternate allele groups: 1) a patch-based group, and 2) a PAR region-based group. Beforehand it only had alleles associated with the patch-based group. Both the core database and the Perl API currently assume that a given gene can only be associated with a single alternate allele group. I have modified a test core database to update this relationship from 1:1 to 1:m.

After testing the results in a web sandbox, it was found that the Gene Allele table shows data from a single alternate allele group, e.g. the alleles from the patch-based group or from a PAR region-based group. PPP2R3B is a good example of this, where the gene on the X chromosome shows 2 gene alleles from the patch-based group: http://ves-hx2-76.ebi.ac.uk:7065/Homo_sapiens/Gene/Alleles?db=core;g=ENSG00000167393;r=X:1-964539;t=ENST00000390665. And the gene on the Y chromosome shows a single gene alleles from the PAR region-based group: http://ves-hx2-76.ebi.ac.uk:7065/Homo_sapiens/Gene/Alleles?db=core;g=ENSG00000288704;r=Y:333933-386955;t=ENST00000681977.

The purpose of this work is to modify the code to bring in all the gene allele data together in this table. This will also allow users to retrieve all the gene allele data, not just the first alternate allele group the code comes across.

## Benefits

This work will allow the webpages to accurately show all gene allele data associated with a given gene. For example, if a gene is associated with more than one alternate allele groups, these will be displayed on the web pages. It will also allow users to retrieve such data via the Perl API.

## Possible Drawbacks

Can't think of any right now.

## Testing

Not yet.

### DIFF
--- a/modules/Bio/EnsEMBL/DBSQL/AltAlleleGroupAdaptor.pm
+++ b/modules/Bio/EnsEMBL/DBSQL/AltAlleleGroupAdaptor.pm
@@ -238,13 +238,28 @@ sub fetch_by_gene_id {
     my $sth = $self->prepare($gene_id_sql);
     $sth->bind_param(1,$gene_id, SQL_INTEGER);
     
+    my @group_ids;
     my $group_id;
     $sth->execute();
     $sth->bind_col(1,\$group_id);
-    $sth->fetch;
+    while ( $sth->fetch ) {
+        print "group id: $group_id\n";
+        push( @group_ids, $group_id );
+    }
+    print "group ids: @group_ids\n";
+    # $sth->fetch;
     $sth->finish;
     if (!$@ && $group_id) {
-        return $self->fetch_by_dbID($group_id);
+        my @aag;
+        foreach my $group (@group_ids) {
+            print "About to fetch aag object using group id $group\n";
+            my $aag = $self->fetch_by_dbID($group);
+            print "aag object: $aag\n";
+            push(@aag, $aag);
+        }
+        print "@aag\n";
+        return \@aag;
+        # return $self->fetch_by_dbID($group_id); # breaks if turned into an array
     }
     return;
 }


magaliruffier:
The main issue here is that we change a method returning a single alt_allele_group to returning a list of alt_allele_groups
This means that any code using this method would need to be updated after the change, not to mention that we are breaking the convention that fetch_by returns a single object whereas fetch_all returns a list
In the interest of backwards compatibility, could we create a new method, fetch_all_by_gene_id that uses the code you just wrote?
the fetch_by_gene_id method could even use the fetch_all method and return the first object, or have an extra option to specify which group to return if more than one, with a default group if nothing is provided
the aim here is to provide additional functionality while minimising disruption to existing code and pipelines

nwillhoft:
> The main issue here is that we change a method returning a single alt_allele_group to returning a list of alt_allele_groups
> This means that any code using this method would need to be updated after the change, not to mention that we are breaking the convention that fetch_by returns a single object whereas fetch_all returns a list
> In the interest of backwards compatibility, could we create a new method, fetch_all_by_gene_id that uses the code you just wrote?
> the fetch_by_gene_id method could even use the fetch_all method and return the first object, or have an extra option to specify which group to return if more than one, with a default group if nothing is provided
> the aim here is to provide additional functionality while minimising disruption to existing code and pipelines

Thanks a lot mag, that sounds like a much better idea to create a `fetch_all_by_gene_id` method that adds to the functionality rather than breaking a load of code and having to make numerous changes.

modules/Bio/EnsEMBL/DBSQL/AltAlleleGroupAdaptor.pm
```diff
@@ -231,20 +231,56 @@ sub fetch_by_dbID {
...
+  Returntype  : Array of Bio::EnsEMBL::AltAlleleGroup objects
+
+=cut
+
+sub fetch_all_by_gene_id {
+    my ($self, $gene_id) = @_;
+
     my $gene_id_sql = q(
         SELECT alt_allele_group_id FROM alt_allele
         WHERE gene_id = ?
```
ens-bwalts, line 260: The backwards-compatibility issue is a bit of a tricky one. This solution is probably as good as it gets, but I don't like that two calls to fetch_by_gene_id could return different alt-allele groups, depending on which one happens to be first in the list.  Consider adding an `ORDER BY alt_allele_group_id` here, so that the first alt-allele group in the returned list will be consistent. It may not be the one the user wants, but it will be much easier to track down the problem than if the error were intermittent!

magaliruffier: that's a very good point, and the order by should cover most of our use cases
in practice, as the PAR genes were added later, they will have a higher alt_allele_group_id, and we should not have that many cases in the first place (only happens for patch/haplotypes overlapping a PAR region)
we might get new cases in the future, although that will probably be a while

nwillhoft: Thanks very much for the ORDER BY suggestion. I've updated the SQL query to include this and therefore make the list of alt_allele_group_id values consistent.

mira13:
We can't be sure, that allele group that has smaller id is patch-based, and bigger - is PAR. Some other check, shoul be done, like comparing chr names. 
Also fetch_all method should return indicator which group is patch-based and which is PAR

modules/Bio/EnsEMBL/DBSQL/AltAlleleGroupAdaptor.pm
```diff
@@ -231,20 +231,56 @@ sub fetch_by_dbID {
...
+
     my $gene_id_sql = q(
         SELECT alt_allele_group_id FROM alt_allele
-        WHERE gene_id = ?
+        WHERE gene_id = ? ORDER BY alt_allele_group_id
     );
     my $sth = $self->prepare($gene_id_sql);
     $sth->bind_param(1,$gene_id, SQL_INTEGER);
     
+    my @group_ids;
```
EbiArnie, line 265: I recommend using fetchall_arrayref here. E.g.:
`   my $group_ids = $sth->fetchall_arrayref();
    $sth->finish;
    if (! $@ && @$group_ids) {
      ...
      foreach my $group (@$group_ids) {
      ...
    `

modules/Bio/EnsEMBL/DBSQL/AltAlleleGroupAdaptor.pm
```diff
@@ -231,20 +231,56 @@ sub fetch_by_dbID {
...
+    while ( $sth->fetch ) {
+        # print "group id: $group_id\n";
+        push( @group_ids, $group_id );
+    }
     $sth->finish;
     if (!$@ && $group_id) {
-        return $self->fetch_by_dbID($group_id);
+        my @aag;
+        foreach my $group (@group_ids) {
+            # print "About to fetch aag object using group id $group\n";
```
EbiArnie, line 277: Please remove this and the other debug comments.

EbiArnie (changes requested):

modules/Bio/EnsEMBL/DBSQL/AltAlleleGroupAdaptor.pm
```diff
@@ -231,20 +231,56 @@ sub fetch_by_dbID {
...
+            my $aag = $self->fetch_by_dbID($group);
+            # print "aag object: $aag\n";
+            push(@aag, $aag);
+        }
+        # print "@aag\n";
+        return \@aag;
     }
     return;
```
EbiArnie, line 285: I think we should return an empty array ref here. This is what other fetch_all methods do.
Otherwise this code in fetch_by_gene_id could try to dereference undef:
`my $aag = @$aag_list[0];`

EbiArnie (changes requested):

modules/Bio/EnsEMBL/DBSQL/AltAlleleGroupAdaptor.pm
nwillhoft: Added in, thanks

modules/Bio/EnsEMBL/DBSQL/AltAlleleGroupAdaptor.pm
nwillhoft: Removed, thanks

modules/Bio/EnsEMBL/DBSQL/AltAlleleGroupAdaptor.pm
nwillhoft: I believe array ref is now only returned, no undef

PR #578, opened 2021-09-06
Release/105 bugfix

magaliruffier:
## Requirements

## Description

Include sanity checking on stable ID version field
## Use case

According to the core schema, the version of a stable ID is a small int, which is an integer of maximum 5 digits.
In the API, there is no validation of this field, but it is used in an extra sql query that can be anything.
This can create very long queries that are not going to return anything as the input field is invalid.
Here, we add some checking on the value in the field before running the SQL query, providing a result faster without needlessly querying the database.

## Benefits

Queries get resolved quicker if input does not make sense.

## Possible Drawbacks

Might add a small performance penalty with the extra check.

## Testing

New tests have been added, both with valid and invalid inputs

Yes

Yes


ens-bwalts (approved):
Changes match those from draft PR. This prevents anything other than a <= 5 digit integer from finding its way into the query as it's built from the version number. Has tests that check for edge cases.

sgiorgetti (approved):

PR #580, opened 2021-09-09
Revert "add database type gene2phenotype"

sgiorgetti:
Reverts Ensembl/ensembl#571

Repo ensembl-gene2phenotype does not exist and changes are preventing the schema patcher to run properly.
Reverting changes temporarily to allow Release 106 code branching to proceed.


TamaraNaboulsi (approved):
Reverting change that was added by [#571](https://github.com/Ensembl/ensembl/pull/571) as it's leading the test db patcher to fail since it requires an ensembl-gene2phetotype repo which doesn't exist. Revert needed for this to work and change should be revised.

PR #586, opened 2021-10-20
Fix column name when selecting from gene_archive table

TamaraNaboulsi:
## Description

Same change as #585 added into release/106

## Use case

Fix sql query to use $type (which could be gene_stable_id, transcript_stable_id, ...) instead of 'stable_id' when selecting from gene_archive.

## Benefits

This script currently fails without this change.

## Possible Drawbacks

None

## Testing


ens-bwalts (approved):
Same as #585

PR #590, opened 2021-11-03
Fix/update trans canonical attr

TamaraNaboulsi:
## Description

Checking if a canonical attribute exists for a transcript and updating accordingly.

## Use case

When updating the db with canonical transcripts ids, we also need to update the newly added transcript attribute related to the canonical flag. We check if the new canonical transcript exists, we update if it does and insert if it doesn't. We also delete the old canonical attribute if it exists.

## Benefits

Updating canonical transcript data in the db.

## Possible Drawbacks

None, but changes to the API should accompany this.

## Testing

### DIFF
--- a/misc-scripts/canonical_transcripts/select_canonical_transcripts.pl
+++ b/misc-scripts/canonical_transcripts/select_canonical_transcripts.pl
@@ -260,6 +260,7 @@ if ($write) {
     # Prepare transcript canonical attribute sqls
     my $trans_select_sth = $dba->dbc->prepare("SELECT value FROM transcript_attrib WHERE transcript_id=? AND attrib_type_id=?");
     my $trans_update_sth = $dba->dbc->prepare("UPDATE transcript_attrib SET value=? WHERE transcript_id=? AND attrib_type_id=?");
+    my $trans_insert_sth = $dba->dbc->prepare("INSERT INTO transcript_attrib (transcript_id, attrib_type_id, value) values (?,?,?)");
     my $trans_delete_sth = $dba->dbc->prepare("DELETE FROM transcript_attrib WHERE transcript_id=? AND attrib_type_id=?");
 
     print "Updating database with new canonical transcripts...\n";
@@ -272,6 +273,9 @@ if ($write) {
         if (my ($new_canonical_exists) = $trans_select_sth->fetchrow_array()) {
           # Update new canonical transcript attribute
           $trans_update_sth->execute(1, $change->[1], $attrib_type_id);
+        } else {
+          # Insert new canonical transcript attribute
+          $trans_insert_sth->execute($change->[1], $attrib_type_id, 1);
         }
 
         # Check if old canonical transcript attribute exists
--- a/misc-scripts/canonical_transcripts/select_canonical_transcripts.pl
+++ b/misc-scripts/canonical_transcripts/select_canonical_transcripts.pl
@@ -258,9 +258,9 @@ if ($write) {
         print "Changin' ". $change->[1]. " on ". $change->[0]."\n" if $verbose;
         $gene_sth->execute( $change->[1], $change->[0]);
 
-	# Updating canonical flag in transcript_attrib table, if exists
-	$trans_select_sth->execute($change->[1], 554);
-	if (my @exists = $trans_select_sth->fetchrow_array()) {
+        # Updating canonical flag in transcript_attrib table, if exists
+        $trans_select_sth->execute($change->[1], 554);
+        if (my @exists = $trans_select_sth->fetchrow_array()) {
           if ($exists[0]) {
             $trans_update_sth->execute(1, $change->[1], 554);
           }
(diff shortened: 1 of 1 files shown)


ens-bwalts (changes requested):
There are a few things missing in this solution:
First, in addition to setting the canonical transcript_attrib for the new canonical transcript, you need to clear the canonical transcript_attrib from the old canonical transcript. Usually this is done by deleting the attrib, rather than setting it to zero. In order to make this simple, you probably will have to change `@change_list` to include the old canonical transcript dbID so that you don't have to search for it.

Second, if the transcript_attrib doesn't exist (if `trans_select_sth` doesn't find anything), then you need to insert the attrib.

Finally, although `attrib_type_id` almost never changes, it's safer and more maintainable to fetch it from the database instead of hard-coding it as 554.

sgiorgetti:
> Second, if the transcript_attrib doesn't exist (if trans_select_sth doesn't find anything), then you need to insert the attrib.

In general, it is true. In this case, Production are patching the core DBs inserting the transcript_attrib ... if we insert these with this script, couldn't we create issues downstream?

TamaraNaboulsi:
> 
> 
> There are a few things missing in this solution: First, in addition to setting the canonical transcript_attrib for the new canonical transcript, you need to clear the canonical transcript_attrib from the old canonical transcript. Usually this is done by deleting the attrib, rather than setting it to zero. In order to make this simple, you probably will have to change `@change_list` to include the old canonical transcript dbID so that you don't have to search for it.
> 
> Second, if the transcript_attrib doesn't exist (if `trans_select_sth` doesn't find anything), then you need to insert the attrib.
> 
> Finally, although `attrib_type_id` almost never changes, it's safer and more maintainable to fetch it from the database instead of hard-coding it as 554.

I didn't really consider the first point so thanks for pointing that out, I'll work on that. Also, the third point is a best-practices issue so thanks for that. As for inserting if not found, this is what we agreed to do because of the issue with production adding these attributes, and for them not getting a duplicate_key error when they try to insert after this script is run. We can discuss this further if you don't agree.

misc-scripts/canonical_transcripts/select_canonical_transcripts.pl
```diff
@@ -248,10 +248,36 @@
...
+    my ($attrib_type_id) = $attrib_type_id_sth->fetchrow_array() || 554;
+
     print "Updating database with new canonical transcripts...\n";
     foreach my $change (@change_list) {
         print "Changin' ". $change->[1]. " on ". $change->[0]."\n" if $verbose;
         $gene_sth->execute( $change->[1], $change->[0]);
+
+        # Check if new canonical transcript attribute exists
+        $trans_select_sth->execute($change->[1], $attrib_type_id);
+        if (my ($new_canonical_exists) = $trans_select_sth->fetchrow_array()) {
```
ens-bwalts, line 268: What happens here if the new canonical transcript does not have an `is_canonical` attrib, so no rows are returned by this query?
What happens if there is a row, but the value of `value` is 0?

TamaraNaboulsi: If no rows are returned, the code will not enter into the if statement and thus won't update anything, which I think is the desired outcome. As for the value, in my opinion it doesn't matter what it is, 0 or otherwise. The code will still go into the if statement and update the row even if the value is 0. Our concern here is with its existence and we want to update the value to 1 either way.

misc-scripts/canonical_transcripts/select_canonical_transcripts.pl
```diff
@@ -248,10 +248,36 @@
...
 
+    # Prepare transcript canonical attribute sqls
+    my $trans_select_sth = $dba->dbc->prepare("SELECT value FROM transcript_attrib WHERE transcript_id=? AND attrib_type_id=?");
+    my $trans_update_sth = $dba->dbc->prepare("UPDATE transcript_attrib SET value=? WHERE transcript_id=? AND attrib_type_id=?");
+    my $trans_delete_sth = $dba->dbc->prepare("DELETE FROM transcript_attrib WHERE transcript_id=? AND attrib_type_id=?");
+
+    # Get attribute id for canonical
+    my $attrib_type_id_sth = $dba->dbc->prepare("SELECT attrib_type_id FROM attrib_type WHERE code=?");
+    $attrib_type_id_sth->execute('is_canonical');
+    my ($attrib_type_id) = $attrib_type_id_sth->fetchrow_array() || 554;
```
ens-bwalts, line 259: attrib_type_id will not change within the same database, so this should be moved outside the foreach my slice loop. (In theory, it should be consistent across all databases because it's synchronised to `master_attrib_type` in the `ensembl_production` db). It also should not default to 554. If, for some reason, there is no `is_canonical` attrib in that databases `attrib_type` table, then I think the script should fail, rather than insert attribs with a type that isn't in `attrib_type`. This also guards against the `attrib_type_id` changing if for some reason Production has to change it.

TamaraNaboulsi: Having discussed this with Stefano, he does agree with you Brandon that the script should fail if for some reason we can't find an id for the is_canonical attribute. I was trying to make sure the script still works regardless of my changes (make it backwards compatible and all), but I see now that that isn't an issue. I have changed the code to reflect that.

PR #591, opened 2021-11-25
Feature/trans canonical attrib

TamaraNaboulsi:
## Requirements

## Description

JIRA ticket: [3844](https://www.ebi.ac.uk/panda/jira/browse/ENSCORESW-3844)

Recently, the attribute 'is_canonical' was introduced into the transcript_attrib table and it serves as a duplication to the canonical_transcript_id column in the gene table. This change it concerned with adding consistency to the APIs by having the canonical transcript attribute be updated both in gene and transcript_attrib tables whenever it changes in one of them. This change is also added into the select_canonical_transcripts.pl (which uses sql queries for now, but testing is still ongoing to consider using APIs instead). Finally, new tests were added in gene.t and transcript.t in order to test this change moving forward.

## Use case

- When the store or update methods for gene are called with a change to the canonical_transcript_id, the is_canonical attribute in the transcript_attrib table is changed (set to 1 for new canonical transcript and deleted for old canonical transcript).
- When the store or update methods for transcript are called with a change to the is_canonical attribute, the canonical_transcript_id in the gene table is also updated.
- Running the select_canonical_transcripts.pl script now also updates transcript_attrib table whenever the canonical transcript is updated in the gene table

## Benefits

Having consistency in the data.

## Possible Drawbacks

This would be my first real change to the core APIs, and thus I might have missed something or added changes in a way that is not usually done when updating the APIs. Thus, I would appreciate any feedback and am open to any fixes.

## Testing

Added new tests for store and update methods into gene.t and transcript.t. These tests succeed, and so does the full test suite (after adding the is_canonical attribute to the core and empty test-gennome-DBs for homo sapiens).

### DIFF
--- a/misc-scripts/canonical_transcripts/select_canonical_transcripts.pl
+++ b/misc-scripts/canonical_transcripts/select_canonical_transcripts.pl
@@ -48,6 +48,7 @@ my $write = 0;
 my $include_non_ref = 1;
 my $include_duplicates;
 my $verbose = 0;
+my $attrib_type_id;
 
 GetOptions( 'dbhost:s'            => \$host,
             'dbport:n'            => \$port,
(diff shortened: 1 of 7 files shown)


TamaraNaboulsi:
Test suite is now suddenly failing, even after reverting last change (before that it was succeeding). But in any case, I will keep this PR open for now just to get some feedback about the main changes and I will be investigating test failures.

PR #592, opened 2021-11-29
Bugfix: add compara patch required for wheat cultivars

JAlvarezJarreta:
## Description

Compara has included wheat cultivars protein trees in e106 and this has flagged a requirement to increase the column of one of our tables.

## Use case

To avoid clashes with the default protein trees, we prefix the name of these with `wheat_cultivars_`. As it usually happens, the largest value was 31 characters long (lucky us!). We have increased it to 50 just in case we have larger prefixes in the future, but this should be long enough to "never" have this issue again :crossed_fingers: 

## Benefits

We can now produce new varieties gene trees without concerns on this matter.

## Possible Drawbacks

None.

## Testing

This patch has been successfully applied to all Compara production pipelines and databases in e106 without raising any incompatibilities or issues.

### DIFF
--- a/modules/t/test-genome-DBs/multi/compara/SQLite/table.sql
+++ b/modules/t/test-genome-DBs/multi/compara/SQLite/table.sql
@@ -1,6 +1,6 @@
 --
 -- Created by SQL::Translator::Producer::SQLite
--- Created on Mon Jun 14 15:26:43 2021
+-- Created on Mon Nov 29 12:16:48 2021
 --
 
 BEGIN TRANSACTION;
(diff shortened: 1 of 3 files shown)


sgiorgetti (changes requested):
It looks like the patch file "patch_105_106_b.sql" is missing

JAlvarezJarreta:
@sgiorgetti, I might be wrong but I run the updates with the `ensembl-test` script and I think it is enough for the patch to be in `ensembl-compara` repository, `release/106` branch (which it is). Let me know if not where would you want it (we had some patches in previous releases and these are not in this repository either).

sgiorgetti:
@JAlvarezJarreta sorry ... getting older and multi-tasking is getting more and more problematic.
The patch file "patch_105_106_b.sql" is missing in the `master` branch of `ensembl-compara` repository. And the problem there is similar to the one for `ensembl` repo: both `master` branches have already bumped up for 107.

JAlvarezJarreta:
I don't think it is an age-related issue, but if it is, I'm getting older too :sweat_smile: But you are right, forgot about that important step, sorry! It should be all good know :+1:

sgiorgetti (approved):
Changes are applied to both SQLite and MySQL schemas, and meta.txt is properly updated.

Looks good to me

PR #597, opened 2021-12-15
Set $cs after fetch by fuzzy match

msmanoj:
## Requirements

## Description
After `_fetch_by_fuzzy_matching` is called, the value for `$cs` remains undef on line: https://github.com/Ensembl/ensembl/compare/release/105...msmanoj:patch-1#diff-cb19b7e2cd91a60f1ac5117d8722fe002e75279535767baff667131d1bb81612L430

This is because the $cs defined on L361 is only available within that block.

This change was introduced in the following commit:
https://github.com/Ensembl/ensembl/commit/948057af657426df9bd0cc10290a8ec36473f54d#diff-cb19b7e2cd91a60f1ac5117d8722fe002e75279535767baff667131d1bb81612

## Use case
This PR fixes the issue reported in the following ticket:
https://helpdesk.ebi.ac.uk/Ticket/Display.html?id=552711

## Benefits

## Possible Drawbacks

## Testing

### DIFF
--- a/modules/Bio/EnsEMBL/DBSQL/SliceAdaptor.pm
+++ b/modules/Bio/EnsEMBL/DBSQL/SliceAdaptor.pm
@@ -358,14 +358,15 @@ sub fetch_by_region {
         } else { # if no slice object with a match returned, try using a fuzzy match
           if (!$no_fuzz) {
 
-            my ($fuzzy_matched_name, $cs) = $self->_fetch_by_fuzzy_matching( $cs, $seq_region_name, $sql, $constraint, \@bind_params );
-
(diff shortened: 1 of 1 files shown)


ens-bwalts (approved):
The travis tests are never going to pass. That is because this is a PR from a fork, so Travis will only test the PR, not the merge with `release/105`. I have manually run the test suite and it passes, so I am comfortable with accepting this PR as is

TamaraNaboulsi (approved):
Change looks good and Brandon has run the test suite and all is passing. I'm not sure that the change at line 369 in needed since $cs is being set to $cs_new before that anyway. But in any case, it doesn't really affect anything.

PR #606, opened 2022-02-15
Fix package name for renamed parsers

TamaraNaboulsi:
## Requirements

## Description

Fixing package names for new xref parsers.

## Use case

The UniProt and RefSeq parsers used for the new pre-parsing step of the xref pipeline were renamed to include "Database" so as not to be confused with the old parsers and to have some separation between the pre-parsing step and the parsing step. At renaming, however, the package name wasn't updated to reflect the new name. This PR fixes that.
Also, these 2 files were skipped during the yearly license update (not sure why) and thus the yearly license update was also added here.

## Benefits

New xref download fails without this change.

## Possible Drawbacks

None

## Testing

No

N/A

Test suite succeeds


sgiorgetti (approved):
Yearly update correctly applied.
Package rename looks good.

PR #608, opened 2022-03-09
Use old parsers when needed

TamaraNaboulsi:
## Description

Making the xref config backwards compatible, to allow users to run the old Xref_update_conf pipeline without errors.

## Use case

Setting values for old parsers (used before new xref update) to be used if running the old xref pipeline or running the new one without pre-parsing.
The code only enters xref_config2sql.pl if no xref_source_dbi is provided, which is the case in the old pipeline or when pre-parsing is off.

## Benefits

Can run old pipeline with updated runnables without errors.

## Possible Drawbacks

N/A

## Testing

Ran the old pipeline (Xref_update_conf) with the 107 code and it ran without errors.

Note: Will need to add this change to main later on.

### DIFF
--- a/misc-scripts/xref_mapping/xref_config.ini
+++ b/misc-scripts/xref_mapping/xref_config.ini
@@ -1477,6 +1477,7 @@ source          = EntrezGene::MULTI
 source          = Reactome::MULTI
 source          = RNACentral::MULTI
 source          = RefSeq_dna::MULTI-vertebrate
+source          = RefSeq_peptide::MULTI-vertebrate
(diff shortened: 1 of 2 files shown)


sgiorgetti (commented):
New parameters are added to the 'ini' file for the old parsers.
Decision about the parser to use is via ternary operator with "defined" function: this makes the "old_parser" an alternative option, while the (new) parser is the default one.

Looks good, but want to catch up before approving.

marcoooo (approved):

PR #612, opened 2022-04-22
Xref/config bckwrds compat

TamaraNaboulsi:
## Description

Final backwards compatibility changes to allow running the new or old xref pipelines. 

## Use case

This change handles the creation of the central xref database and the intermediate xref_update database. It specifically uses the old parsers for RefSeq_dna and UniProt instead of the new ones if no pre-parsing is being performed. It also doesn't include RefSeq_peptide in the vertebrates sources if we're pre-parsing as in that case this will be included with the RefSeq_dna parsing.

Note: This change includes the changes added to release/107 in PR #608 

## Benefits

New pipeline can be run with pre-parsing set to 0, and old pipeline can be run successfully.

## Possible Drawbacks

None

## Testing

### DIFF
--- a/misc-scripts/xref_mapping/xref_config2sql.pl
+++ b/misc-scripts/xref_mapping/xref_config2sql.pl
@@ -40,6 +40,8 @@ use Config::IniFiles;
 my $file = (defined $ARGV[0] && -f $ARGV[0]) ? $ARGV[0] : 'xref_config.ini';
 warn "using ", $file;
 
+my $preparse = defined $ARGV[1] ? $ARGV[1] : 0;
+
 my $config = Config::IniFiles->new(-file => $file);
 if(! defined $config) {
   foreach my $e (@Config::IniFiles::errors) {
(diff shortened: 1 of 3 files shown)


misc-scripts/xref_mapping/xref_config2sql.pl
```diff
@@ -175,6 +177,8 @@
     my $source_section = sprintf( "source %s", $source_name );
     $source_section =~ s/\s$//;
 
+    next if ($preparse && $species_id == 7742 && $source_name eq 'RefSeq_peptide::MULTI-vertebrate');
```
sgiorgetti, line 180: Suggest to put a comment in the code to make explicit 7742 is related to vertebrates.
This should turn out useful also because "vertebrates" is no species.

sgiorgetti (commented):
Added $preparse parameter to handle new/old parsers in Database.pm
Added old parsers to the "ini" file.
Xref_config2sql looks good too. Pls, add in comment line as per my comment there
Otherwise, changes look good to me.

sgiorgetti (approved):
All look good now.

PR #613, opened 2022-05-06
Added the protein_coding_LoF biotype that replaces polymorphic_pseudo…

jmgonzmart:
…gene in human and mouse

## Requirements

## Description

The protein_coding_LoF biotype was introduced in release 107 to replace the polymorphic_pseudogene biotype at the transcript level in human and mouse. 
The TranscriptSelector module is used by the script that determines the canonical transcript of a gene based on an algorithm that prioritizes the 'coding' biotypes in its transcript set:
    misc-scripts/canonical_transcripts/select_canonical_transcripts.pl
The proposed changes give protein_coding_LoF transcripts a similar priority level than polymorphic_pseudogene transcripts.

## Use case

The following protein_coding genes would get a 'processed_transcript' (non-coding) canonical transcript instead of the protein_coding_LoF transcript without the proposed changes.
ENSG00000282181 
ENSG00000244682

## Benefits

By adding this new biotype in TranscriptSelector we avoid assigning the wrong canonical transcript to genes containing only protein_coding_LoF and non-coding (eg. processed_transcript and retained_intron) transcripts.

## Possible Drawbacks

None. 

## Testing

No.

No. I just run the select_canonical_transcripts.pl script after modifying my copy of TranscriptSelector.pm.

### DIFF
--- a/modules/Bio/EnsEMBL/Utils/TranscriptSelector.pm
+++ b/modules/Bio/EnsEMBL/Utils/TranscriptSelector.pm
@@ -144,6 +144,7 @@ my %biotype_priority = ('protein_coding' => 1,
                         'nonsense_mediated_decay' => 2,
                         'non_stop_decay' => 2,
                         'polymorphic_pseudogene' => 2,
+                        'protein_coding_LoF' => 2,
                         'other' => 3,
 );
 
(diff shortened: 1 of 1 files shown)


sgiorgetti (commented):
Hello @jmgonzmart - only one question.

The title of this PR seems to suggest that the new biotype is replacing an old one.
However, the old biotype "polymorphic_pseudogene" looks like being still there.
Is this the expected behaviour?

Thx

sgiorgetti (approved):
As per conversation with @jmgonzmart , we'll leave the old biotype as it can be still in use in species other than human.

PR #616, opened 2022-05-23
Trim the DOS/Windows line ending from files

TamaraNaboulsi:
## Description

Trimming the Reactome lines to remove the DOS/Windows newline character "\r".

## Use case

The Reactome files have recently changed their newline format (DOS/Windows). This creates a problem when parsing the lines and extracting the data from them, specifically the species name which happens to be at the end of the line.

A solution proposed by Jose Gonzalez (who discovered this problem) includes removing any non-word character "\W" at the end of the species name. However, this is limited to Latin characters and might cause problems if a species name has a special character at the end. I decided that specifically removing the "\r" instead is a safer way to go. In addition, if the Reactome files go back to the previous format, this won't break anything.

## Benefits

The Reactome parser can correctly read the species names.

## Possible Drawbacks

None

## Testing


sgiorgetti (approved):
The chomp is taking care of LF; so, only the CR is to be dealt with.

Good to me

PR #617, opened 2022-05-24
Trim the DOS/Windows line ending from files

TamaraNaboulsi:
## Description

Same as #616 but for the main branch instead.


EbiArnie:
OK

EbiArnie (approved):

PR #618, opened 2022-05-26
Remove X padding of 5' incomplete codon protein strings

nwillhoft:
## Requirements

## Description

A mis-match in data between Ensembl and UniProt has been identified and investigated by Andy Yates and Jane Loveland. mRNA sequences that have an incomplete codon at their 5' end (i.e. have a start phase of 1 or 2) currently have one or two Ns added in to complete the codon. These extra Ns are being translated into an X amino acid at the protein sequence level, which the Xref pipeline can no longer correctly map to a UniProt SwissProt entry (as it contains this extra X), where applicable. It is estimated that this currently affects around 9000 Xref entries.

The purpose of this PR is to remove these extra Ns for 5' incomplete codons.

## Use case

From Andy's email:

This was originally flagged by Ben and was in quite a prominent gene TRBC1 (ENSG00000211751.9). The protein (ENSP00000487742.1) should have been xref'd to P01850 but instead had the TrEMBL entry A0A5H1ZRT1 created. This is because there is an extra N introduced into the coding sequence to maintain frame for the translation. You can see this if you look at the exon page on Ensembl (screenshot included):
 
https://www.ensembl.org/Homo_sapiens/Transcript/Exons?db=core;g=ENSG00000211751;r=7:142791694-142793368;t=ENST00000633705
 
In order to become more “in-line” with UniProt the right thing to do here is to remove the extra X from the Ensembl protein. That will mean we match the UniProt SwissProt entry, which is good for everyone. You can see the API doing this addition of the extra Ns in the translatable_seq in that method where the number of extra Ns (remember this is not in the protein yet) is equal to the phase we’re in. Phase 1 means add 1 N, phase 2 means add 2 Ns.
 
https://github.com/Ensembl/ensembl/blob/release/106/modules/Bio/EnsEMBL/Transcript.pm#L918
 
This method is called by Transcript::translate():
 
https://github.com/Ensembl/ensembl/blob/release/106/modules/Bio/EnsEMBL/Transcript.pm#L2092

## Benefits

Removing these extras Ns at the 5' end of incomplete codons will allow our Xref pipeline to correctly match a protein to its UniProt SwissProt entry, where applicable, instead of generating a new TrEMBL entry.

## Possible Drawbacks

## Testing

Yes, a test has been added to `transcript.t`. One of the example transcripts and its associated data have been added to the relevant files. Unit tests have been updated to reflect the addition of these data.

All pass following my unit test updates.

No regression.


modules/t/transcript.t
```diff
@@ -226,6 +226,10 @@ $tr = $ta->fetch_by_stable_id( "ENST00000300425" );
 is( $tr->translate()->seq() =~ /P$/, 1, "Incomplete codon is not translated");
 is( $tr->translate(1)->seq() =~ /PL$/, 1, "Incomplete codon is padded then translated");
 
+# # get a new transcript to check incomplete 5' codon behaviour
+# my $n_padded_tr = $ta->fetch_by_stable_id( "ENST00000633705" );
+# is( $n_padded_tr->translate()->seq() =~ /^DLNK/, 1, "Incomplete 5' codon is not translated as an X, has been removed" );
```
TamaraNaboulsi, line 231: Hi @nwillhoft. Is this section meant to be commented out? Isn't this the new test being added for your change?

nwillhoft: Thanks a lot for pointing this out @TamaraNaboulsi, that section was not meant to be commented out.

nwillhoft:
Have rebased my branch to include recent Compara changes so that the build for the branch now passes. These now-incorporated changes are unrelated to this PR but as my branch was out-of-date, it was causing the branch build to fail. All checks are now passing.

TamaraNaboulsi (approved):
Removing the X amino acid that results from translating the N padding at the 5' end. A test case along with its data has been added (lots of other tests wee affected and that also was fixed). Looks good to me.

sgiorgetti (approved):
Removing the dummy X amino acid by ignoring the fake N padding at the 5' end. A test case along with its data has been added.

Looks good to me.

PR #619, opened 2022-06-16
Xref/new source genecards

TamaraNaboulsi:
## Description

A new xref source, GeneCards, was added into the xref pipeline. It is a dependent source and it depends on the HGNC direct mapping to Ensembl. Since the UniProt files also include mentions of the GeneCards source, a skip was necessary there as the number of xrefs we get from UniProt is almost half the ones we get from HGNC, and almost all of them are a subset of the HGNC ones. This change also includes an update to the HGNC test script.

The accession value for these GeneCards xref is set to only the HGNC ID (instead of "HGNC:ID". That was done in order to accommodate the way that web reads this info to display the xref link on the website.

Note: For adding the new source into the external_db tables, the patch created is suffixed with "_b" instead of "_a" to reserve the latter for the schema version bump to 109.

## Use case

We've been asked to reciprocate the GeneCards mapping.

## Benefits

Another xref source is added which enriches our xref data.

## Possible Drawbacks

None

## Testing

hgnc.t was modified to handle the new changes

tests are passing

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -22,7 +22,7 @@
     - apt-get install -y libssl-dev sqlite3
     - git clone --branch=main --depth=1 https://github.com/Ensembl/ensembl-test.git
     - git clone --branch=main --depth=1 https://github.com/Ensembl/ensembl-io.git
-    - git clone --branch=master --depth=1 https://github.com/Ensembl/ensembl-variation.git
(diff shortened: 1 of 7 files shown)


sgiorgetti (approved):
Looks good.
Thanks for fixing main branch for variation ;-)

PR #620, opened 2022-06-21
Missing code in RefSeq new parser

TamaraNaboulsi:
## Description

This change involves adding some code that was missed when the RefSeq parsers were updated to read data from a central xref Db instead of from the files at each run. Also, since this missing code relies on RefSeq_mRNA data which the same parser retrieves, we had to revert back to running the parser twice in a pipeline run: once for RefSeq_mRNA and RefSeq_ncRNA and another pass for RefSeq_peptide. In addition, a bug fix in the way the pairs table is accessed was needed.

Note: Some of the changes here are added as patches to fix the 108 run but will be fixed properly for 109 in another PR.

## Use case

This missing code handles the dependent EntrezGene and WikiGene xrefs which were directly affecting the number of transcript xrefs and leading to DC errors. In addition, passing through the parser once isn't sufficient as we thought when the parser was updated and we need to have 2 passes through it (like we did before the update).

## Benefits

CompareXref DC should now succeed for species which had this problem.

## Possible Drawbacks

None

## Testing

### DIFF
--- a/misc-scripts/xref_mapping/XrefParser/RefSeqDatabaseParser.pm
+++ b/misc-scripts/xref_mapping/XrefParser/RefSeqDatabaseParser.pm
@@ -40,32 +40,39 @@ sub run {
     croak "Need to pass source_id, species_id and xref_source";
   }
 
+  my $source_name = $self->get_source_name_for_source_id($source_id, $dbi);
   my @source_ids;
-  my $peptide_source_id =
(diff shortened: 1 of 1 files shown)


sgiorgetti (approved):
The 2-pass process is handled properly.
Typo - accession1/accession2 - is being corrected.

Good to me

PR #624, opened 2022-08-09
Fix/xref refseq

TamaraNaboulsi:
## Description

This change includes what was added in #620 with a few additions that were put in as a temporary patch in the change merged into release/108. This change includes the addition of code that was missed when the new RefSeq parsers (pre-parser and central db parser) were devised. This also fixes a typo in selecting from the pairs table from the central DB. In addition, the RefSeq parser needs to run twice, once for dna and once for peptide, in order to add all the relevant data, and thus this behavior was restored (after it was removed in #612 ). Lastly, this change also includes the addition of 2 indexes into the xref and pairs tables to make accessing the central DB quicker.

## Use case

The missing RefSeq parser code was leading to a decrease in the number of xrefs for EntrezGene, WikiGene, and RefSeq_peptide.

## Benefits

Correct xrefs are read from the central DB and inserted into the species xref DB.

## Possible Drawbacks

None

## Testing

No

### DIFF
--- a/misc-scripts/xref_mapping/XrefParser/RefSeqDatabaseParser.pm
+++ b/misc-scripts/xref_mapping/XrefParser/RefSeqDatabaseParser.pm
@@ -89,6 +89,8 @@ sub run {
   my ($xref_id, $accession, $version, $label, $description, $info_type, $parsed_seq, $type, $status, $dep_xref_id, $dep_accession, $dep_version, $dep_label, $dep_description, $dep_source_id, $dep_species_id, $linkage_source_id, $synonym, $refseq_pair);
 
   my @xrefs;
+  my @dep_refseq_ids;
+  my @dep_entrez_ids;
   my $count = 0;
   foreach my $xref_source_id (@source_ids) {
     $get_xref_sth->execute($species_id, $xref_source_id);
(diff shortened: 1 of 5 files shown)


EbiArnie:
This works and is OK as is.

But as just discussed, I have a few suggestions.

Suggestion: To make things faster, maybe try out if execute_array helps. Along these lines:

```perl
if (defined $refseq_pair and $refseq_ids{$refseq_pair}) {
              foreach my $refseq_id (@{ $refseq_ids{$refseq_pair} }) {
                foreach my $entrez_id (@{ $entrez_ids{$dep_accession} }) {
                  push @first, $refseq_id;
                  push @second, $entrez_id;
                } 
                foreach my $wiki_id (@{ $wiki_ids{$dep_accession} }) {
                  push @first, $refseq_id;
                  push @second, $entrez_id;
                }
              }
            }

my @tuple_status;
my $tuples = $add_dependent_xref_sth->execute_array(
    { ArrayTupleStatus => \@tuple_status },
    \@first,
    \@last,
);
if ($tuples) {
    print "Successfully inserted $tuples records\n";
}
else {...};

add_dependent_xref_sth->commit;
```

Idea: turn the lines with
```perl
my $mrna_source_id = $self->get_source_id_for_source_name('RefSeq_mRNA','refseq', $dbi);
 push @source_ids, $mrna_source_id;
```
into something along these lines:
```perl
my @array;
my $dbi = ...;
my $add = sub {
    my $src_name = shift();
    my $src_db = shift();
    push @array,  $self->get_source_id_for_source_name(...)
};
$add->( "mrna",  "refseq" );
$add->( "ncrna", "refseq" );
```


Idea: Make this shorter by passing a hash reference:
```perl
 $dep2{ACCESSION} = $dep_accession;
          $dep2{LABEL} = $entrez{$dep_accession};
          $dep2{VERSION} = $dep_version;
          $dep2{DESCRIPTION} = $dep_description;
          $dep2{SOURCE_ID} = $wiki_source_id;
          $dep2{LINKAGE_SOURCE_ID} = $linkage_source_id;
          push @{$xref->{DEPENDENT_XREFS}}, \%dep2;
```
```perl
push @{$xref->{DEPENDENT_XREFS}} , {
              ACCESSION => $dep_accession,
              LABEL => $dep_version,
              ...
              }
```

TamaraNaboulsi:
Thank you @EbiArnie for your suggestions. I have added the execute_array change as we're always looking for ways to speed up the xref parsing and the tests seem promising. A future optimization would be to collect more of these values to further make use of the execute_array functionality (will be adding this to the list of optimizations I'm creating). I've also applied your third suggestion for DEPENDENT_XREFS to make the code shorter and better-looking.

As for you second suggestion concerning the source IDs, I do agree that your idea will result in a more pretty code and eliminate redundancy. However, after looking into it further, it might make the code less readable. The xref code is already complicated and I didn't want to add to this extra complication, especially since future maintainers of it might not be well-versed in perl.

Thanks again and I appreciate your review!

EbiArnie (approved):
Looks good!

PR #627, opened 2022-08-25
Revert 618 bugfix/fix padding method

sgiorgetti:
## Description

Reverting changes from `release/109` because of CRUCS discussion and https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-6643
Keeping changes in `main` branch

## Use case

N/A

## Benefits

N/A

## Possible Drawbacks

Mismatches between Ensembl and Uniprot

## Testing

Test suite is reverted back to 108 version.


EbiArnie (approved):
OK

PR #628, opened 2022-08-27
Fixing refseq parser

TamaraNaboulsi:
## Description

Same as #624 but against release/109 branch


sgiorgetti (approved):
Same as #624 
Looks good

PR #629, opened 2022-09-07
Revert "Remove X padding of 5' incomplete codon protein strings"

nwillhoft:
Reverts Ensembl/ensembl#618 upon request from the Variation team. Their `postreleasefix/109` branch of `ensembl-variation` has travis tests pointing to the `ensembl` repo `main` branch, and their tests are failing as a result because they point to the `main` branch where the padding fix is in place.


sgiorgetti (approved):
18 files were changed by #618 and the same are reverted back here.
Looks good to me.
Let's wait for Travis to complete successfully, before merging these changes in.

PR #632, opened 2022-10-03
Removing insert statement

TamaraNaboulsi:
## Description

Due to some suspected conflict between this patch and the production database, the insert statement in the patch have been removed (commented out).


sgiorgetti (approved):
As agreed. Looks good

PR #634, opened 2022-10-06
Update ftp links to https

veidenberg:
This PR updates all ftp URLs to use https (to be compatible with web browsers).
The affected hostnames have been tested to work with https.
Other affected repos are listed in the JIRA ticket.

Webteam JIRA ticket: https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-6632
Sandbox: http://wp-np2-1d.ebi.ac.uk:1680

### DIFF
--- a/misc-scripts/xref_mapping/XrefParser/VGNCParser.pm
+++ b/misc-scripts/xref_mapping/XrefParser/VGNCParser.pm
@@ -32,7 +32,7 @@ XrefParser::VGNCParser
 A parser class to parse the VGNC source.
 VGNC is the official naming source for some vertebrates species
 
--data_uri = ftp://ftp.ebi.ac.uk/pub/databases/genenames/vgnc/tsv/vgnc_gene_set_All.txt.gz
+-data_uri = https://ftp.ebi.ac.uk/pub/databases/genenames/vgnc/tsv/vgnc_gene_set_All.txt.gz
 -file_format = TSV
 -columns = [
     taxon_id
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ To clone the Ensembl Core API, use the following command:
 ```
 git clone https://github.com/Ensembl/ensembl.git
 ```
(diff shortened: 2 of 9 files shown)


sgiorgetti (commented):
@veidenberg I am good and comfortable with all the changes but one (VGNCParser) I'd like to double check

misc-scripts/xref_mapping/XrefParser/VGNCParser.pm
```diff
@@ -32,7 +32,7 @@ XrefParser::VGNCParser
 A parser class to parse the VGNC source.
 VGNC is the official naming source for some vertebrates species
 
--data_uri = ftp://ftp.ebi.ac.uk/pub/databases/genenames/vgnc/tsv/vgnc_gene_set_All.txt.gz
```
sgiorgetti, line 35: @TamaraNaboulsi @veidenberg 
This URL does not work, whereas the HTTPS one does.
How could the Xref work? Is it a new/recent change (I obviously missed)?

veidenberg: Yes, many ftp URLs (including ones hosted by Ensembl) have been migrated to be accessible over https, because web browsers have dropped support for ftp:// protocol. The ftp:// version should still be accessible with an FTP client.

TamaraNaboulsi: Hi both. The VGNC link whcih xref uses to download the file is set to the correct http one, as you can see [here](https://github.com/Ensembl/ensembl-production/blob/main/modules/Bio/EnsEMBL/Production/Pipeline/Xrefs/xref_sources.json#L207). It seems like this comment in VGNCParser.pm wasn't updated when the link was fixed. I will make sure to fix these comments with the next xref changes.

TamaraNaboulsi (approved):
Change looks good

PR #635, opened 2022-10-07
fix translate

sgiorgetti:
## Description

The call to Bio:PrimarySeqI->translate() does not match the prototype in BioPerl 1.6.x, and should be revised

## Use case

Non-standard start codons are not translated properly, regardless the specified codon table.
Example: the microbial CDS sequence "**GTG**TTTTCCGAT..." translates to "**V**FSD..."

## Benefits

Fix

## Possible Drawbacks

N/A

## Testing


modules/Bio/EnsEMBL/Transcript.pm
```diff
@@ -2161,9 +2161,21 @@ sub translate {
...
+#       $complete_codons, $throw, $codonTable, $orf, $start_codon, $offset);
+# 
+# Temporarily hardcoding the two parameters below.
+# possibly to drive later on using *_start_NF or *_end_NF transcript attributes
+
+  my $cds_complete = undef;
+  my $codons_complete = undef;
+
+  my $translation = $peptide->translate( -terminator => undef, -unknown => undef, -frame => undef, 
+                         -codonTable_Id => $codon_table_id,
```
vsitnik, line 2175: Do we ignore case here?

modules/Bio/EnsEMBL/Transcript.pm
```diff
@@ -2161,9 +2161,21 @@ sub translate {
                                -alphabet => 'dna',
                                -id       => $display_id );
 
-  my $translation =
-    $peptide->translate( undef, undef, undef, $codon_table_id, undef,
-                         undef, $complete5, $complete3 );
+# Signature for BioPerl 1.6 should be (https://metacpan.org/pod/Bio::PrimarySeqI)
+#     my ($terminator, $unknown, $frame, $codonTableId, $complete,
```
vsitnik, line 2165: is it `codontable_id` ?
https://metacpan.org/pod/Bio::PrimarySeqI#translate

sgiorgetti: yes, it is

modules/Bio/EnsEMBL/Transcript.pm
vsitnik: found https://metacpan.org/dist/BioPerl/source/lib/Bio/PrimarySeqI.pm#L580

modules/Bio/EnsEMBL/Transcript.pm
```diff
@@ -2161,21 +2173,20 @@ sub translate {
...
+ 
   my $cds_complete = undef;
   my $codons_complete = undef;
-
-  my $translation = $peptide->translate( -terminator => undef, -unknown => undef, -frame => undef, 
-                         -codonTable_Id => $codon_table_id,
-                         -complete => $cds_complete, -complete_codons => $codons_complete,
-                         -throw => undef, -codonTable => undef,
-                         -orf => undef, -start => undef, -offset => undef );
+  $cds_complete = 0 if ( $mrna5_nf + $cds5_nf + $mrna3_nf + $cds3_nf > 0);
```
vsitnik, line 2182: Imho,  only `cds` related flags should be checked ($cds5_nf + $cds3_nf).

nwillhoft: Hi @vsitnik - out of curiosity, if the CDS start (or end) is not found, would it not be the case that the mRNA start (or end) is also not found? I was wondering if it could be a useful check to have both CDS and mRNA attribute information, and also if there might be a case where the mRNA_*_NF attribute could be flagged as incomplete but the CDS-based attribute might be missing.

vsitnik: @nwillhoft , as far as I understand it, the CDS is just a substring of mRNA. I suspect you can have partial mRNA but intact CDS. But for translation (peptide sequence inference) we need just CDS in this case.

nwillhoft: Thanks @vsitnik

TamaraNaboulsi:
2 new tests were added to transcript.t to test the translate method changes. One tests setting the cds_complete option, and the other tests the codon_complete option. New test data was added (new gene, transcript, translation, exons, etc...) and other test scripts were updated as well in terms of counting genes and such. Other fixes were also needed to handle having the CDS complete transcript attribute, as not having that leads to warnings when translate is called and those warnings lead to test failures.

In addition, this change also necessitated the fixing of a bareword directory handle due to a new policy in Perl::Critic added recently.

nwillhoft (approved):
In this PR, the `translate` method in `Transcript.pm` has been updated to ensure accurate translations.

Attributes are now used to test whether the CDS start and end are found or not, and the CDS complete value is set accordingly. These replace the previously defined `$complete5` and `$complete3` attributes. The parameters required for the BioPerl translate method have also been updated.

Two test transcripts (and their associated test data in corresponding files) have been added in to test various translation outputs and counts.

The `DUMPDIR` bareword handle has also been replaced with a variable.

All checks are passing, approving changes.

sgiorgetti:
Converted back to draft: further investigation is required to assess behaviour of the $complete flag, in some cases apparently related to microbial (only?) CDS

PR #636, opened 2022-10-25
Fixes to handle source file changes

TamaraNaboulsi:
## Description

UniProt and MGI files have recently changed and thus their parsers needed updating to handle those changes.

## Use case

The UniProt Swissprot now includes versioned stable IDs in the Ensembl mapping lines. This leads to problems with isoform xrefs and potentially other issues. Thus, the version is trimmed from the stable ID, maintaining the original code behavior.
The MGI MRK_List2.rpt file can contain special characters that makes the MGI_desc_parser fail. Adding the 'binary' option into the Text::CSV constructor fixes this.

## Benefits

No failures in MGI parser and correct uniprot isoform xrefs.

## Possible Drawbacks

Need to check impact on browser

## Testing

No

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -74,7 +74,7 @@ test:perl5.30-mysql:
 .dependent_template:
   stage: test
   only:
-    - master
+    - main
     - /^release/\d+$/
 
 # Actual trigger jobs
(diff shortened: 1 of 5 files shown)


sgiorgetti (approved):
Version is stripped off from stable ID, as expected.
Binary option is enabled in Text::CSV, which is OK.

Look good to me

PR #638, opened 2022-11-04
ENSCORESW-4084 - Ensure new Ensembl stable ids have correct prefix

nwillhoft:
## Requirements

## Description

Following the Genes on Patches / Y-PAR project work, a case has occurred where newly-generated Ensembl stable ids contain a ENS[G|T|P|E]1 prefix, instead of a 0. This PR ensures that the number portion of the stable id starts with a zero. 

## Use case

Related ticket at: https://www.ebi.ac.uk/panda/jira/browse/ENSCORESW-4084

Jose discovered in his data checks that the transcripts, exons and translations from genes in the Y PAR regions of the havana database (nwillhoft_yattrib_test) have non-standard stable ids, eg. ENST10010355966, ENSE10028847446, ENSP10000364511.

Background of these data from Jose: "There are a number of stable ids starting with 'ENST1' instead of 'ENST0' in the database, which were introduced to tag dubious transcript models when the so-called 'nomerge' database was created. The 'nomerge' database is the current havana database. Those stable ids would have been replaced by the stable id mapping script in the former version of the script, but this script was recently modified to honour new stable ids coming from the havana database for another reason. This is an unintended consequence of that change. I am guessing that you used these 'ENST100' stable ids as the starting point to create new transcript stable ids for the chrY PAR transcripts, and the same for exons and translation stable ids. Gene stable ids look correct though. Could you please ensure that the stable ids assigned to features on the chrY PAR use the standard 'ENST0'/'ENSP0'/'ENSE0' prefixes when the final modified havana database is prepared?"

## Benefits

Explicitly checks the stable id prefix.

## Testing

### DIFF
--- a/misc-scripts/generate_stable_ids.pl
+++ b/misc-scripts/generate_stable_ids.pl
@@ -145,7 +145,7 @@ sub get_max_stable_id_from_gene_archive {
   my ( $dbi, $type ) = @_;
 
   # Try to get from relevant archive.
-  my $sth = $dbi->prepare("SELECT MAX($type) FROM gene_archive WHERE $type LIKE 'ENS%'");
+  my $sth = $dbi->prepare("SELECT MAX($type) FROM gene_archive WHERE $type LIKE 'ENS_0%'");
   $sth->execute();
 
   my $rs;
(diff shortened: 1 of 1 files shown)


jmgonzmart (changes requested):
I think that the modified queries would only work with human stable ids. They would need to take the species prefixes into account. A posible solution would be the following:
```
SELECT MAX(stable_id) FROM gene WHERE stable_id REGEXP 'ENS[A-Z]{0,3}(G|T|P|E)0[0-9]{10}';
SELECT MAX(gene_stable_id) FROM gene_archive WHERE gene_stable_id REGEXP 'ENS[A-Z]{0,3}(G|T|P|E)0[0-9]{10}';
```
This assumes that species prefixes are always 0 (human) or 3 letters in length. It would be convenient to check this with all the teams that produce annotation and may use the script.

nwillhoft:
Closing this PR for now. There is no broken code here and further discussions are needed across teams to ensure the updates would be ok across all species.

PR #640, opened 2022-11-16
Travis fix

TamaraNaboulsi:
Fixing the broken travis run:
- reverting back to testing perl 5.26 on trusty while resting 5.30 on focal (errors when testing 5.26 with focal)
- adding back ensembl-git-tools

If we really don't want to use ensembl-git-tools, we can update the repo names to include the full link


sgiorgetti (approved):
Looks good to me

PR #641, opened 2022-12-08
Add info_text for Y chromosome xrefs

TamaraNaboulsi:
## Description

The core DB has dangling object_xref records related to Y-PAR genes after the xref pipeline is run on the new core DB (containing Y-PAR genes and and genes on patches).

## Use case

When running the xref pipeline using the core DB with the new Y-PAR genes, xrefs are duplicated from X-PAR ones and dumped into the file created by CoordinateMapper.pm in the dump_xref method. However, when this data is uploaded into the core DB, these duplicated xrefs cause a duplicate unique key error in the database `UNIQUE KEY id_index (dbprimary_acc, external_db_id, info_type, info_text, version)`. Also, since the object_xref data is uploaded before the xref is, all object xrefs telated to these duplicated xrefs would already be added. This leads to having dangling object xrefs int he core DB with no parent xref rows. To fix that, a string "Y Chromosome" has been added into the info_text field (which was previously kept empty) to differentiate the Y-PAR xrefs and allow inserting them into the DB.

## Benefits

No duplicate unique key error, leading to no dangling object_xref records.

## Possible Drawbacks

N/A

## Testing

Test suite is failing due to the Perl::Critic issue which was fixed in #635 . Once that is merged in, the tests can be re-run.

### DIFF
--- a/misc-scripts/xref_mapping/XrefMapper/CoordinateMapper.pm
+++ b/misc-scripts/xref_mapping/XrefMapper/CoordinateMapper.pm
@@ -539,6 +539,7 @@ sub run_coordinatemapping {
                 delete( $unmapped{$coord_xref_id} );
                 $mapped{$coord_xref_id}{'reason'}      = undef;
                 $mapped{$coord_xref_id}{'reason_full'} = undef;
+                $mapped{$coord_xref_id}{'chr_name'} = $chr_name;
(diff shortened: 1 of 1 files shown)


sgiorgetti (approved):
Looks good to me

PR #642, opened 2022-12-12
Change use of bareword dir handle for new ProhibitBarewordDirHandles…

TamaraNaboulsi:
## Description

Fixing the use of bareword directory handles due to a new policy in Perl::Critic added recently.

## Use case

A new policy was recently added into the Perl::Critic module that prohibits the use of bareword directory handles. This is causing the test housekeeping_perlCritic.t to fail.

## Benefits

No failure in housekeeping_perlCritic.t

## Possible Drawbacks

N/A

## Testing


EbiArnie (approved):
OK

PR #643, opened 2022-12-12
Fix/xref synonyms

TamaraNaboulsi:
## Description

Handling of Greek characters in HGNC xrefs synonyms.

## Use case

Some synonyms present in the HGNC xref file contain the Greek characters α, β, and γ. Without any special parsing, these synonyms are garbled in the database, leading to them appearing as garbled on the Ensembl browser. To fix this, the Text::Unidecode module is being used to translate these characters into their equivalent English letters.

This solution was chosen instead of translating the character to their corresponding words (α to 'alpha', β to 'beta', etc...) because this solution doesn't require having a hard-coded mapping hash in the code. 

## Benefits

No garbled characters in external synonyms.

## Possible Drawbacks

This change is restricted to the HGNC parser, as this is where the special characters have been noticed. However, this same change can be later added to the wider xref synonym handling for other sources.

## Testing

Added Text::Unidecode to the cpanfile used by the test suite

Test suite is still failing due to another issue fixed in #642 , will be re-run once that is merged in.

### DIFF
--- a/misc-scripts/xref_mapping/XrefParser/HGNCParser.pm
+++ b/misc-scripts/xref_mapping/XrefParser/HGNCParser.pm
@@ -71,6 +71,7 @@ use warnings;
 use Carp;
 use Text::CSV;
 use utf8;
+use Text::Unidecode;
 
 use parent qw( XrefParser::BaseParser );
 
@@ -456,6 +457,8 @@ sub add_synonyms_for_hgnc {
     $dead_string =~ s/"//xg;
     my @dead_array = split( ',\s', $dead_string );
     foreach my $dead (@dead_array){
+      utf8::decode($dead);
+      $dead = unidecode(uc($dead));
       $self->add_to_syn($name, $source_id, $dead, $species_id, $dbi);
     }
   }
@@ -465,6 +468,8 @@ sub add_synonyms_for_hgnc {
     $alias_string =~ s/"//xg;
     my @alias_array = split( ',\s', $alias_string );
     foreach my $alias (@alias_array){
+      utf8::decode($alias);
+      $alias = unidecode(uc($alias));
       $self->add_to_syn($name, $source_id, $alias, $species_id, $dbi);
     }
   }
--- a/misc-scripts/xref_mapping/XrefParser/HGNCParser.pm
+++ b/misc-scripts/xref_mapping/XrefParser/HGNCParser.pm
@@ -70,7 +70,8 @@ use strict;
 use warnings;
 use Carp;
 use Text::CSV;
-use utf8;
+#use utf8;
+use Encode;
 use Text::Unidecode;
 
 use parent qw( XrefParser::BaseParser );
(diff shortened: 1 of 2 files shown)


misc-scripts/xref_mapping/XrefParser/HGNCParser.pm
```diff
@@ -456,6 +457,8 @@ sub add_synonyms_for_hgnc {
     $dead_string =~ s/"//xg;
     my @dead_array = split( ',\s', $dead_string );
     foreach my $dead (@dead_array){
+      utf8::decode($dead);
```
EbiArnie, line 460: I recommend using Encode::decode. Encode is a core module. While utf8::decode works for the greek characters we have here, Encode::decode is more robust and can deal with invalid data. Since we don't have any control over the data we get, I think Encode::decode is safer.
I'd also drop the "use utf8;" - this is only useful if you use utf8 characters in the source code, which does not seem to be the case here.
Example usage:
```
use Encode;
my $str = decode("UTF-8", $bytes);
```

misc-scripts/xref_mapping/XrefParser/HGNCParser.pm
```diff
@@ -71,6 +71,7 @@ use warnings;
 use Carp;
 use Text::CSV;
 use utf8;
```
EbiArnie, line 73: I recommend to remove this, see comment below.

misc-scripts/xref_mapping/XrefParser/HGNCParser.pm
```diff
@@ -465,6 +468,8 @@ sub add_synonyms_for_hgnc {
     $alias_string =~ s/"//xg;
     my @alias_array = split( ',\s', $alias_string );
     foreach my $alias (@alias_array){
+      utf8::decode($alias);
```
EbiArnie, line 471: Same as above.

EbiArnie (changes requested):
I recommend using the Encode module, see other comments. Looks good otherwise.

misc-scripts/xref_mapping/XrefParser/HGNCParser.pm
TamaraNaboulsi: Hi Arne. Thanks for the review. While I agree with you, I only used utf8::decode because utf8::encode in used earlier in the script, so I kept it for consistency. I will however test around with you suggestions and update accordingly.

misc-scripts/xref_mapping/XrefParser/HGNCParser.pm
```diff
@@ -70,7 +70,9 @@ use strict;
 use warnings;
 use Carp;
 use Text::CSV;
-use utf8;
+#use utf8;
```
EbiArnie, line 73: Please remove the comments. If we need to remember, this is always there in the git history.

misc-scripts/xref_mapping/XrefParser/HGNCParser.pm
```diff
@@ -204,7 +206,8 @@ CCDS
   }) or croak "Cannot use file $file: ".Text::CSV->error_diag ();
 
   # make sure it's utf8
-  utf8::encode($mem_file);
+  #utf8::encode($mem_file);
```
EbiArnie, line 209: Same here.

EbiArnie (changes requested):
OK, apart from two nits.

EbiArnie (approved):
OK

PR #644, opened 2022-12-14
Add backup/nobackup parameter

thibauthourlier:
The script is writing a backup of the meta_coord table in a file. When the script is run and the current directory is not writeable it skips the database with a warning. However the warning might not be seen if the script is not run interactively.
By using --nobackup the script will not try to write the backup file

## Description

I have added a new option, `--nobackup` which do not write a backup of the meta_coord table to the current directory.
The backup is not needed as the the old values are probably wrong or missing some values.

## Use case

When running a pipeline which execute the update_meta_coords.pl script and the current directory is not writeable, the analysis will be successful however the meta_coord table will not have been updated which would cause problem later. Because the script is not run interactively, the fact that the database was skipped cannot be easily seen.

## Benefits

The script can be run from anywhere.

## Possible Drawbacks

None the default is still the write the backup file

## Testing

NA

NA

NA

### DIFF
--- a/misc-scripts/meta_coord/update_meta_coord.pl
+++ b/misc-scripts/meta_coord/update_meta_coord.pl
@@ -59,6 +59,8 @@ USAGE:
   --dbpattern   Specifies a regular expression for (possibly) matching
                 multiple databases.
 
+  --nobackup    Do not backup the meta_coord table in a file.
+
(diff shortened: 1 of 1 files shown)


sgiorgetti (approved):
Looks good

PR #646, opened 2022-12-16
Version bump to 110 with test db patching

TamaraNaboulsi:
## Description

Version bump to 110 and test DB patching

## Use case

Bumping the version to 110 before branching and patching the test DBs, including new patches added in #645 

## Benefits

Version update

## Possible Drawbacks

None

## Testing


nwillhoft (approved):
Changes look good for this PR. Tests pass and have summarised changes below:
- Version bump from 109 to 110
- Meta table data is patched accordingly for version bump
- Patches for schema changes are also added into meta text files
- The unique index for gene_id in alt_allele is consistently removed in SQL files
- Numerous Compara-based patches have been incorporated for Release 109 and 110

PR #650, opened 2023-01-10
Add info_text for Y chromosome xrefs

TamaraNaboulsi:
## Description

Same change as #641 but for branch 110


sgiorgetti (approved):
I'll miss the funny "FIXME (possibly)".

All's good to me otherwise.

PR #655, opened 2023-03-30
Maintaining manually inserted xrefs

TamaraNaboulsi:
## Description

Changes to allow harmonization of the vertebrates xref pipeline with plants species.

## Use case

In order to be able to run the current vertebrates xref pipeline for plants, there was a requirement to maintain manually inserted xrefs as the main display xrefs for plants species, as that is a case common to these species. The usual behavior of the xref pipeline already did that in the sense that it doesn't delete any existing xrefs that don't belong to sources that the pipeline itself retrieves. However, when reaching the point of setting display xref IDs for genes (which in turn sets the gene name), only xrefs that are retrieved in the current pipeline run would be considered. Thus, many plants genes lost their previous names (replaced with other names coming for other sources), and that wasn't optimal for these species. Thus, this change maintains the display xref IDs that belong to these manually curated xrefs and only sets new ones if no manual ones are already set.

## Benefits

Harmonization of the xref pipeline to run for both vertebrates and plants species.

## Possible Drawbacks

No pipeline option/flag has been added to make this change configurable, so it now applies to all species and all runs. However, since manually curated xrefs have a higher value than any other xrefs, they should always be treated as a priority.

## Testing

The full xref flow, download and process parts, was run for 3 different plants species as well as 2 different vertebrates. The results seem okay and Guy is happy with the plants results. The vertebrates species have not been affected by this change as they don't have manually curated xrefs.

### DIFF
--- a/misc-scripts/xref_mapping/XrefMapper/DisplayXrefs.pm
+++ b/misc-scripts/xref_mapping/XrefMapper/DisplayXrefs.pm
@@ -446,10 +446,6 @@ sub set_display_xrefs{
 
   print "Using xref_off set of $xref_offset\n" if($self->verbose);
 
-  #my $reset_sth = $core_dbi->prepare("UPDATE gene SET display_xref_id = null");
-  #$reset_sth->execute();
-  #$reset_sth->finish;
- 
   my $reset_sth = $core_dbi->prepare("UPDATE transcript SET display_xref_id = null WHERE biotype NOT IN ('LRG_gene')");
   $reset_sth->execute();
   $reset_sth->finish;
--- a/misc-scripts/xref_mapping/XrefMapper/DisplayXrefs.pm
+++ b/misc-scripts/xref_mapping/XrefMapper/DisplayXrefs.pm
@@ -239,7 +239,7 @@ sub genes_and_transcripts_attributes_set{
     $self->mapper->set_gene_descriptions();
   }
   else{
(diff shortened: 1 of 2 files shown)


misc-scripts/xref_mapping/XrefMapper/DisplayXrefs.pm
```diff
@@ -446,15 +446,15 @@ sub set_display_xrefs{
 
   print "Using xref_off set of $xref_offset\n" if($self->verbose);
 
-  my $reset_sth = $core_dbi->prepare("UPDATE gene SET display_xref_id = null");
-  $reset_sth->execute();
-  $reset_sth->finish;
+  #my $reset_sth = $core_dbi->prepare("UPDATE gene SET display_xref_id = null");
```
EbiArnie, line 449: Please remove these, no need to keep them around.

EbiArnie (changes requested):
Just a nitpick

misc-scripts/xref_mapping/XrefMapper/DisplayXrefs.pm
TamaraNaboulsi: You really dislike commented out code, don't you :D To be honest I do too, I just forget to delete before committing. Will remove! Thanks

EbiArnie (approved):
OK

PR #657, opened 2023-05-18
Add BRAKER stable_id prefix

leannehaggerty:
## Requirements

## Description

Updating the stable_id mapping code to allow mapping between BRAKER annotations

## Use case

The stable_id mapping pipeline was failing for BRAKER annotations as these have the prefix "BRAKER" rather than "ENS"

## Benefits

Mapping between BRAKER annotation will now work!

## Possible Drawbacks

I see none :) 

## Testing

Didn't add/change any tests, but don't suspect this small change will affect anything.
I have tested the pipeline with these changes and it was successful!


nwillhoft (approved):
PR to allow the stable ID mapping pipeline to map between sequence regions with BRAKER stable ids. Checks look ok as the test suite is passing.

PR #658, opened 2023-06-01
Update Registry.pm

dpopleton:
Updated _group_to_adaptor_class to use hash and include ontology

## Requirements

## Description

Modified the group_to_adaptor_class to include %group2adaptor.

## Use case

We couldn't load ontology dbs for an unknown reason since 107. We had to modify the _group_to_adaptor within our code. This should prevent that for happening to any of the existing adapters.

## Benefits

This should be more future proof, as it uses the group2adaptor hash rather than its own. It also fixes the problem of not loading ontology.

## Possible Drawbacks

Using the hash could break something, but I can't see what at the moment, but I am no expert on use cases for this pipeline. Should this be undesirable, an alternative is to just add : 'Bio::EnsEMBL::Compara::DBSQL::DBAdaptor',  ontology => 'Bio::EnsEMBL::DBSQL::DBAdaptor' to the hash.

## Testing

No

### DIFF
--- a/modules/Bio/EnsEMBL/Registry.pm
+++ b/modules/Bio/EnsEMBL/Registry.pm
@@ -2507,16 +2507,13 @@ sub load_registry_from_db {
 
 sub _group_to_adaptor_class {
   my ($self, $group) = @_;
-  my $class = {
-    core => 'Bio::EnsEMBL::DBSQL::DBAdaptor',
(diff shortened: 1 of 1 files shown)


sgiorgetti (approved):
Changes are clear
- replaced hardcoded adaptors in `_group_to_adaptor_class`
to use in `load_registry_from_url`

Test suite successful

Fine by me

PR #666, opened 2023-10-11
Restrain using latest DBD::mysql version

nakib103:
## Description

DBD::mysql have released new version (5.001) on Oct 4, 2023. This version removes support for MySQL 5 -
https://metacpan.org/dist/DBD-mysql/changes

Hence we need to use lower version than that.

## Use case

Using DBAdaptors to connect to Ensembl databases which use MySQL version 5

## Possible Drawbacks

Should be changed back if anytime in the future Ensembl database servers upgraded to MySQL 8x.

## Testing

Unit test should pass on this PR


sgiorgetti (approved):
Good by me

PR #677, opened 2024-01-15
Changed constraint to DBD::mysql to allow bugfixes

sgiorgetti:
## Description

We currently want to use MySQL 5.x, which is unsupported by DBD::mysql v5.
However, we do want to use the latest DBD::mysql v4, and we are already at 4.052 at the time of writing this.

## Use case

We want to use the latest DBD::mysql v4.

## Benefits

Allowing bug fixing patches to DBD::mysql

## Possible Drawbacks

None

## Testing

The test suite worked just fine.


EbiArnie (approved):
OK

PR #679, opened 2024-02-06
Xref fixes

TamaraNaboulsi:
## Description

Fixes for some xref bugs that were discovered in the 112 run.

## Use case

There was a bug in parsing UniProt gene names that was discovered in the recent 112 xref run which didn't include the colon ":" character in the regex used.
Another bug that was found was related to populating the xref intermediate DBs where data was being copied over from the central DB. The way the data was copied was by creating a .sql file with the insert statements. However, when running the pipeline for multiple species at once (normal process), they were all rewriting the same sql file over and over again which lead to some of the intermediate DBs having no data. The added fix removed this copying of the central DB data altogether, as there already exists a file that has this data (populate_metadata.sql) which is created when the central DB is created. Instead, the preparse flag value was added into this metadata filename in order to handle multiple runs of the pipeline (maybe with different preparse flags).

## Benefits

No rewriting of the same file for no reason, no corrupted intermediate DBs, and no wrong UniProt gene names.

## Possible Drawbacks

None

## Testing

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -5,6 +5,7 @@ requires 'IO::Compress::Gzip';
 requires 'URI::Escape';
 requires 'Config::IniFiles';
 requires 'Gzip::Faster';
+requires 'List::MoreUtils';
 
 test_requires 'Test::Warnings';
 test_requires 'Test::Differences';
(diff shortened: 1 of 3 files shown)


sgiorgetti (approved):
OK

PR #663, opened 2023-08-22
Version update and schema patches for Release 112

nwillhoft:
## Requirements

- Filling out the template is required. Any pull request that does not include enough information to be reviewed in a timely manner may be closed at the maintainers' discretion;
- Review the [contributing guidelines](https://github.com/Ensembl/ensembl/blob/master/CONTRIBUTING.md#why-could-my-pull-request-be-rejected) for this repository; remember in particular:
    - do not modify code without testing for regression
    - provide simple unit tests to test the changes
    - if you change the schema you must patch the test databases as well, see [Updating the schema](https://github.com/Ensembl/ensembl/blob/master/CONTRIBUTING.md#updating-the-schema)
    - the PR must not fail unit testing

## Description

The release version is incremented to 112 in the API code. Version and database schemas for test genomes databases are also bumped to 112.

An extra patch is added (by request from the Production team) to allow the `meta_value` to be NULL.

44 files are changed in this PR, matching the file change count for Release 111 (#656).

## Use case

Before branching the code for Release 112, the API version and the database schemas need patching to 112.

## Benefits

Version updates. Datachecks for Production team relating to the `meta_value` should now pass.

## Possible Drawbacks

N/A

## Testing

_Have you added/modified unit tests to test the changes?_

_If so, do the tests pass/fail?_

Tests are failing as there is a version mismatch of 111 and 112 between the software and database versions. I believe that this will be resolved once this PR is applied.

Tests relating to meta database schema checks are failing. They appear to be checking the patched 112 schema against an expected old 111 schema, which of course is failing. I believe it should point to the correct 112 schema once these patches are merged in.

_Have you run the entire test suite and no regression was detected?_

PR #680, opened 2024-02-07
Updated core meta table definition to allow NULL meta_value and exten…

marcoooo:
…d the meta_key VARCHAR length

## Description

For concurrent MVP/RR handover, we need to extend the length of meta_key for some stats to be inserted the right way. 
As of now the compute of stats try to insert <NULL> values in meta table as well, even if the web requirement is obsolete now. we still need to allow this for the this particular release branch and script version.   

## Use case

Concurrent handover allowing NULL values and inserting non truncated meta_key names used for MVP.

## Benefits

Enable concurrent Handover without DC failures.

## Possible Drawbacks

None. 

## Testing

### DIFF
--- a/sql/table.sql
+++ b/sql/table.sql
@@ -322,6 +322,9 @@ INSERT INTO meta (species_id, meta_key, meta_value)
 INSERT INTO meta (species_id, meta_key, meta_value)
   VALUES (NULL, 'patch', 'patch_109_110_c.sql|Allow gene id to belong to multiple alt allele groups');
 
+INSERT INTO meta (species_id, meta_key, meta_value)
+  VALUES (NULL, 'patch', 'patch_109_110_d.sql|Extend meta_key length to 64 - allow NULL in meta_value');
+
 /**
 @table meta_coord
 @colour #C70C09
--- a/sql/table.sql
+++ b/sql/table.sql
@@ -296,7 +296,7 @@ CREATE TABLE IF NOT EXISTS meta (
   meta_id                     INT NOT NULL AUTO_INCREMENT,
   species_id                  INT UNSIGNED DEFAULT 1,
   meta_key                    VARCHAR(64) NOT NULL,
-  meta_value                  VARCHAR(255) NULL,
+  meta_value                  VARCHAR(255) DEFAULT NULL,
 
   PRIMARY   KEY (meta_id),
   UNIQUE    KEY species_key_value_idx (species_id, meta_key, meta_value),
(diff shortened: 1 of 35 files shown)


JAlvarezJarreta (approved):

sql/table.sql
```diff
@@ -295,8 +295,8 @@ CREATE TABLE IF NOT EXISTS meta (
 
   meta_id                     INT NOT NULL AUTO_INCREMENT,
   species_id                  INT UNSIGNED DEFAULT 1,
-  meta_key                    VARCHAR(40) NOT NULL,
-  meta_value                  VARCHAR(255) NOT NULL,
+  meta_key                    VARCHAR(64) NOT NULL,
+  meta_value                  VARCHAR(255) NULL,
```
sgiorgetti, line 299: Would you mind to change this for consistency to 
`meta_value                  VARCHAR(255) DEFAULT NULL,`

Thx

sgiorgetti (changes requested):
minor change required for consistency. Thx

sql/table.sql
marcoooo: Done my dear sir.

sgiorgetti (approved):
Fine by me. Thx Marc!

dpopleton (approved):

leannehaggerty (approved):
Thank you Marc!

nwillhoft (changes requested):
Relating to the patching of the test genome databases, I believe there are a couple of variation-related dbs that require patching here too (please see release/112 patching file for example: https://github.com/Ensembl/ensembl/pull/663/files). I think this would increase the file count by another 4 files.

marcoooo:
> Relating to the patching of the test genome databases, I believe there are a couple of variation-related dbs that require patching here too (please see release/112 patching file for example: https://github.com/Ensembl/ensembl/pull/663/files). I think this would increase the file count by another 4 files.

Hi @nwillhoft I am not sure what you mean? Do you mean I should patch the test files `table.sql` everywhere?

nwillhoft:
> > Relating to the patching of the test genome databases, I believe there are a couple of variation-related dbs that require patching here too (please see release/112 patching file for example: https://github.com/Ensembl/ensembl/pull/663/files). I think this would increase the file count by another 4 files.
> 
> Hi @nwillhoft I am not sure what you mean? Do you mean I should patch the test files `table.sql` everywhere?

@marcoooo Hi, When the test genome databases are patched on our side, there are a certain number of files that we look for and I can see here that some of them are missing. Just looking into why they weren't picked up this time.

sgiorgetti (approved):
ok

nwillhoft (approved):
After discussions with Stefano and going through the files and test suite, seems like these changes are all good.

PR #685, opened 2024-03-04
Missing Method for Module

TamaraNaboulsi:
## Requirements

## Description

Fix for calling a missing method for object.

## Use case

The method add_linkage_type is called on a variable when the linkage_type value exists. The variable can either be an IdentityXref or an OntologyXref object. However, only OntologyXref has this method defined as the linkage_type only makes sense for an ontology xref. A fix was added to only call this method if the correct object is being used.

## Benefits

Bug fix.

## Possible Drawbacks

None.
However, additional checks will be done to figure out if the occurrence of this scenario makes sense in the domain of xrefs (having a linkage_type value when the xref is an IdentityXref one. This fix is needed either way, though.

## Testing


EbiArnie (approved):
OK

PR #687, opened 2024-03-12
Fixed double init

sgiorgetti:
## Description

Fixing warning because of double "init".

## Benefits

Get rid of unwanted warnings.

## Possible Drawbacks

None

## Testing

Test suite ran fine.


TamaraNaboulsi (approved):
sth variable initialized twice. Looks good.

PR #688, opened 2024-03-12
Fixed double init

sgiorgetti:
## Description

Sister PR to #687 - this one against `main` branch
Fixing warning because of double "init".

## Benefits

Get rid of unwanted warnings.

## Possible Drawbacks

None

## Testing

Test suite ran fine.


TamaraNaboulsi (approved):
Same as #687 . Looks good.

PR #690, opened 2024-03-21
Fix incorrect escape for percent char in SQL statement

sgiorgetti:
## Description

## Use case

Bug fixing.

## Benefits

## Possible Drawbacks

None

## Testing

The test suite ran fine.


sgiorgetti:
Wrong target branch :-(

PR #691, opened 2024-03-21
Fixed incorrect escape for percent in SQL stmt

sgiorgetti:
## Description

## Use case

Bug fixing.

## Benefits

## Possible Drawbacks

None

## Testing

The test suite ran fine.


vsitnik (approved):
cool. thank you!

vsitnik (approved):
cool. thanks!

EbiArnie (approved):
This is looking good.

One comment though.
INSTR is understood by MySQL and SQLite, which is what we care about here.
The SQL standard would suggest POSITION though.
But since SQLite doesn't seem to have POSITION, this should go in as is.

PR #692, opened 2024-03-25
Fixed incorrect escape for percent in SQL stmt

sgiorgetti:
## Description

Sister PR to #691 

## Use case

Bug fixing.

## Benefits

## Possible Drawbacks

None

## Testing

The test suite ran fine.


EbiArnie (approved):
OK

PR #693, opened 2024-03-27
Fix/transl

sgiorgetti:
## Description

Updating transcript->translate() method to properly use alternate genetic codes (codon tables), and fix an issue with the starting codon, which sometimes was not getting translated as methionine.
The ultimate aim of this is reducing the need/burden of SeqEdits.

## Use case

Under some circumstances the translation process fails w.r.t. the start codon not becoming 'M', especially when alternate codon tables are required.

## Benefits

Reduce the need and maintenance burden of manual post-translation corrections (SeqEdit's).

## Possible Drawbacks

Issues with translations, if not carried out properly.

## Testing

Tests were not deemed to be adequate and added some - a minimum set TBH - to mitigate this.


modules/Bio/EnsEMBL/Transcript.pm
```diff
@@ -2091,79 +2091,82 @@ sub get_all_translateable_Exons {
...
+      chop $mrna for 1 .. $delta;
     }
   }
-  $codon_table_id ||= 1;    # default vertebrate codon table
+  
+  if ( CORE::length($mrna) < 1 ) { return undef }
+#  if ( CORE::length($mrna) == 3 && $codon_table->is_ter_codon( $last_mrna_codon ) ) { return undef }
+
+  my $display_id = $self->translation->display_id()
+    || scalar( $self->translation() );
```
vsitnik, line 2141: could you give a hint what  the `scalar( $self->translation() )` is? Thanks

modules/Bio/EnsEMBL/Transcript.pm
```diff
@@ -2091,79 +2091,82 @@ sub get_all_translateable_Exons {
...
+    # versions
+    if ( $complete_codon ) {
+      # If we want to do the bad behavior of bioperl 1.6.1 and older
+      # where we guess the last codon if incomplete, pad an N
+      # to the mrna sequence
+      $mrna .= 'N' x (3 - $delta);
+    } else {
+      # Otherwise trim those last bp's off so the behavior is
+      # consistent across bioperl versions
+      chop $mrna for 1 .. $delta;
```
vsitnik, line 2133: we sort of invalidating first and last codons here.
And later they are used for chopping last AA from peptide without checking if it's a "*" one or not.

modules/Bio/EnsEMBL/Transcript.pm
```diff
@@ -2091,79 +2091,82 @@ sub get_all_translateable_Exons {
...
-	  # If we want to do the bad behavior of bioperl 1.6.1 and older
-	  # where we guess the last codon if inomplete, pad an N
-	  # to the mrna sequence
-	  $mrna .= 'N';
-      } else {
-	  # Otherwise trim those last two bp off so the behavior is
-	  # consistent across bioperl versions
-	  substr( $mrna, -2, 2, '' );
-      }
+  if ( $codon_table->is_ter_codon( $last_mrna_codon ) ) {
```
vsitnik, line 2158: `$lasct_mrna_codon` can be outdated because of 2120-2133.
Shoudl we add check for the "*" at the end of the $p_seq?

modules/Bio/EnsEMBL/Transcript.pm
```diff
@@ -2091,79 +2091,82 @@ sub get_all_translateable_Exons {
...
-
-  my $display_id = $self->translation->display_id()
-    || scalar( $self->translation() );
-
-  my $peptide = Bio::Seq->new( -seq      => $mrna,
-                               -moltype  => 'dna',
-                               -alphabet => 'dna',
-                               -id       => $display_id );
+  # Update the first amino acid to M, as appropriate
+  if ( substr($p_seq,0,1) ne 'M' && $codon_table->is_start_codon( $first_mrna_codon ) ) {
```
vsitnik, line 2163: Same, `$first_mrna_codon` can be outdated because of the mrna size alteration ( = 0 mod 3) on 2120-2133

vsitnik (commented):
Could you check this bit with outdating first / last codon values once mrna is being modified to fulfill = 0 mod 3? Thanks.

modules/Bio/EnsEMBL/Transcript.pm
vsitnik: cool. should be ok

vsitnik (approved):
cool. thank you!

modules/Bio/EnsEMBL/Transcript.pm
```diff
@@ -2091,79 +2091,82 @@ sub get_all_translateable_Exons {
...
+      $mrna .= 'N' x (3 - $delta);
+    } else {
+      # Otherwise trim those last bp's off so the behavior is
+      # consistent across bioperl versions
+      chop $mrna for 1 .. $delta;
     }
   }
-  $codon_table_id ||= 1;    # default vertebrate codon table
+  
+  if ( CORE::length($mrna) < 1 ) { return undef }
```
EbiArnie, line 2135: We need at least length 3 for the substr. While Perl will not complain, would this make sense?
If not, this should be "< 3" or even "< 6". Having start and stop codon be the same thing also does not make too much sense.

EbiArnie (changes requested):
One nitpick, the rest looks good to me.

modules/Bio/EnsEMBL/Transcript.pm
vsitnik: What is returned on L2512
```
my $p_seq = $codon_table->translate($mrna, 0);
```
for an empty $mrna?
If it's an object, this is not equivalent to undef.

vsitnik: Sorry. Upon thinking. We can have a partial codon that we can translate :)
And sometimes we don't have starts and stops. I assume the question if the model having just one AA derived from a partial codon is a valid one should be answered somewhere else. In the calling code, i.e. This bit is just for translating, imho. Up to you, though.

EbiArnie: OK, if this could make sense in some situations, leave it as-is.

EbiArnie (approved):
OK

modules/Bio/EnsEMBL/Transcript.pm
vsitnik: It makes no sense. But can work. Sorry.

PR #694, opened 2024-03-27
Added support for the new 'gencode_primary' transcript attribute

sgiorgetti:
## Description

A new transcript attribute "gencode_primary" has being introduced (for human).
This attribute is meant to eventually - i.e. in the long run - replace GENCODE basic.

This change is necessary to support the displaying of the new attribute on the web site, as well as having it available in the published annotation files.

## Use case

Check a transcript for the "gencode_primary" attribute.
This is entirely similar to the already available check for the GENCODE Basic attribute.

## Benefits

It's a must have to properly handle the new attribute.

## Possible Drawbacks

None

## Testing

TBC

### DIFF
--- a/modules/Bio/EnsEMBL/Transcript.pm
+++ b/modules/Bio/EnsEMBL/Transcript.pm
@@ -3154,7 +3154,7 @@ sub summary_as_hash {
 
 =head2 gencode_basic
 
-  Example       : $gencode_basic = $transcript->tag();
+  Example       : $gencode_basic = $transcript->get_all_Attributes('gencode_basic')->[0];
   Description   : Returns true if gencode_basic is set
   Returns       : boolean
 =cut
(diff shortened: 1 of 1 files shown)


nwillhoft (changes requested):
The example and subroutine names in the newly added code need updating from gencode_basic to gencode_primary

modules/Bio/EnsEMBL/Transcript.pm
```diff
@@ -3142,7 +3142,18 @@ sub summary_as_hash {
...
-  $summary_ref->{'tag'} = 'basic' if $self->gencode_basic();
+
+  my @tags;
+  push @tags, 'basic' if $self->gencode_basic();
+  push @tags, 'Ensembl_canonical' if $self->is_canonical();
+  push @tags, 'primary' if $self->gencode_primary();
+
+  my $mane = $self->mane_transcript();
+  if ($mane) {
+    push @tags, $mane->type() if ($mane->type());
```
nwillhoft, line 3153: @sgiorgetti Do we need to define `type` for this `mane_transcript` object? I can see in the REST server that `type` is available and printed out but I wasn't sure where it's being defined, and if it needs to be defined here?

sgiorgetti: @nwillhoft thanks for bringing this up. Reporting here the result of our discussion for the records.
The `mane_transcript` Transcript method returns a `MANE` object, which has the `type` method defined in it. This latter method returns the "type" of MANE - i.e. "MANE_Select" or "MANE_Plus_Clinical".

nwillhoft (approved):
Changes to `modules/Bio/EnsEMBL/Transcript.pm` look good. A `gencode_primary` subroutine has been added in to identify whether a transcript has a GENCODE Primary attribute. A typo in the `gencode_basic` subroutine has been fixed. Tags have also been added to `summary_as_hash` to provide transcript-related info for: MANE, canonical, GENCODE Basic and GENCODE Primary properties.

PR #696, opened 2024-04-04
Update tags to fix GTF/GFF3 tag data

nwillhoft:
Found that GTF and GFF3 file tests are failing in `ensembl-io` due to tags having different values to those expected. For example, have found that modifying tags in this repo now allows `ensembl-io` to dump out "GENCODE Primary" instead of "primary" in GTF and GFF3 file tests.

Related PR: https://github.com/Ensembl/ensembl-io/pull/160.


sgiorgetti (approved):
ok

PR #698, opened 2024-04-16
Add c patch to 111_112 test databases

nwillhoft:
Before bumping version to 113, we need to add the c patch (extend meta_key value) to the 112 test databases.

This also adds the missing 111_112 b and c patch data to variation and ontology-based test db data.


sgiorgetti (approved):
OK

PR #700, opened 2024-04-17
Removed GitLab dependent builds

sgiorgetti:
## Description

GitLab pipeline triggers dependent builds on Travis for the `ensembl`-dependent repos (Variation, Regulation, Compara, REST). This does not work anymore because of current limitation of Travis not running concurrent builds for us.
This behaviour was specifically implemented for GitLab pipeline,  it's inconsistent with the current approach, and it's removed ... at least for the time being.

## Use case

PRs merged into `main` trigger a GitLab CI/CD pipeline, which in turn triggers builds for other repos on Travis.
This GitLab CI/CD pipeline step fails because Travis can't start the dependent builds fast enough.

## Benefits

Avoiding CI/CD failures

## Possible Drawbacks

Less prompt detection of failures in dependent code/repos.
These will pop up, when builds are triggered there, independently.

## Testing

N/A

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -70,45 +70,7 @@ test:perl5.38-sqlite:
 # Triggers for dependent builds
 #
 
-# The template. It doesn't presently support PRs before they are
-# merged (would need extended run condition and better selection of
-# downstream branches) - but then again, we do not trigger dependent
-# builds for PRs on Travis either.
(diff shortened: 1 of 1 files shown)


nwillhoft (approved):
Dependent build jobs are timing out due to reduced travis build runs. Changes in this PR to remove such jobs look good. Build checks are passing.

PR #701, opened 2024-04-22
Update tag from 'basic' to 'gencode_basic'

nwillhoft:
If a transcript has a GENCODE Basic annotation, currently it is described in files with a `basic` tag annotation. This PR updates that tag to be more descriptive with the tag `gencode_basic`. This also matches recent work to add the `gencode_primary` tag to files, where applicable.

Tests have been added to test both true and false instances of this tag being present.


sgiorgetti (approved):
looks good

PR #702, opened 2024-04-23
[113] Update Release branch (gitlab & gencode_basic)

nwillhoft:
The `gencode_basic` tag updates need bringing into the `release/113` branch (already pushed to main). The main branch also has recent gitlab-related changes that are missing from `release/113`, so suggesting to bring both changes in here.


sgiorgetti (approved):
OK

PR #706, opened 2024-05-30
Fix the id we use for Bio::Seq

EbiArnie:
- We use an id to construct a Bio::Seq object. Before, we would pass in an actual object reference instead of a string. This could result in a segfault later. This occurs because BioPerl will clone the object. The Transcript object contains DBI handles, these will frequently result in segfaults when cloned. Force a string instead.


sgiorgetti (changes requested):
Tests are failing because of two reasons
- missing copyright update
-  test suite considering Ensembl schema version-1 as benchmark for the changes, instead of going through the patches.

The second one can be forgiven, as we know these changes are irrelevant fir that.
The first failure must be resolved.

PR #708, opened 2024-06-14
Xref ZFIN Parser Fix

TamaraNaboulsi:
## Requirements

## Description

Changing ZFIN gene name mapping from only considering SWISSPROT to all Uniprot.

## Use case

A decrease has been noticed in the Zebrafish core DB for gene names assigned by ZFIN. It turns out that the ZFIN xref parser only adds ZFIN xrefs when they are associated with SWISSPROT data. A change was added to consider all Unirpot data (including trEMBL) in an effort to get back some of those lost gene names.

## Benefits

More ZFIN gene names.

## Possible Drawbacks

The trEMBL data can be considered less reliable than the SWISSPROT one, but because we are getting the mapping from ZFIN then we depend on them to verify that the mappings are accurate.

## Testing

The xref processing pipeline was run with this change and the number of ZFIN gene names increased by close to 4500.

### DIFF
--- a/misc-scripts/xref_mapping/XrefParser/ZFINParser.pm
+++ b/misc-scripts/xref_mapping/XrefParser/ZFINParser.pm
@@ -45,7 +45,7 @@ sub run {
   my $file = @{$files}[0];
   my $dir = dirname($file);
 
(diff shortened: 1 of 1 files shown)


sgiorgetti (approved):
OK

PR #710, opened 2024-06-21
Handle carriage returns

TamaraNaboulsi:
## Description

Fix in Xenbase parser.

## Use case

After changing the file downloaded for the Xenbase source, the parser is now failing due to a difference in the file. There exist some lines that end in a carriage return rather than a new line. This change fixes that by reading the lines verbatim and then replacing the carriage return.

## Benefits

Xenbase parser succeeds.

## Possible Drawbacks

None

## Testing

Parser was run with this change and the xrefs were correctly added into the intermediate DB.

### DIFF
--- a/misc-scripts/xref_mapping/XrefParser/XenopusJamboreeParser.pm
+++ b/misc-scripts/xref_mapping/XrefParser/XenopusJamboreeParser.pm
@@ -86,10 +86,13 @@ sub run {
   my $input_file = Text::CSV->new({
     sep_char       => "\t",
     empty_is_undef => 1,
+    binary => 1,
+    verbatim => 1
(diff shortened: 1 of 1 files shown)


EbiArnie (approved):
I have the feeling that using Text::CSV is making things harder instead of easier here.
Not sure how exactly setting verbatim works so that we still split at line endings, but it seems to work, so:
OK.

TamaraNaboulsi:
> I have the feeling that using Text::CSV is making things harder instead of easier here. Not sure how exactly setting verbatim works so that we still split at line endings, but it seems to work, so: OK.

I agree. It seems that Text::CSV is being used in a lot of parsers for the sole reason of consistency I guess, or to skip the line splitting step. In any case, since all parsers will be ported to python soon, I saw no need to overhaul now.

PR #712, opened 2024-07-01
Xref Changes for main

TamaraNaboulsi:
## Description

Same change as #710  and #708  applied for the main branch.


sgiorgetti (approved):
OK

PR #714, opened 2024-08-21
Xref Changes and Fixes

TamaraNaboulsi:
## Description

During the 113 xref run, several issues were flagged and these changes attempt to fix them.

## Use case

These changes consist of some bug fixes along with improvements to xrefs. These include:
- Making the ZFIN xrefs DIRECT ones (read from the ensembl_1_to_1 file downloaded from ZFIN), in addition to the DEPENDENT xrefs added based on UniProt and RefSeq
- Adding WikiGene xrefs along with EntrezGene ones when a mapping is found in the otherfeatures db
- Uniprot_gn xrefs were missing identity_xref rows when dependency is moved from the SEQUENCE_MATCH Uniprot to the DIRECT one, which was leading to loss in gene names
- Adding HGNC descriptions to Uniprot_gn xrefs, as they are used as display xrefs but have no descriptions since they are dependents
- Adding the STRING source to all species as it was missing in some Plants DBs (as a Uniprot DEPENDENT source)
- Adding the parsing of RefSeq_peptide to Plants species
- Fixing the handling of RefSeq descriptions as it was leaving in the "{ECO}" text, leading to a critical DC failure
- Fixing the handling of multiline gene names in the Uniprot file, which was leading to having a newline in the name

## Benefits

Bug fixes and improvements.

## Possible Drawbacks

The majority of gene names in most species will now come from Uniprot_gn instead of an HGNC projection. While this shouldn't be an issue, as the actual names are the same and descriptions still come from HGNC, it's still a noticeable change.

## Testing

Waiting for Codon to be back to complete testing.

### DIFF
--- a/misc-scripts/xref_mapping/XrefMapper/ProcessPrioritys.pm
+++ b/misc-scripts/xref_mapping/XrefMapper/ProcessPrioritys.pm
@@ -318,7 +318,12 @@ sub process_dependents{
       foreach my $ensembl_id (@{ $ensembl_ids{$object_type} }) {
         # Add new object_xref for each best_ensembl_id. 
         $insert_dep_ox_sth->execute($new_master_xref_id, $object_type, $ensembl_id, $dep_xref_id);
-        ## If there is a linkage_annotation, it is a go xref
+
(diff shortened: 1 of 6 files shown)


sgiorgetti (approved):
Looks good

PR #723, opened 2024-09-11
Copyright optimisation

sgiorgetti:
## Description

Centralising the copyright string with the year into the NOTICE file, and fixing the existing LICENSE file, to avoid massive update of the year across the repo.

## Use case

Every year the copyright must be updated. Because the copyright info - for historical reason - is contained in many files, the procedure is risky and annoying.
By centralising the year bit into a single NOTICE file, the entire updating procedure becomes trivial.

## Benefits

Centralising the copyright string with the year into the NOTICE file, and fixing the existing LICENSE file, to avoid massive update of the year across the repo.
The copyright update will be easier to do and check.

## Possible Drawbacks

None

## Testing

Tests modified to check for the new copyright header.
Test suite runs fine.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ before_install:
 
 install:
   - cpanm -v --sudo --notest XML::LibXML # https://github.com/shlomif/perl-XML-LibXML/pull/87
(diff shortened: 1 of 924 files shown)


EbiArnie (approved):
OK

PR #730, opened 2025-01-27
[115] Version update and schema patches

nwillhoft:
Update the API version to 115 and add 115 db schema patches.


sgiorgetti:
@nwillhoft  - changes look good by skimming through them ... is there a specific reason why the PR is in draft state? Anything I missed?

nwillhoft:
> @nwillhoft - changes look good by skimming through them ... is there a specific reason why the PR is in draft state? Anything I missed?

Thanks for checking @sgiorgetti. Had left in draft state as I wanted to double-check all the files were there, and it looks like it's good to go.

sgiorgetti (approved):
LGTM - Thx

PR #733, opened 2025-02-12
TranscriptSelector optimisation for genes with only one transcript

sgiorgetti:
## Description

TranscriptSelector optimisation for genes with only one transcript.

## Use case

Speed up of the `misc-scripts/canonical_transcripts/select_canonical_transcripts.pl`

## Benefits

Speed up  of the `misc-scripts/canonical_transcripts/select_canonical_transcripts.pl`

## Possible Drawbacks

None

## Testing

No specific tests were added.

@vsitnik tested by running `misc-scripts/canonical_transcripts/select_canonical_transcripts.pl`

Test suite could successfully ran locally.


EbiArnie (approved):
OK

PR #742, opened 2025-07-28
Gene object biotype assessment 

JackCurragh:
fixed equals subroutine to correctly evaluate biotype of self and target gene rather than self vs self

## Requirements

## Description

This PR fixes the equals subroutine in the Gene module. Previously the biotype check evaluated the biotype of $self vs $self meaning this check always passed even if the biotype of $self and the target gene ($gene) differed. This PR alters one line of the module and adds tests to confirm the correct behaviour.

## Use case

## Benefits

## Possible Drawbacks

## Testing

Yes
Pass
I have run just `gene.t`

### DIFF
--- a/modules/Bio/EnsEMBL/Gene.pm
+++ b/modules/Bio/EnsEMBL/Gene.pm
@@ -371,7 +371,7 @@ sub equals {
     return 0;
   }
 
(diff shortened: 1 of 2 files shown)


nwillhoft (approved):
Changes look good to correctly compare two genes in the `equals` subroutine. New tests are also passing. Thanks for this fix and adding these tests in.

PR #747, opened 2026-01-07
Use InnoDB engine in schema SQL

leannehaggerty:
Please do not review/merge until discussed in release meeting!

## Description

Beta processing pipelines now require databases to use the InnoDB storage engine for performance and locking behaviour.
Genebuild pipelines currently create core databases using sql/table.sql from the Ensembl repository, which explicitly specifies ENGINE=MyISAM.

This change updates all schema SQL files to replace MyISAM with InnoDB, ensuring newly created core databases meet the engine requirements for beta processing and upcoming handover validation checks.

## Use case

Recent beta processing has identified that databases created with the MyISAM engine are significantly slower to process, as pipelines are optimised for InnoDB (row-level locking, transaction support, improved concurrency).

Databases submitted with MyISAM tables now block handover and will fail the newly introduced engine validation datacheck. Updating the schema SQL ensures that genebuild pipelines create compliant databases by default, avoiding manual post-creation engine conversion before handover.

## Benefits

Ensures all newly created core databases use InnoDB by default
Aligns genebuild pipeline output with current beta processing and handover requirements
Prevents performance degradation caused by MyISAM table-level locking
Removes the need for manual ALTER TABLE ... ENGINE=InnoDB steps before submission

## Possible Drawbacks

Changing the default storage engine may alter performance characteristics compared to legacy MyISAM-based databases
Existing workflows that explicitly rely on MyISAM behaviour (if any) may need review
Conversion applies only to newly created databases; existing databases are unaffected and may still require manual conversion

## Testing
No unit tests were added or modified (schema-only change)
Verified that schema SQL no longer contains ENGINE=MyISAM

NOTE: this needs to be discussed/ approved in beta release meeting, then tested

### DIFF
--- a/sql/patch_39_40_g.sql
+++ b/sql/patch_39_40_g.sql
@@ -30,7 +30,7 @@ CREATE TABLE ditag (
 
        PRIMARY KEY ( ditag_id )
 
-) ENGINE=MyISAM DEFAULT CHARSET=latin1;
+) ENGINE=InnoDB DEFAULT CHARSET=latin1;
 
 CREATE TABLE ditag_feature (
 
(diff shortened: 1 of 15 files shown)


leannehaggerty:
Discussed in RelCo meeting: move this to an internally used branch so that genebuild pipelines can use this to create dbs in InnoDB but it doesn't affect other users. 
PR will be closed and new branch, including tests in genebuild pipelines, communicated in next RelCo meeting.

PR #751, opened 2026-05-29
DBEntryAdaptor._type_by_external_id changes in wildcard default logic

sgiorgetti:
## Description

This is for internal use only - at least I conceived it as such - for reducing the LIKE comparisons in queries triggered by `[_%]` in accessions.

The current behaviour is not to consider `[_%]` as wildcards by default, but standard characters.
Consequently the comparison operator is "=".
This behaviour can be overridden: in this case wildcards can be used, but their presence is allowed after the 3rd character in the accession string to prevent too broad queries. 

This change alters the contract and the impact is reflected in Gene, Transcript, and Translation adaptors.

Also when querying by external_db names, the filtering clause is built like `db_name LIKE "$external_name%"`.
This is making the query execution significantly slower.

The tests in `gene.t` required tuning, accordingly.

## Use case

DB overload due to abuse or misuse of wildcards in accessions or external db names.

## Benefits

The DB will breathe again.

## Possible Drawbacks

Some regression in edge cases/pages.

## Testing

`gene.t` logic needed updating because of the changes in the default logic, and wildcard behaviour.
Also added tests for ensuring results are unchanged when the db_name prefix matches one or multiple external_db rows.
Tests pass and no regression has been detected so far, but better testing would be desirable.

### DIFF
--- a/modules/Bio/EnsEMBL/DBSQL/DBEntryAdaptor.pm
+++ b/modules/Bio/EnsEMBL/DBSQL/DBEntryAdaptor.pm
@@ -2019,15 +2019,21 @@ SQL
     }
   }
 
+  my @external_db_ids;
   if ( defined($external_db_name) ) {
-    # Involve the 'external_db' table to limit the hits to a particular
-    # external database.
+    # Preserve the prefix match on db_name, but keep external_db out of the lookup
+    # so the xref indices remain usable
 
-    $from_sql .= 'external_db xdb, ';
-    $where_sql .=
-        'xdb.db_name LIKE '
-      . $self->dbc()->db_handle()->quote( $external_db_name . '%' )
-      . ' AND xdb.external_db_id = x.external_db_id AND';
+    @external_db_ids = @{
+        $self->dbc()->sql_helper()->execute_simple(
+            -SQL    => 'SELECT external_db_id FROM external_db WHERE db_name LIKE ?',
+            -PARAMS => [ [ $external_db_name . '%', SQL_VARCHAR ] ],
+        )
+    };
+    return unless @external_db_ids;
+
+    my $external_db_placeholders = join ', ', '?' x @external_db_ids;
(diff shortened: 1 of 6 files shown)


azangru:
Looks good.

Just wanted to point out that even when bots aren't hammering the site with ids that cause these queries with the `LIKE` operator to occur, the database is still slow.

Also, as you pointed out, the website indiscriminately sends queries to the database for every id. If it is possible to prove that `_` or `%` are invalid characters in these ids anyway, then perhaps the code could be modified in ensembl-website to fail early and to prevent such wasteful db queries.

sgiorgetti:
kind reminder @TamaraNaboulsi @veidenberg 
@azangru  - thanks ... but got greedy and added a new bit. Would you mind?
Thank you all!!

azangru (approved):

modules/Bio/EnsEMBL/DBSQL/DBEntryAdaptor.pm
```diff
@@ -2024,15 +2019,21 @@ SQL
...
+
+    @external_db_ids = @{
+        $self->dbc()->sql_helper()->execute_simple(
+            -SQL    => 'SELECT external_db_id FROM external_db WHERE db_name LIKE ?',
+            -PARAMS => [ [ $external_db_name . '%', SQL_VARCHAR ] ],
+        )
+    };
+    return unless @external_db_ids;
+
+    my $external_db_placeholders = join ', ', '?' x @external_db_ids;
```
veidenberg, line 2035: `'?' x @external_db_ids` repeats the placeholder as string, creating invalid SQL queries like `... IN (???) ...` when there are multiple external DB matches (instead of `(?, ?, ?)`). Repeating the placeholder in list context should fix it:
```suggestion
    my $external_db_placeholders = join ', ', ('?') x @external_db_ids;
```

sgiorgetti: right ...

PR #752, opened 2026-06-03
Query speed optimisations

sgiorgetti:
## Description

SQL Optimisations: index creation for speeding up queries.
For doing it in a quick-ish fashion, I introduced a nasty&dangerous thing, which is forcing the use of an index.
This is "necessary" because MySQL optimiser gets confused by other (similar) existing index, and does not want to use the last and better one.
To avoid excessive index manipulation at this stage I forced the use of the best index for the query.

This must be changed the sooner the better to avoid painful surprises down the road.

## Use case

Slow queries.

## Benefits

Faster queries.

## Possible Drawbacks

Pain because of direct/hardcoded index name.

## Testing

No test required changing, but test DBs had to be regenerated
Test suite could run successfully.

### DIFF
--- a/modules/Bio/EnsEMBL/DBSQL/GeneAdaptor.pm
+++ b/modules/Bio/EnsEMBL/DBSQL/GeneAdaptor.pm
@@ -528,12 +528,14 @@ sub fetch_all_by_domain {
 
   throw("domain argument is required") unless ($domain);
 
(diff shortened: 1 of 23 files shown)


azangru (approved):
As a frontend developer, I am deeply impressed and humbled by this sql.

TamaraNaboulsi (approved):
Looks good to me.
The only thing I would stress is to mention somewhere that this should be changed later to avoid problems. Maybe a comment in the GeneAdaptor.pm code itself?

PR #753, opened 2026-06-08
Fix placeholder generation for external_db_ids

sgiorgetti:
## Description

'?' x @external_db_ids repeats the placeholder as string, creating invalid SQL queries like ... IN (???) ... when there are multiple external DB matches (instead of (?, ?, ?)). Repeating the placeholder in list context should fix it:

## Benefits

As described

## Possible Drawbacks

no

## Testing

no


TamaraNaboulsi (approved):
LGTM

